PR #1, opened 2026-04-25
port mention resolution, smart mention stripping, empty response filter, and verbosity config from GX internal bot

PranavBakre:
Slack mentions now resolve to display names in prompts and thread history. Only the bot's own @mention is stripped from incoming messages; other user mentions are preserved as context. Empty or NO_SLACK_MESSAGE responses are suppressed. Session verbosity defaults to quiet (SESSION_DEFAULT_VERBOSITY).

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,7 +1,7 @@
 import { loadConfig } from "./config.ts";
 import { createSlackApp } from "./slack/app.ts";
 import { registerEventHandlers } from "./slack/events.ts";
-import { formatToolStatuses, shouldPostResponseToSlack } from "./slack/formatting.ts";
+import { formatToolStatuses, extractAssistantText, shouldPostResponseToSlack } from "./slack/formatting.ts";
 import { SlackResponder } from "./slack/responder.ts";
 import { SessionManager } from "./session/manager.ts";
 import { createSessionStore } from "./session/store/factory.ts";
@@ -42,6 +42,11 @@ sessionManager.onEvent = (session, event) => {
   }
   if (session.verbosity === "quiet") return;
   if (event.type === "assistant") {
+    const text = extractAssistantText(event);
+    if (text) {
+      responder.updateStatus(session.channel, session.threadId, text);
+    }
+
     const statuses = formatToolStatuses(event);
     for (const status of statuses) {
       log.info("tool", `thread=${session.threadId} ${status}`);
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,7 +1,7 @@
 import { loadConfig } from "./config.ts";
 import { createSlackApp } from "./slack/app.ts";
 import { registerEventHandlers } from "./slack/events.ts";
-import { formatToolStatuses, extractAssistantText } from "./slack/formatting.ts";
+import { formatToolStatuses, shouldPostResponseToSlack } from "./slack/formatting.ts";
 import { SlackResponder } from "./slack/responder.ts";
 import { SessionManager } from "./session/manager.ts";
 import { createSessionStore } from "./session/store/factory.ts";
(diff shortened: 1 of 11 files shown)


src/slack/thread-context.ts
```diff
@@ -84,30 +127,32 @@ async function fetchThreadHistory(
...
-          text: (m.text ?? "").replace(/<@[A-Z0-9]+>\s*/g, "").trim(),
+          user,
+          userName: user === "unknown" ? user : await resolveUserName(app, user),
+          text: (await resolveSlackMentions(app, m.text ?? "")).trim(),
           ts: m.ts!,
           isBot: !!(botUserId && m.user === botUserId),
           fileNames,
         };
-      });
+      }));
```
pranav-growthx, line 153: **Should-fix [Query Performance]:** `Promise.all` over up to 100 thread messages, where each message fires `resolveUserName` + `resolveSlackMentions` (which itself calls `resolveUserName` per mention). On a thread's first load with many unique participants, this bursts dozens of concurrent `users.info` API calls. Slack's Tier 2 rate limit for `users.info` is ~20 req/min -- a 50-message thread from 20+ users will hit it.

The `userNameCache` helps on repeat users, but the first encounter fires all at once via `Promise.all`.

**Suggested fix:** Pre-collect all unique user IDs from the thread messages and inline mentions, resolve them in a single pass (sequentially or with a concurrency limiter), *then* build the `ThreadMessage` array using the pre-populated cache. This also avoids redundant lookups when the same user both authored a message and is mentioned in another.

src/session/types.ts
```diff
@@ -29,7 +30,8 @@ export interface ThreadSession {
 
 export function createSession(
   threadId: string,
-  channel: string
+  channel: string,
+  defaultVerbosity: SessionVerbosity = "quiet",
```
pranav-growthx: **Should-fix [Business Logic]:** The default parameter here is `"quiet"`, but `config.ts:69` defaults `SESSION_DEFAULT_VERBOSITY` to `"normal"`. In production, `manager.ts` always passes the config value explicitly, so production sessions default to **"normal"** (not "quiet").

The PR description says "Session verbosity defaults to quiet" which contradicts the actual production behavior when no env var is set.

Either:
- Change the config default in `config.ts` from `"normal"` to `"quiet"` to match the stated intent, or
- Change this parameter default to `"normal"` so both defaults agree

The mismatch creates confusion about which default is authoritative.

src/index.ts
```diff
@@ -42,13 +42,6 @@ sessionManager.onEvent = (session, event) => {
...
   if (session.verbosity === "quiet") return;
   if (event.type === "assistant") {
-    // Show text content as live status (gets overwritten each turn)
-    const text = extractAssistantText(event);
-    if (text) {
-      responder.updateStatus(session.channel, session.threadId, text);
-    }
-
-    // Show tool use as status
     const statuses = formatToolStatuses(event);
```
pranav-growthx, line 52: **Should-fix [Product Thinking]:** This PR silently removes the live text status updates (previously handled by `extractAssistantText`). On `main`, when verbosity was not "quiet", Claude's intermediate text content blocks were shown as live status messages in the Slack thread. Now only tool-use statuses are shown.

For users who explicitly set `!normal` or `!verbose`, they lose text streaming visibility -- they'll only see tool names, not what Claude is writing.

If this removal is intentional (e.g., the live text was too noisy or caused issues), no code change needed, but it should be documented in the PR description since it's a user-visible behavior change. If unintentional, the `extractAssistantText` block should be preserved here.

pranav-growthx (changes requested):
Requesting changes -- three should-fix issues found. See inline comments.

pranav-growthx (approved):
Re-reviewed the full PR (all 3 commits) after previous should-fix items were addressed.

**Previous findings -- all resolved:**
1. Slack API rate limit -- TODO comment added at thread-context.ts:130, documented in v2-backlog.md. Accepted as low-priority tech debt.
2. Verbosity default mismatch -- createSession default parameter changed from "quiet" to "normal" to match config.ts. Test updated.
3. Live text status removed -- extractAssistantText restored in formatting.ts, streaming text status block restored in index.ts.

**Re-review result:** Two consecutive clean passes across all six review dimensions (logic, safety, product, query performance, consistency, surface). No new issues found.

Approved.

PR #3, opened 2026-04-30
feat(worktree): per-thread worktrees + register_worktree MCP tool (Scope-1)

PranavBakre:
## Summary

Lays Scope-1 of the bug-pipeline worktree feature so persistent agents stop touching the human dev's bare repos under \`~/openclaw-projects/\`.

- New \`ThreadSession.worktreePaths: Record<string,string>\` for multi-repo bug threads (existing single \`worktreePath\` preserved for the \`!repo\` flow).
- New \`RepoConfig.worktreeSetupCommand?: string\` — optional setup script the WorktreeManager runs instead of \`git fetch + worktree add\`.
- New MCP tool \`mcp__slack-bot__register_worktree({ thread_id, repo, branch? })\` — lead calls on intake; tool creates the worktree and persists the path via the refetch-then-mutate pattern.
- \`buildWorkspaceBlock\` renders a multi-repo \`<workspace>\` block when \`worktreePaths\` is non-empty; single-repo format unchanged for non-bug threads.
- Agent prompts updated:
  - \`common/runtime-environment.md\`: paths come from the workspace block, never \`~/openclaw-projects/\` directly (universal rule, scoped to be safe for non-bug agents).
  - \`lead.md\`: new intake step calling \`register_worktree\` per routed repo.
  - \`reproducer.md\`: phase-2 placeholder — no \`git checkout\` in bare repo and no \`pnpm dev\` self-spawn; wait for \`!devserver\` (lands in step 3).
  - \`thinker.md\`: workspace-block reference at the top.

Design doc: \`docs/features/bug-pipeline-worktrees.md\`. This PR ships only the "Build order → 1. Scope-1" section.

**Out of scope (deferred):** dev-server lifecycle (\`devCommand\`/\`devPort\`/\`readyUrl\`), \`proper-lockfile\` queue, \`!devserver\` directive, \`lifecycle/dev-server.ts\`. Those are step 2/3 of the rollout.

## Test plan

- [x] \`bun test\` — 182 pass / 0 fail (added \`thread-context.test.ts\` and \`worktree/manager.test.ts\`; updated \`types.test.ts\` and \`args.test.ts\` for the new field)
- [ ] Manual: file a bug in #bugs-backlog after merge, confirm worktrees get created, agents reference them in their tool calls (grep stream output for the worktree path)
- [ ] Manual: verify \`!repo gx-backend\` flow in #junior still creates a single-repo workspace block (no regression)

### DIFF
--- a/src/mcp/slack-server.ts
+++ b/src/mcp/slack-server.ts
@@ -15,10 +15,14 @@ import { WebClient } from "@slack/web-api";
 import { z } from "zod";
 import { readFile } from "fs/promises";
 import { log } from "../logger.ts";
+import type { SessionStore } from "../session/store/interface.ts";
+import type { WorktreeManager } from "../worktree/manager.ts";
 
 const MCP_PORT = Number(process.env.MCP_PORT ?? "3456");
 
 let slack: WebClient;
+let sessionStore: SessionStore | undefined;
+let worktreeManager: WorktreeManager | undefined;
 
 function registerTools(server: McpServer) {
   server.registerTool(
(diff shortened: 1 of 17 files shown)


src/mcp/slack-server.ts
```diff
@@ -229,13 +233,87 @@ function registerTools(server: McpServer) {
...
+              type: "text" as const,
+              text: `Error: unknown repo "${repoName}". Check REPOS config for valid names.`,
+            },
+          ],
+        };
+      }
+
+      let worktreePath: string;
+      try {
+        worktreePath = await worktreeManager.createWorktree(repoName, thread_id, branch);
```
PranavBakre: 🔴 **Blocker — `branch` passed as `baseRef`: parameter type mismatch**

`createWorktree(repoName, thread_id, branch)` — but `createWorktree`'s third parameter is `baseRef?: string` (the *git starting-point commit*, e.g. `origin/main`), not a branch name. The MCP tool schema documents `branch` as a *branch name override* for the new worktree (description: "Branch name override. Defaults to slack/<thread_id>"), which is a completely different concept.

Effect:
- Default call (no `branch` arg): `createWorktree(repo, threadId, undefined)` → uses `repo.defaultBase` → correct by accident.
- Call with override, e.g. `{ branch: "feature/xyz" }` intending to name the new worktree branch differently: `git worktree add <path> -b slack/<threadId> feature/xyz` — starts the worktree at commit `feature/xyz` instead of `origin/main`, AND the branch is STILL named `slack/<threadId>` not `feature/xyz`. The override silently does the wrong thing on both axes.

The design doc §"Who creates worktrees" says the branch defaults to `slack/<threadId>` "unless caller overrides" — the override is clearly meant to change the new branch's *name*, not the base commit.

Note: lead.md only shows the default call `register_worktree({ thread_id, repo })` with no branch override, so production is safe for now. But the `branch` field is exposed in the tool schema and its semantics are wrong — a future caller will use it incorrectly.

**Fix:** Add a separate `branchOverride?: string` param to `createWorktree` alongside the existing `baseRef?`. Use `branchOverride ?? branchName` as the `-b` arg in `git worktree add`. In the MCP handler, pass `branch` as `branchOverride` (not `baseRef`), and drop it from the `createWorktree` `baseRef` position.

src/worktree/manager.ts
```diff
@@ -25,14 +25,27 @@ export class WorktreeManager {
     const branchName = `slack/${threadId}`;
     const base = baseRef ?? repo.defaultBase;
```
PranavBakre: 🟡 **Should-fix — `base` is dead code in the `worktreeSetupCommand` branch**

`const base = baseRef ?? repo.defaultBase` is computed unconditionally, but when `repo.worktreeSetupCommand` is set the value is never forwarded — the script only receives `[worktreePath, branchName]`. The design doc says 2 args, so this is intentional. But sitting before the if/else it looks like an oversight. Move it into the `else` branch:

```ts
if (repo.worktreeSetupCommand) {
  const setupCmd = ...;
  await this.runCommand([setupCmd, worktreePath, branchName], repo.path);
} else {
  const base = baseRef ?? repo.defaultBase; // only used here
  await this.runGit(['fetch', 'origin'], repo.path);
  await this.runGit(['worktree', 'add', worktreePath, '-b', branchName, base], repo.path);
}
```

src/worktree/manager.test.ts
```diff
@@ -0,0 +1,118 @@
...
+  writeFileSync(failScript, `#!/usr/bin/env bash\necho "permission denied" >&2\nexit 1\n`);
+  chmodSync(failScript, 0o755);
+});
+
+afterAll(() => {
+  rmSync(repoRoot, { recursive: true, force: true });
+});
+
+describe("WorktreeManager.createWorktree", () => {
+  it("falls back to git worktree add when worktreeSetupCommand is not set", async () => {
```
PranavBakre, line 69: 🟡 **Should-fix — no test for `baseRef` forwarding**

The default-flow test doesn't assert that an explicit `baseRef` actually reaches `git worktree add` as the starting-point commit. The test repo's self-remote has only `main`, so passing a wrong ref would still succeed (the command would just fail at push time, not here). Add a case that creates a second commit / tag in `beforeAll`, passes it as `baseRef`, and asserts the worktree's HEAD SHA matches. Without this, the parameter mismatch in `slack-server.ts:272` has zero test signal.

PranavBakre (commented):
Review complete — changes requested. See inline comments.

**Summary of findings:**
- 🔴 1 Blocker (`src/mcp/slack-server.ts:272`)
- 🟡 2 Should-fix (`src/worktree/manager.ts:26`, `src/worktree/manager.test.ts:63`)

PR #4, opened 2026-04-30
feat(dev-server): junior owns dev-server lifecycle (Scope-2a)

PranavBakre:
## Summary

Scope-2a of the bug-pipeline worktree feature. Today reproducer phase-2 spawns \`pnpm dev\` itself and the child process leaks when the agent's turn ends; this PR moves dev-server lifecycle ownership from the agent to junior. Independently fixes the leaked-process bug, and is the prerequisite for Scope-2b's queue.

- New \`RepoConfig\` fields: \`devCommand?\`, \`devPort?\`, \`readyUrl?\`. Repos that don't run dev servers leave them unset.
- New \`DevServerManager\` (\`src/lifecycle/dev-server.ts\`): tracks \`{pid, branch, startedAt, lastUsedAt}\` per repo. Methods: \`bootstrap()\`, \`ensure(repo, branch)\`, \`kill(repo)\`, \`killAll()\`, \`status()\`. Idle TTL of 20 min sweeper started in constructor.
- \`bootstrap()\`: ensures the dev-server worktree at \`<repo>/.claude/worktrees/slack-dev-server\` exists for each repo with a \`devCommand\`; advisory port-conflict check via TCP probe; logs loudly without aborting if something else holds the port (Scope-2b's lock acquire path will surface it definitively).
- \`ensure()\` uses \`git fetch && git reset --hard origin/<branch>\` rather than \`checkout\` because checkout rejects when the target branch is checked out elsewhere (the bare repo). Falls back to local-only branches when no upstream.
- \`devCommand\` splits on whitespace via \`Bun.spawn\` (no shell) — handles \`"pnpm dev"\`, \`"npm run dev"\`, \`"bun run dev"\`. Shell-quoted commands not supported (deferred, JSDoc'd).
- Wired into junior boot: instantiated after \`WorktreeManager\` in \`src/index.ts\`, \`bootstrap()\` runs before \`app.start()\`, \`killAll()\` runs in graceful shutdown.

**Out of scope (deferred to Scope-2b):** \`proper-lockfile\` queue, \`!devserver\` directive, \`.lock\`/\`.queue\` files, prompt edits to reproducer/lead pointing at the directive. \`DevServerManager.ensure()\` is the wrap point for the lock acquire; Scope-2b will wrap it, not replace it.

Design doc: \`docs/features/bug-pipeline-worktrees.md\`.

## Test plan

- [x] \`bun test\` — 202 pass / 0 fail (184 pre-existing + 18 new in \`dev-server.test.ts\`)
- [x] Scope-1's \`!repo\` flow untouched (no regression — single \`worktreePath\` path doesn't go through \`DevServerManager\`)
- [ ] Manual: junior boot logs the bootstrap step for each \`devCommand\` repo; conflicting listener triggers the loud warning
- [ ] Manual: graceful shutdown kills tracked dev-server PIDs

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -6,6 +6,22 @@ export interface RepoConfig {
   defaultBase: string;
   /** Optional setup script for worktree creation. Run as `<repo.path>/<command> <worktreePath> <branch>`. */
   worktreeSetupCommand?: string;
+  /**
+   * Command to start the dev server (e.g. "pnpm dev", "npm run dev").
+   * Split on whitespace and passed directly to Bun.spawn — no shell.
+   * Leave unset for repos that don't run a dev server.
+   */
+  devCommand?: string;
+  /**
+   * Port the dev server listens on (e.g. 3000, 8000).
+   * Used for readiness probing and external-listener conflict detection.
+   */
+  devPort?: number;
+  /**
+   * URL junior curls to confirm the dev server is ready (expects HTTP 200).
+   * Defaults to http://localhost:<devPort> when unset.
+   */
+  readyUrl?: string;
 }
 
 export type SessionStoreKind = "memory" | "sqlite";
--- a/src/index.ts
+++ b/src/index.ts
@@ -12,6 +12,7 @@ import { cleanupStaleSessions } from "./lifecycle/cleanup.ts";
 import { AgentRouter } from "./agents/router.ts";
 import { SupportRouter } from "./support/router.ts";
 import { WorktreeManager } from "./worktree/manager.ts";
+import { DevServerManager } from "./lifecycle/dev-server.ts";
 import { startMcpServer } from "./mcp/slack-server.ts";
 import { log } from "./logger.ts";
 
@@ -23,6 +24,7 @@ log.info("boot", `Session store: ${config.session.store}`);
 const sessionManager = new SessionManager(store, config);
 const agentRouter = new AgentRouter(config.repos, ".claude/agents");
 const worktreeManager = new WorktreeManager(config.repos);
+const devServerManager = new DevServerManager(config.repos, worktreeManager);
 startMcpServer(config.slack.botToken, store, worktreeManager);
 sessionManager.agentRouter = agentRouter;
 sessionManager.worktreeManager = worktreeManager;
@@ -114,7 +116,7 @@ sessionManager.onError = (session, error) => {
 
 registerHomeTab(app, store, config.session.homeWindowMs);
 
-setupGracefulShutdown(sessionManager, store);
+setupGracefulShutdown(sessionManager, store, devServerManager);
 
 // Periodic health checks
 setInterval(() => {
@@ -134,6 +136,10 @@ setInterval(() => {
 }, config.session.cleanupIntervalMs);
 
 (async () => {
+  // Bootstrap dev-server worktrees and check for external port conflicts before
+  // accepting any Slack events.
+  await devServerManager.bootstrap();
+
   await app.start();
 
   // Resolve bot identity before registering event handlers
--- a/src/index.ts
+++ b/src/index.ts
@@ -137,8 +137,17 @@ setInterval(() => {
 
 (async () => {
   // Bootstrap dev-server worktrees and check for external port conflicts before
-  // accepting any Slack events.
-  await devServerManager.bootstrap();
+  // accepting any Slack events. Failure here is non-fatal — log loudly and
+  // continue so other features keep working even if a single repo's worktree
+  // setup is broken (missing remote, dirty state, etc.).
+  try {
+    await devServerManager.bootstrap();
+  } catch (err) {
+    log.error(
+      "boot",
+      `dev-server bootstrap failed: ${err instanceof Error ? err.message : String(err)}`,
+    );
+  }
 
   await app.start();
 
--- a/src/lifecycle/dev-server.test.ts
+++ b/src/lifecycle/dev-server.test.ts
@@ -25,7 +25,7 @@ import { tmpdir } from "node:os";
 import { join } from "node:path";
 import type { RepoConfig } from "../config.ts";
 import { WorktreeManager } from "../worktree/manager.ts";
-import { DevServerManager } from "./dev-server.ts";
+import { DevServerManager, resolveReadyUrl } from "./dev-server.ts";
 
 // ---------------------------------------------------------------------------
 // Helpers shared across suites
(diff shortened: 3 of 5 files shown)


PranavBakre:
## Review: feat(dev-server): junior owns dev-server lifecycle (Scope-2a)

**Verdict: changes-requested** — two should-fix items, two nice-to-haves. GitHub blocks self-review REQUEST_CHANGES; findings are posted here instead.

---

### Findings

#### 🟡 Should-fix

**`src/index.ts:141` — bootstrap failure can silently abort boot**

`bootstrap()` is awaited here with no `try/catch`. If it throws, the IIFE exits before `app.start()` — the bot never starts accepting Slack events, with no log line explaining why.

Right now `bootstrap()` is defensively written so it won't throw: `worktreeExists` has an internal try/catch, `isPortHeld` wraps everything in a Promise that always resolves. But the design doc states explicitly: *failure to bootstrap should NOT crash boot — it should log and proceed.* That contract is implicit and fragile. Any future addition to the `bootstrap()` loop that isn't equally defensive will silently kill the bot on startup.

Fix — wrap the call site:

```ts
try {
  await devServerManager.bootstrap();
} catch (err) {
  log.error("boot", `dev-server bootstrap failed: ${err}`);
  // non-fatal: dev-server features won't work but the bot still starts
}
```

Alternatively, add an explicit try/catch around the outer `for` loop inside `bootstrap()` itself — mirrors the existing per-createWorktree guard and extends non-fatal policy to `worktreeExists` and `isPortHeld` too.

---

**`src/lifecycle/dev-server.ts:70` — concurrent `ensure()` calls double-spawn with no guard**

Two callers entering `ensure("gx-backend", "fix/foo")` simultaneously (once Scope-2b wires `!devserver` to real concurrent threads) will both:

1. Read `current = undefined` — both miss the reuse branch.
2. Both call `await this.kill(repoName)` — no-op since state is empty.
3. Both run `git fetch` and `git reset --hard` on the **same worktree path** concurrently — two concurrent git writes to the same working tree index.
4. Both call `Bun.spawn(...)` — two processes competing for the same port.
5. The second `state.set(repoName, { pid: newPid })` overwrites the first — the first spawned PID is orphaned and holds the port forever.

Scope-2b's `proper-lockfile` is the correct fix and is explicitly deferred. What's needed before 2b lands is a comment naming this precondition so a future caller doesn't add a second code path to `ensure()` without realising the hazard:

```ts
// NOTE: ensure() is not re-entrant for the same repo. Two concurrent callers
// will double-spawn and orphan one PID. Scope-2b adds a per-repo filesystem
// lock (proper-lockfile) to serialise access. Until then, ensure() must only
// be called from a single serialised dispatch path.
async ensure(repoName: string, branch: string): Promise<DevServerInfo> {
```

A soft in-flight guard (`Map<string, Promise<DevServerInfo>>`) would be a stronger belt-and-suspenders fix if you want protection before 2b lands.

---

#### 🟢 Nice-to-have

**`src/lifecycle/dev-server.test.ts:406` — idle TTL sweeper detection logic is untested**

The test calls `manager.kill()` directly rather than invoking the sweeper callback. This proves `kill()` works but never exercises the `if (s.pid != null && now - s.lastUsedAt > IDLE_TTL_MS)` detection condition in the timer. To cover the detection logic you'd need an injectable TTL param or a test-only `sweepNow()` method.

---

**`src/lifecycle/dev-server.test.ts:343` — placeholder test asserts JS string interpolation, not the function under test**

`expect(`http://localhost:41234`).toBe(`http://localhost:${41_234}`)` asserts that JS numeric literals interpolate — it doesn't call `resolveReadyUrl()`. Would be more useful as a direct call with and without `readyUrl` set.

---

### What's clean

- **`ensure()` reuse-vs-restart logic**: correct. Branch match + live PID → reuse. Branch mismatch or dead PID → kill-then-respawn.
- **Liveness check**: `process.kill(pid, 0)` is the standard probe. Used correctly in `isPidAlive()`.
- **`kill()` SIGKILL fallback**: SIGINT → 200ms-interval poll for 5s → SIGKILL. Not a tight loop, ESRCH is caught on both signals.
- **Idle sweeper**: `clearInterval` in `dispose()`, `unref()` so it doesn't hold the event loop, errors caught per-repo.
- **Socket cleanup in `isPortHeld()`**: `settled` flag prevents double-destroy on all three events (connect, error, timeout).
- **Worktree path for dev-server**: `<repo>/.claude/worktrees/slack-dev-server`. Branch `dev-server-slot/<repoName>` — distinct namespace from thread worktrees (`slack/<threadId>`). No collision possible.
- **Bootstrap idempotency**: guarded by `worktreeExists` check before `createWorktree`.
- **`config.ts` additions**: all three fields (`devCommand`, `devPort`, `readyUrl`) are optional, backward compatible, JSDoc'd.
- **`shutdown.ts`**: optional param, `killAll()` runs in the same `Promise.all` as session teardown — parallel, no ordering requirement.
- **index.ts wiring**: `devServerManager` created before `setupGracefulShutdown`, `bootstrap()` before `app.start()`.
- **No regression to Scope-1**: `DevServerManager` is wired only at `index.ts`, not threaded into session management or support routing.
- **No leaked secrets**.
- **Typecheck clean. 18/18 tests pass.**

---

*Two clean passes required for approval. The two should-fix items above need resolution before this can be approved.*

PranavBakre:
Re-review of commit 7a6606c. Verified all four items from the first-pass changes-requested:

**1. bootstrap try/catch (index.ts:138-143)**
catch calls log.error("boot", ...) through the structured logger (stdout + daily log file), then the IIFE continues to app.start(). Not silent, not blocking.

**2. ensure() re-entrancy comment (dev-server.ts)**
JSDoc sits directly on ensure(), names the orphaned-PID failure mode, identifies Scope-2b's lockfile path as today's only caller, and gives the exact in-memory fix (Map<repoName, Promise<DevServerInfo>>) for a future second caller. Load-bearing and in the right place.

**3. DevServerManagerOptions constructor**
Third parameter is additive with a default of {}. Production call site in index.ts is untouched. Defaults resolve to the original constants (20 min TTL, 60s sweep interval).

**4. sweepNow() and resolveReadyUrl tests**
Both idle-TTL tests plant state into the private map, exercise the same predicate as the interval sweeper (pid != null && now - lastUsedAt > idleTtlMs), and assert on kill outcome + PID liveness. resolveReadyUrl tests assert the actual derivation (port fallback vs explicit URL precedence). Not tautologies.

Two consecutive clean passes. **approved**

PR #5, opened 2026-04-30
feat(dev-server): !devserver directive + per-repo lockfile queue (Scope-2b)

PranavBakre:
## Summary

Final scope of the bug-pipeline worktree feature. Wraps Scope-2a's \`DevServerManager.ensure()\` in a per-repo \`proper-lockfile\` queue, exposes the \`!devserver\` directive (humans + agents post it the same way), and rewires reproducer phase-2 + lead's validation orchestration to use it.

This closes the RE-ENTRANCY PRECONDITION flagged in Scope-2a — \`DevServerQueue.acquire\` is now the lock that makes concurrent \`ensure()\` callers safe.

### Code

- **Dep**: \`proper-lockfile@4.1.2\` + \`@types/proper-lockfile\`.
- **\`src/lifecycle/dev-server-queue.ts\`** (new): \`DevServerQueue\` class. \`acquire(repo, branch, threadId, slotTimeoutMs=10*60_000)\` writes to \`<repo>/.claude/dev-server/.queue\` (NDJSON, append-only — O_APPEND is atomic for small writes), acquires \`.lock\` via proper-lockfile (10 min stale), atomically writes \`.lock.meta.json\` (write-tmp + rename), calls \`DevServerManager.ensure()\`. \`release\` clears meta + queue line + unlocks. Idempotent. \`stealStale(repo)\` kills any orphan PID listening on \`devPort\` when proper-lockfile takes over a stale lock. \`readQueueDepth(repo)\` exposes \`{holder, waiters}\` for \`!devserver status\`.
- **\`src/support/router.ts\`**: \`parseDevserverDirective\` plus inline handlers for \`!devserver <branch> [repo]\`, \`!devserver status\`, \`!devserver kill <repo>\`. Directive intercepted in \`AgentDispatcher.handleMessage\` before the agent-routing path. Reply messages post via the slack-bot MCP server's slack client. Repo defaults to all entries in \`session.worktreePaths\` when omitted.
- **\`src/lifecycle/dev-server.ts\`**: \`bootstrap()\` now writes \`<repo>/.claude/dev-server/.gitignore\` so \`.lock*\` and \`.queue*\` don't show as untracked. RE-ENTRANCY PRECONDITION comment updated to point at \`DevServerQueue.acquire\` as the resolution.
- **\`src/index.ts\`**: instantiates \`DevServerQueue\` after \`DevServerManager\`, passes it + sessionStore + slackClient + repos to \`SupportRouter\`.

### Prompt edits

- **\`reproducer.md\`** phase=validation: post \`!devserver <branch>\` and wait for junior's \`ready @ localhost:<port>\`. Handles failed / port-held / slot-timeout cases. Describes same-branch warm reuse vs branch-change restart. Auto-release after 10 min.
- **\`lead.md\`**: new "Validation phase" section. Categorical — dispatch \`!devserver <branch>\` first, wait for ready, then dispatch \`!reproducer validate\`.

### Test plan

- [x] \`bun test\` — 234 pass / 0 fail (204 pre-existing + 25 new in \`dev-server-queue.test.ts\` + 5 new in \`router.test.ts\`)
- [x] No regression to \`!repo\` flow, Scope-1 multi-repo workspace block, or Scope-2a lifecycle methods
- [ ] Manual: post \`!devserver fix/foo gx-backend\` in #bugs-backlog, confirm ready reply, confirm \`.lock.meta.json\` + \`.queue\` files appear under \`<repo>/.claude/dev-server/\`
- [ ] Manual: post a second \`!devserver fix/bar gx-backend\` while the first holds the lock, confirm \"queued behind 1\" message and serial execution
- [ ] Manual: \`!devserver status\`, \`!devserver kill gx-backend\`

### Out of scope (deferred to step 4)

- \`docs/features/process-lifecycle.md\` — needs to mention DevServerQueue as the serialization layer.
- \`docs/features/bug-pipeline-worktrees.md\` — status line should flip from \"design / not yet built\" to \"landed in #3 / #4 / this PR\".
- Promote internal \`devServerManager\` cast in \`handleDevserverKill\` to a public \`DevServerQueue.kill()\` method (small follow-up).

### DIFF
--- a/src/lifecycle/dev-server-queue.ts
+++ b/src/lifecycle/dev-server-queue.ts
@@ -1,0 +1,395 @@
+/**
+ * DevServerQueue — Scope-2b
+ *
+ * Wraps DevServerManager with a per-repo filesystem lock (proper-lockfile) so
+ * concurrent callers are serialized. This is the answer to the RE-ENTRANCY
+ * PRECONDITION comment in DevServerManager.ensure() — callers should go through
+ * DevServerQueue.acquire() rather than calling ensure() directly.
+ *
+ * Lock file layout (all under <repo>/.claude/worktrees/slack-dev-server/):
+ *   .lock            — proper-lockfile's lock directory (created by the library)
+ *   .lock.meta.json  — holder metadata: { holderThreadId, holderPid, branch, acquiredAt }
+ *   .queue           — NDJSON, one line per waiter: { threadId, branch, enqueuedAt }
+ *
+ * The .gitignore at that path covers .lock* and .queue* so they don't show as
+ * untracked (gitignore written by DevServerManager.bootstrap() — Scope-2a).
+ */
+
+// proper-lockfile: lock() returns a release function; unlock() is not needed separately.
+import { lock as lockFile } from "proper-lockfile";
+import { writeFileSync, readFileSync, existsSync, mkdirSync } from "node:fs";
+import { rename, writeFile } from "node:fs/promises";
+import { join } from "node:path";
+import { log } from "../logger.ts";
+import type { RepoConfig } from "../config.ts";
+import type { DevServerManager, DevServerInfo } from "./dev-server.ts";
+
+// ---------------------------------------------------------------------------
+// Public types
+// ---------------------------------------------------------------------------
+
+export interface HolderMeta {
+  holderThreadId: string;
+  holderPid: number;
+  branch: string;
+  acquiredAt: number;
+}
+
+export interface WaiterMeta {
+  threadId: string;
+  branch: string;
+  enqueuedAt: number;
+}
+
+export interface QueueDepth {
+  holder: HolderMeta | null;
+  waiters: WaiterMeta[];
+}
+
+export interface AcquireResult {
+  /** Call this to release the lock and update metadata. Idempotent. */
+  release: () => Promise<void>;
+  info: DevServerInfo;
+}
+
+// ---------------------------------------------------------------------------
+// Default slot timeout: 10 minutes
+// ---------------------------------------------------------------------------
+const DEFAULT_SLOT_TIMEOUT_MS = 10 * 60 * 1_000;
+
+// ---------------------------------------------------------------------------
+// DevServerQueue
+// ---------------------------------------------------------------------------
+
+export class DevServerQueue {
+  private devServerManager: DevServerManager;
+  private repos: RepoConfig[];
+
+  constructor(devServerManager: DevServerManager, repos: RepoConfig[]) {
+    this.devServerManager = devServerManager;
+    this.repos = repos;
+  }
+
+  /**
+   * Acquire the per-repo lock, ensure the dev server is running on `branch`,
+   * and return a handle with a `release()` function.
+   *
+   * The lock path is the dev-server worktree directory. `proper-lockfile`
+   * creates a `.lock` subdirectory inside it — this means the worktree must
+   * already exist (created by bootstrap()).
+   *
+   * Behaviour:
+   *   1. Append `{ threadId, branch, enqueuedAt }` to `.queue` (O_APPEND NDJSON).
+   *   2. Call `proper-lockfile.lock()` with stale=slotTimeoutMs. Blocks until acquired.
+   *   3. Call `devServerManager.ensure(repoName, branch)`.
+   *   4. Write `.lock.meta.json` atomically (tmp + rename).
+   *   5. Return `{ release, info }`.
+   *
+   * `release()`:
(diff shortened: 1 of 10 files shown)


src/lifecycle/dev-server-queue.ts
```diff
@@ -0,0 +1,395 @@
...
+    appendToQueue(lockDir, waiterEntry);
+
+    log.info("dev-server-queue", `enqueued thread=${threadId} repo=${repoName} branch=${branch}`);
+
+    let releaseLockFile: (() => Promise<void>) | null = null;
+    let released = false;
+
+    try {
+      // Step 2: acquire the filesystem lock. Retries forever with jitter.
+      releaseLockFile = await lockFile(lockDir, {
```
PranavBakre, line 130: **Blocker: `onCompromised` not wired — process crashes on stale-lock steal**

The `lockFile()` call uses proper-lockfile's default `onCompromised` handler, which is `(err) => { throw err; }`. `setLockAsCompromised` is invoked from inside a `setTimeout` callback (`updateLock`). An uncaught throw from a `setTimeout` propagates as an uncaught exception and crashes the Bun/Node process.

This fires in exactly the scenario the design doc describes: when a stale lock is stolen by the next acquirer, the incumbent process's update timer detects the mtime is no longer ours, calls `setLockAsCompromised`, the default handler throws, and the process dies.

`stealStale()` already has the right orphan-kill logic but is never called anywhere. Wire it as the `onCompromised` callback:

```ts
releaseLockFile = await lockFile(lockDir, {
  stale: slotTimeoutMs,
  retries: { forever: true, minTimeout: 1_000, maxTimeout: 5_000 },
  onCompromised: (err) => {
    log.warn(
      'dev-server-queue',
      `lock compromised repo=${repoName}: ${err.message}; running stealStale`,
    );
    this.stealStale(repoName).catch((killErr) => {
      log.warn('dev-server-queue', `stealStale failed: ${killErr}`);
    });
    released = true;
    removeFromQueue(lockDir, threadId);
  },
});
```

Note: `onCompromised` is synchronous — proper-lockfile does not await the callback. The `stealStale` call must be fire-and-forget with a catch.

src/support/router.ts
```diff
@@ -1,13 +1,73 @@
...
+
+  // !devserver <branch> [repo]
+  // Branch names can contain slashes and hyphens. If there are two tokens
+  // separated by whitespace, the last one is the (optional) repo name.
+  // We use a simple split: if exactly one token — it's the branch; if two —
+  // first is branch, second is repo.
+  // Edge: "status" and "kill" are reserved first tokens handled above.
+  if (!rest) return null;
+
+  const tokens = rest.split(/\s+/);
```
PranavBakre, line 68: **Blocker: `!devserver kill` (bare, no repo) silently misparsed as an acquire directive**

The `killMatch` regex at line 47 requires `\S+` after `kill`, so bare `!devserver kill` does not match. Execution falls through to the branch-parse block here. `tokens = ['kill']`, length 1, returns `{ kind: 'acquire', branch: 'kill', repos: [] }` — which attempts a full dev-server acquire for a branch literally named `"kill"`.

A human forgetting the repo argument gets no error and a spurious acquire attempt instead.

Fix: add an explicit guard before the branch-parse fallthrough:

```ts
// !devserver kill with no repo — reject and explain
if (rest === 'kill') {
  return null; // caller should reply with usage hint
}
```

If `parseDevserverDirective` returns `null`, `handleDevserverDirective` is not called and the event is silently dropped. Better to add a `{ kind: 'malformed'; reason: string }` variant so the handler can post a usage reply to Slack.

PranavBakre (commented):
Code review — two blockers found. See inline comments.

PranavBakre:
Re-review (second pass): both prior blockers verified fixed. No new issues found.

**Blocker 1 — `onCompromised` wiring** (`src/lifecycle/dev-server-queue.ts` lines 133–151): `lockFile()` is called with a synchronous handler (no `async` keyword). `stealStale` is invoked fire-and-forget with a `.catch`. `released = true` is set synchronously before the async chain fires, so a subsequent `release()` hits the `if (released) return` guard and is a no-op. Handler does not throw. Fully correct.

**Blocker 2 — `!devserver kill` bare-kill guard** (`src/support/router.ts` lines 47–58): `rest = line.trim().slice("!devserver".length).trim()` means both `"!devserver kill"` and `"!devserver kill   "` produce `rest === "kill"`, which hits the malformed path before the acquire tokenizer. `handleDevserverDirective` dispatches the malformed variant to `postSlack`. Tests assert both the discriminant (`kind === "malformed"`) and that `reason` contains `"kill <repo>"`.

237 / 237 tests pass. No regressions.

Two consecutive clean passes. **approved**

PR #8, opened 2026-04-30
feat(commands): add !mute and !unmute thread commands

PranavBakre:
## Summary

- Adds `!mute` to silence a bot on a thread without resetting the session — all incoming messages are dropped silently until unmuted
- Adds `!unmute` to resume normal operation
- Mute check runs **after** command parsing so `!unmute` works even while muted
- `muted` field persists in SQLite via JSON blob; `??= false` in `normalizeSession` handles existing sessions

## Test plan

- [ ] `!mute` in active thread → bot acknowledges, subsequent messages get no response
- [ ] `!unmute` while muted → bot acknowledges and resumes responding
- [ ] `!status` while muted → still responds (commands bypass mute gate)
- [ ] Bot restart with muted session → mute persists (SQLite-backed)
- [ ] `!help` shows `!mute` and `!unmute` entries

### DIFF
--- a/src/session/manager.ts
+++ b/src/session/manager.ts
@@ -82,6 +82,9 @@ export class SessionManager {
     this.rememberSeenMessage(dedupeKey);
 
     const session = await this.getOrCreateSession(event);
+
+    if (session.muted) return;
+
     const agentSession = this.getOrCreateAgentSession(session, agentName);
 
     if (agentSession.status === "busy") {
@@ -120,6 +123,8 @@ export class SessionManager {
       const handled = await this.handleCommand(session, event);
       if (handled) return;
     }
+
+    if (session.muted) return;
 
     if (session.status === "busy") {
       session.pendingMessages.push({
@@ -221,6 +226,8 @@ export class SessionManager {
           "`!verbose` — Verbose output",
           "`!adhoc <text>` — Add ad-hoc item to calendar",
           "`!bugs <text>` — Add bug item to calendar",
+          "`!mute` — Stop responding until !unmute",
+          "`!unmute` — Resume responding",
           "`!help` — Show this help",
         ].join("\n");
         this.onCommandResponse?.(event, helpText);
@@ -325,6 +332,20 @@ export class SessionManager {
           `4. Reply with a one-line confirmation.`,
         ].join("\n");
         return false;
+      }
+
+      case "mute": {
+        session.muted = true;
+        await this.store.set(session.threadId, session);
+        this.onCommandResponse?.(event, "Muted. I'll stop responding until you `!unmute`.");
+        return true;
+      }
+
+      case "unmute": {
+        session.muted = false;
+        await this.store.set(session.threadId, session);
+        this.onCommandResponse?.(event, "Unmuted.");
+        return true;
       }
 
       default:
--- a/src/session/manager.ts
+++ b/src/session/manager.ts
@@ -199,6 +199,7 @@ export class SessionManager {
           : "never";
         const lines = [
           `*Status:* ${session.status}`,
+          `*Muted:* ${session.muted ? "yes" : "no"}`,
           `*Agent:* ${session.agentType ?? "default"}`,
           `*Repo:* ${session.targetRepo ?? "none"}`,
           `*Worktree:* ${session.worktreePath ?? "none"}`,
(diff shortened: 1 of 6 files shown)


src/session/manager.ts
```diff
@@ -121,6 +124,8 @@ export class SessionManager {
       if (handled) return;
     }
 
+    if (session.muted) return;
```
PranavBakre, line 127: **Should-fix (Business Logic -- drain path bypasses mute gate):** This mute check guards new incoming messages in `runSingleSession`, but the drain path in `onRunComplete` (line ~614) calls `runClaudeWithAgent` directly -- it never re-checks `session.muted`.

Scenario:
1. Thread is busy (Claude processing a message).
2. User sends another message -> buffered in `pendingMessages`.
3. User sends `!mute` -> `handleCommand` runs, sets `session.muted = true`, returns.
4. Claude finishes -> `onRunComplete` sees `pendingMessages.length > 0` -> drains them through `runClaudeWithAgent` -> Claude responds, violating the mute.

Fix: in `onRunComplete`, before draining, check `fresh.muted`. If muted, discard the pending messages (matching the PR description: "all incoming messages are dropped silently") and transition to idle.

```ts
// In onRunComplete, before the pending-messages drain:
if (fresh.muted) {
  if (isTopLevel) {
    fresh.pendingMessages = [];
    fresh.status = "idle";
  } else if (agentSession) {
    agentSession.pendingMessages = [];
    agentSession.status = "done";
  }
  await this.store.set(fresh.threadId, fresh);
  return;
}
```

src/session/manager.ts
```diff
@@ -327,6 +334,20 @@ export class SessionManager {
         return false;
       }
 
+      case "mute": {
```
PranavBakre, line 338: **Should-fix (Product Thinking -- `!status` doesn't show mute state):** The `!status` handler (line ~198) does not include `muted` in its output. A user who mutes a thread and comes back later has no way to verify if mute is active -- they just see silence and wonder if the bot is broken.

The PR test plan lists "`!status` while muted" as a verification step, but there is nothing in the status output to verify.

Add a line to the status block:
```ts
`*Muted:* ${session.muted ? "yes" : "no"}`,
```

---

**Nice-to-have (UX -- idempotent feedback):** Consider responding differently when `!mute` is sent on an already-muted thread ("Already muted.") or `!unmute` on a non-muted thread ("Not muted."). Currently both silently re-set the boolean. A user who sends `!mute` twice might think it toggled off and back on. Not a blocker -- the state is correct regardless.

PranavBakre (commented):
Two should-fix issues and one nice-to-have. The mute gate placement is correct (commands run before the mute check, so `!unmute` works while muted). SQLite migration with `??= false` is solid. But the drain path and `!status` output have gaps. See inline comments.

src/session/manager.ts
PranavBakre: Verified in commit 98f5f23. The drain path now checks `fresh.muted` before processing buffered messages:

```ts
if (pendingMessages.length > 0 && fresh.muted) {
  // discard buffered messages, set status to idle
  ...
}
```

This correctly handles the scenario I described: messages buffered before `!mute` are discarded when the current turn completes, instead of being drained through `runClaudeWithAgent`. The `fresh` re-read from the store ensures the mute flag reflects mid-turn `!mute` commands. Fix looks good.

src/session/manager.ts
PranavBakre: Verified in commit 98f5f23. `!status` now includes:

```ts
`*Muted:* ${session.muted ? "yes" : "no"}`
```

Placed right after the status line, before agent/repo info. The test plan's "`!status` while muted" step is now verifiable.

Re: the nice-to-have (idempotent mute/unmute messaging) -- still not addressed, but that was explicitly marked as non-blocking. No objection to merging without it.

PranavBakre (commented):
**Re-review: Approved** (posting as comment because GitHub blocks self-approval)

Re-review of fix commit 98f5f23. Both should-fix issues from the original review are addressed:

1. **Drain path** -- `onRunComplete` now checks `fresh.muted` before draining buffered messages. When muted, pending messages are discarded and the session transitions to idle instead of spawning a new Claude process. The `fresh` re-read from the store correctly picks up mid-turn mute commands.

2. **!status mute visibility** -- Status output now includes a `Muted: yes/no` line, making the mute state observable.

No new issues introduced by the fix commit. The `agentSession.status = "idle"` (vs `"done"`) in the mute-discard path is a defensible choice -- the agent didn't complete work on those messages, they were discarded, so "idle" is semantically accurate.

Zero blockers, zero should-fix items. Ready to merge.

PR #9, opened 2026-04-30
fix(slack): surface user IDs in thread context so Claude can tag people

PranavBakre:
## Summary

- `resolveSlackMentions` now produces `@Name (<@USERID>)` instead of just `@Name` — Claude sees the Slack mention format in thread history and copies it
- Thread history prefixes now include the user ID: `User(Dash <@U12345>):` instead of `User(Dash):`
- Added a one-line preamble note: "To tag a user, use their Slack mention format `<@USERID>`"

**Root cause:** Thread context stripped `<@USERID>` → `@Name` before sending to Claude, leaving the model with no source for user IDs. Claude wrote plain `@Name` in responses, which Slack does not convert to a real mention/notification.

## Test plan

- [ ] Send a message in a thread — reply should use `<@USERID>` format when mentioning participants (renders as a real Slack tag)
- [ ] Verify thread history still reads naturally (display names still present alongside IDs)

### DIFF
--- a/src/slack/thread-context.ts
+++ b/src/slack/thread-context.ts
@@ -67,7 +67,7 @@ export async function resolveSlackMentions(
   let resolved = text;
   for (const match of matches) {
     const name = await resolveUserName(app, match[1]);
-    resolved = resolved.replace(match[0], `@${name}`);
+    resolved = resolved.replace(match[0], `@${name} (<@${match[1]}>)`);
   }
   return resolved;
 }
@@ -174,6 +174,7 @@ export async function buildPromptPreamble(
     `Thread: ${threadTs}`,
     `You are responding in this thread. You already have the full thread history below.`,
     `Do NOT use Slack search or read tools to find this thread — you already have all the context you need.`,
+    `To tag a user, use their Slack mention format \`<@USERID>\` (shown in thread history as \`User(Name <@USERID>)\`). Plain \`@Name\` does not notify them.`,
     ``,
     `If you decide this message does NOT need a reply (e.g. it's noise, already handled, or you've finished silent work), your final response must be exactly the sentinel \`${NO_SLACK_MESSAGE}\` and nothing else — no surrounding text, no explanation, no quotes. Anything else will be posted to the channel verbatim.`,
     ``,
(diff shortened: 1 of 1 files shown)


src/slack/thread-context.ts
```diff
@@ -67,7 +67,7 @@ export async function resolveSlackMentions(
   let resolved = text;
   for (const match of matches) {
     const name = await resolveUserName(app, match[1]);
-    resolved = resolved.replace(match[0], `@${name}`);
+    resolved = resolved.replace(match[0], `@${name} (<@${match[1]}>)`);
```
PranavBakre: **Should-fix (Bug):** When the same user is `<@>`-mentioned twice in one message, this produces garbled output.

Trace with input `"Hey <@U123> and <@U123>"`:
- `matchAll` collects two matches, both with `match[0] = "<@U123>"`
- Iter 1: `replace("<@U123>", "@Name (<@U123>)")` → `"Hey @Name (<@U123>) and <@U123>"`
- Iter 2: `replace("<@U123>", "@Name (<@U123>)")` finds the *first* `<@U123>` — which is now the one inside the parens from iter 1 — producing `"Hey @Name (@Name (<@U123>)) and <@U123>"`

The bug structure existed pre-PR but was latent: the old replacement `@${name}` didn't contain the search string `<@U123>`, so `replace`'s first-occurrence semantics never bit. The new replacement embeds the search string, which activates it.

**Fix:** Use `String.replace` with a regex and callback instead of iterating matches. This is both correct and more efficient (resolves each unique user ID once):

```ts
const ids = [...new Set(matches.map(m => m[1]))];
const names = Object.fromEntries(
  await Promise.all(ids.map(async id => [id, await resolveUserName(app, id)]))
);
return text.replace(mentionPattern, (_, id) => `@${names[id]} (<@${id}>)`);
```

This also addresses the TODO on line 216 about batch-resolving user IDs — same optimization, different call site.

PranavBakre (commented):
One should-fix found. The duplicate-mention bug is the only issue -- the rest of the change is clean. See inline comment for details and a fix.

PranavBakre (commented):
Re-review of fix commit c626753: the double-replacement garbling issue is correctly fixed.

The single-pass `String.replace(regex, callback)` approach operates on original string positions, not on intermediate replacement output, so the same `<@USERID>` appearing twice no longer matches inside a prior replacement's parenthesized copy. Parallel name resolution via `Promise.all` on deduplicated IDs is a nice bonus. Edge cases (multiple users, no matches, name-resolution failures) all handled correctly.

PR #10, opened 2026-04-30
feat(home): show mute status on home tab

PranavBakre:
## Summary

- Stats summary gains a *Muted: N* count (only shown when N > 0)
- Each muted session shows `:no_bell: *Muted*` on its status line, between the busy/idle indicator and the agent/repo info

### DIFF
--- a/src/slack/home.ts
+++ b/src/slack/home.ts
@@ -74,13 +74,20 @@ function buildHomeBlocks(
   const idle = allSessions.filter((s) => s.status === "idle").length;
   const draining = allSessions.filter((s) => s.status === "draining").length;
   const withErrors = allSessions.filter((s) => s.lastError !== null).length;
+  const muted = allSessions.filter((s) => s.muted).length;
+
+  const summaryParts = [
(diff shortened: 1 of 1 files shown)


PranavBakre (commented):
Reviewed: two consecutive clean passes, no blockers or should-fix items found. Cannot self-approve via this account, so leaving as a comment review.

Verified:
- `session.muted` is always a defined boolean (`createSession` at src/session/types.ts:75 initializes to `false`; sqlite store backfills via `??=` at src/session/store/sqlite.ts:194), so the filter and ternary are safe in both stores.
- Conditional spread for the Muted summary entry preserves the exact prior output when no sessions are muted.
- Refactor to `summaryParts.join(" |  ")` keeps the rendered text identical to the previous template literal for the zero-muted case.
- `:no_bell:` flag placement between status and agent info correctly reflects that muted threads can still be mid-turn (mute only blocks new messages — see src/session/manager.ts:86,127).

No DB queries added. No security, runtime, type-safety, or product-flow concerns. Recommend merging.

PR #11, opened 2026-04-30
feat(pipeline): gate reproducer on read-only bugs only

PranavBakre:
Gate !reproducer on read-only bugs. Write-path bugs skip straight to !thinker to avoid prod side-effects.

### DIFF
--- a/.claude/agents/lead.md
+++ b/.claude/agents/lead.md
@@ -24,12 +24,12 @@ If you find yourself doing ANY of the following, STOP — that work belongs to a
 
 | You're tempted to... | STOP. Dispatch instead: |
 |---|---|
-| Open Playwright / browser tools to verify something | `!reproducer reproduce: <the thing>` |
+| Open Playwright / browser tools to verify something | `!reproducer reproduce: <the thing>` (read-only bugs only — see classification rule above) |
 | Read product code in `~/openclaw-projects/<repo>/` | `!thinker <hypothesis seed>` (thinker reads code as part of its job) |
 | Restart dev servers, check ports, run `npm/pnpm/bun run dev` | `!reproducer` (reproducer owns the dev environment) |
 | Edit any file in `~/openclaw-projects/<repo>/` | `!thinker proceed` (thinker writes the fix in its scoping phase) |
 | Run `git checkout`, create branches, open PRs in target repos | `!thinker proceed` |
-| "Verify the fix works" with a browser | `!reproducer validate the fix on branch <name>` |
+| "Verify the fix works" with a browser | `!reproducer validate the fix on branch <name>` (read-only bugs only — same server, same mutation risk) |
 | "This is a small fix, faster to do myself" | NO. The architecture is the architecture. Dispatch. |
 
 The temptation to do work yourself is strongest when the fix looks small. That's exactly when the architecture's audit value matters most — every bypass is a precedent that erodes the discipline. Even when you're 100% sure you know the fix, dispatch `!thinker` and let the pipeline run.
@@ -76,7 +76,7 @@ Concretely:
 
 6a. *(Read-only bug)* Dispatch `!reproducer <prompt>` with observability context (failing endpoint, exception class, deploy state, affected user). Reproducer reads the files itself but a tight target prompt prevents cold exploration. If the bug looks access-gated, mention the admin-creds path explicitly so reproducer applies the impersonation fallback. Then proceed: reproducer → thinker → review.
 
-6b. *(Write-path bug)* **Skip reproducer entirely.** Dispatch `!thinker <prompt>` directly with the observability findings and affected-user context. Note in the Slack thread: "Skipping reproducer — write-path bug, reproduction would fire real prod writes. Going straight to thinker." Proceed: thinker → review.
+6b. *(Write-path bug)* **Skip reproducer entirely — both phases.** Dispatch `!thinker <prompt>` directly with the observability findings and affected-user context. Note in the Slack thread: "Skipping reproducer — write-path bug, both reproduction and validation would fire real prod writes." Proceed: thinker → review.
 
 **Identity rule when dispatching reproducer — non-negotiable:**
 
@@ -90,7 +90,7 @@ If you don't yet have an affected user ID, ASK in the thread before dispatching 
 Invariants:
 
 - Observability always precedes reproduction and validation.
-- Reproducer is only dispatched for read-only bugs. Write-path bugs go straight to thinker — no exceptions, even if the write looks "safe" or "small."
+- **Reproducer (both phases) is only dispatched for read-only bugs.** Write-path bugs go straight to thinker, and skip validation after review — no exceptions, even if the write looks "safe" or "small." Both phases walk the same server with the same mutation risk.
 - If reproduction is `mismatch`, do not proceed to scoping the wrong issue.
 - If reproduction is `not-reproduced`, escalate to a human instead of retrying blindly.
 
@@ -135,6 +135,8 @@ Two ways the cycle breaks. Pick the right one:
 Never post both commentary AND a `NO_SLACK_MESSAGE` — pick one.
 
 ## Validation phase: gate `!reproducer validate` on a `!devserver` ready reply
+
+**Read-only bugs only.** Write-path bugs skip both reproduction and validation phases — after review approval, go directly to the merge flow. The validation section below applies only when this bug was classified as read-only.
 
 When the thinker has opened a fix PR and you're ready to validate, the dev server must be running on the fix branch BEFORE you dispatch the reproducer. Junior owns the dev-server slot — agents (including reproducer) MUST NOT spawn `pnpm dev` themselves.
 
--- a/.claude/agents/lead.md
+++ b/.claude/agents/lead.md
@@ -12,7 +12,13 @@ Your job is to triage the report, dispatch the right persistent agents, synthesi
 
 ## CATEGORICAL RULE — every bug routes through the pipeline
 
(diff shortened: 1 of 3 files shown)


.claude/agents/lead.md
```diff
@@ -66,7 +72,11 @@ Concretely:
...
+5. **Classify the failure path before dispatching.** Ask: "Does reaching the failure require submitting a form, clicking a generate/save/create button, or triggering a mutation?" If yes → write-path bug, go to step 6b. If no → read-only bug, go to step 6a.
+
+6a. *(Read-only bug)* Dispatch `!reproducer <prompt>` with observability context (failing endpoint, exception class, deploy state, affected user). Reproducer reads the files itself but a tight target prompt prevents cold exploration. If the bug looks access-gated, mention the admin-creds path explicitly so reproducer applies the impersonation fallback. Then proceed: reproducer → thinker → review.
+
+6b. *(Write-path bug)* **Skip reproducer entirely.** Dispatch `!thinker <prompt>` directly with the observability findings and affected-user context. Note in the Slack thread: "Skipping reproducer — write-path bug, reproduction would fire real prod writes. Going straight to thinker." Proceed: thinker → review.
```
PranavBakre: 🔴 **Blocker — internal contradiction with the validation phase.**

Step 6b says write-path bugs go `thinker → review`, full stop, and the new invariant on line 93 reinforces this with no exceptions. But the unchanged `Validation phase` section (around line 137) still says `!reproducer validate` is dispatched after thinker opens a PR, and the unchanged STOP table row ("Verify the fix works with a browser → `!reproducer validate the fix on branch <name>`") assumes reproducer is always available for validation.

The PR's stated rationale (avoid prod side-effects from Playwright walking a mutation flow) only applies to **Phase 1 reproduction** — that runs on the prod-connected environment. **Phase 2 validation runs on local-dev** (`!devserver` flow, branch checked out, localhost), where mutations are intended and safe. So the gate as written shouldn't necessarily extend to Phase 2.

Either way, lead.md is inconsistent with itself today. Pick one:
- **(a)** Phase 2 validation still runs for write-path bugs (most likely correct — prod safety isn't the concern on local-dev). In that case, 6b should read `Proceed: thinker → reproducer (validation only) → review`, and add a note that reproducer is dispatched only for phase=validation, not phase=reproduction. Reproducer.md (lines 7–14, 38–42) currently assumes it has its own `reproduction.md` to read in Phase 2; that assumption breaks for write-path bugs and needs an explicit fallback ("if no reproduction.md, walk the path described in scoping.md cold").
- **(b)** Phase 2 is also intentionally skipped. Then say so explicitly, justify it (no `reproduction.md` to anchor the walk?), update the Validation phase section to gate on "read-only bug only," and update the STOP table row to qualify reproducer-validate as read-only too.

The current diff does (a) implicitly via the unchanged validation section but says (b) in step 6b. One of them has to give.

.claude/agents/lead.md
```diff
@@ -66,7 +72,11 @@ Concretely:
...
 3. Fan out observability first with **parallel Task calls** to `nr-research`, `sentry-fetch`, and `vercel-status` in a single assistant message (concurrent execution).
 4. Read the three output files, synthesize key findings into one Slack message that references each file path. Don't dump raw NRQL or Sentry events — surface what matters (failing endpoint, blast radius, deploy correlation, exception class).
-5. Dispatch `!reproducer <prompt>` with observability context (failing endpoint, exception class, deploy state, affected user). Reproducer reads the files itself but a tight target prompt prevents cold exploration. If the bug looks access-gated, mention the admin-creds path explicitly so reproducer applies the impersonation fallback.
+5. **Classify the failure path before dispatching.** Ask: "Does reaching the failure require submitting a form, clicking a generate/save/create button, or triggering a mutation?" If yes → write-path bug, go to step 6b. If no → read-only bug, go to step 6a.
```
PranavBakre, line 75: 🔴 **Blocker — broken handoff to thinker for write-path bugs.**

`thinker.md` Phase 1 (lines 13–16) lists `$BUG_DIR/reproduction.md` as a required input — "what the reproducer actually saw." Under this PR, write-path bugs skip reproducer entirely, so `reproduction.md` will not exist when thinker is dispatched.

The thinker doesn't have a documented branch for "reproduction.md is missing." In practice it'll either:
- Read fail / no-op silently and proceed without that anchor (degraded hypothesis quality — Phase 1 is calibrated around having a concrete failing path + network signal).
- Treat it as a read error and abort.

Fix options (not mutually exclusive):
- Update `thinker.md` to handle the write-path branch: "if `reproduction.md` is absent, the bug is write-path and you're operating from observability + report only. Lean harder on Sentry stack frames, NR endpoint signatures, and code reading."
- Update step 6b to spell out what extra context lead must include in the dispatch prompt (exact failing endpoint, request body shape, deploy correlation) to compensate for the missing reproduction trace.

Without one of these, the write-path path silently degrades thinker's Phase 1 quality — which is the most consequential phase in the pipeline.

.claude/agents/lead.md
```diff
@@ -12,7 +12,13 @@ Your job is to triage the report, dispatch the right persistent agents, synthesi
 
 ## CATEGORICAL RULE — every bug routes through the pipeline
 
-**EVERY bug, no exceptions, regardless of how small/obvious/trivial the fix looks, goes through `!reproducer` and `!thinker`.** The pipeline gates exist for consistency and audit, not just for hard bugs. A 2-line CSS fix follows the same path as a backend race condition.
+**EVERY bug, no exceptions, regardless of how small/obvious/trivial the fix looks, goes through `!thinker`.** The pipeline gates exist for consistency and audit, not just for hard bugs. A 2-line CSS fix follows the same path as a backend race condition.
+
+**`!reproducer` is gated on read-only bugs.** Reproducer walks the UI by clicking through the actual product — on a prod-connected environment. If the failure path requires submitting a form, clicking Generate/Save/Create, or triggering any write operation (POST/PUT/PATCH/DELETE mutation), running reproducer would fire real side-effects on prod: LLM token spend, DB writes, emails, payments. For those bugs, skip reproducer and dispatch `!thinker` directly with the observability context.
+
```
PranavBakre, line 18: 🟡 **Should-fix — sister docs are now stale.**

This PR introduces a categorical change (reproducer is conditional, not unconditional) but only touches `lead.md`. Two other docs assert the old invariant:

- `docs/features/persistent-agents.md` line 56: "Reproducer runs _after_ observability completes — it needs failing endpoints, exception classes, deploy state as context." Implies unconditional dispatch.
- `docs/features/persistent-agents.md` line 62: "**Observability ALWAYS precedes UI verification.** Both for reproduction phase and validation phase. The reproducer runs both phases with observability context, never cold." Same — assumes reproducer always runs.
- `docs/features/bug-pipeline-worktrees.md` describes reproducer phase-1 as a standard pipeline stage with no read-only/write-path qualifier.

These should pick up the gate, otherwise next time someone reads them as architecture truth they'll re-derive the wrong rule. (Side note: the PR description text mentioned `docs/features/thread-commands.md` would change — that doc covers `!reset`/`!build`/etc., not bug-pipeline orchestration. Not updating it is correct; the doc that needs updating is `persistent-agents.md`.)

.claude/agents/lead.md
```diff
@@ -12,7 +12,13 @@ Your job is to triage the report, dispatch the right persistent agents, synthesi
...
+Classify before dispatching reproducer:
```
PranavBakre, line 19: 🟢 **Nice-to-have — STOP table doesn't carry the new gate.**

The categorical rule you just added qualifies reproducer as read-only-only, but the unchanged STOP table rows still say:
- "Open Playwright / browser tools to verify something → `!reproducer reproduce: <the thing>`"
- "Verify the fix works with a browser → `!reproducer validate the fix on branch <name>`"

If a future operator (or this same agent on a tired turn) consults the table without re-reading the categorical rule above it, they'll dispatch reproducer for a write-path bug. Worth adding `(read-only bugs only — see classification rule above)` to the reproducer rows so the table is self-consistent.

PranavBakre (commented):
Two blockers around internal consistency (validation phase contradicts step 6b; thinker.md depends on reproduction.md which write-path bugs won't produce), one should-fix on stale sister docs, one nice-to-have on the STOP table. The read-only/write-path classification itself is sound — the gaps are in the seams with the rest of the pipeline.

PranavBakre:
**Re-review: APPROVED — two consecutive clean passes.**

(GitHub blocks self-approval since this PR was opened under the same account, so leaving the verdict as a comment.)

Verified the three fixes:

1. **Validation phase contradiction (Blocker 1) — RESOLVED.** `lead.md:139` now opens the validation section with "Read-only bugs only. Write-path bugs skip both reproduction and validation phases — after review approval, go directly to the merge flow." The invariant at line 93 reinforces "both phases" with no exceptions. STOP-table rows at lines 27 and 32 carry the read-only qualifier. The merge flow at lines 152-165 doesn't depend on reproducer validation, so write-path's path (review-approved → merge flow) is internally consistent.

2. **Thinker handoff (Blocker 2) — RESOLVED.** `thinker.md:16-18` marks `reproduction.md` as "may be absent" and provides an actionable fallback. The fallback is complete: explains *why* it's absent (write-path bug skipped Phase 1), names the substitutes (observability files + direct code reading), tightens "verify with cheap evidence" from guidance to mandate ("every hypothesis needs a code-read or DB query check, not just observability inference"), and instructs Message 1 to disclose the missing trace so humans see the reduced epistemic state. Coheres with line-29 verification guidance and line-109's "don't dispatch reproducer directly" rule — for write-path bugs, lead correctly won't redispatch reproducer, so there's no escape-hatch gap.

3. **Sister doc staleness (Should-fix) — RESOLVED.** `persistent-agents.md` Full Vision bullets (55-56) and Invariant #1 (62) updated. Terminology ("read-only" / "write-path" / "both phases") used consistently across all three files.

No new blockers, should-fixes, or surface issues introduced. Safe to merge.

PR #12, opened 2026-04-30
feat(thinker): mock-run hypothesis when reproducer was skipped

PranavBakre:
## Summary

- For write-path bugs where `!reproducer` is skipped (both phases), thinker now runs a cheap local verification script after choosing a hypothesis and before posting Message 1
- Flow: query real prod data via MongoDB MCP → write inline `/tmp/hypothesis-check.ts` importing the suspect function from the worktree → run it → report `confirmed`/`inconclusive`/`errored` in Message 1's `verify:` column
- If hypothesis can't be localised to a single function (timing, multi-step state), the step is skipped and noted
- Adds `mcp__mongodb__find`, `mcp__mongodb__aggregate`, `mcp__mongodb__count` to thinker's tool manifest

## Test plan

- [ ] Review the mock-run section for completeness and edge case handling
- [ ] Verify the MongoDB MCP tool names match what's actually available in the agent runtime
- [ ] Verify Message 1 format still parses correctly with mock-run result in the verify column

### DIFF
--- a/.claude/agents/thinker.md
+++ b/.claude/agents/thinker.md
@@ -1,7 +1,7 @@
 ---
 name: thinker
 description: Persistent thinker — generates root-cause hypotheses, verifies each, picks the most likely, and writes the fix plan. Resists anchoring on the proximate cause.
-tools: Read, Write, Edit, Bash, Grep, Glob, mcp__mongodb__find, mcp__mongodb__aggregate, mcp__mongodb__count
+tools: Read, Write, Edit, Bash, Grep, Glob, mcp__mongodb__find, mcp__mongodb__aggregate, mcp__mongodb__list-databases, mcp__mongodb__list-collections, mcp__mongodb__collection-schema
 ---
 
 You are the Thinker persistent agent in a bug thread. Your job spans diagnosis AND scoping the fix — they are inseparable in practice and split into two phases.
@@ -36,14 +36,19 @@ Rank the hypotheses by likelihood after verification. Recommend ONE as the root 
 
 After choosing a hypothesis, run a cheap local script to add evidence before posting Message 1:
 
-1. **Localise the suspect code.** Identify the exact function or transformation you believe is broken (e.g. `events-service.ts:handleCreate`). If the hypothesis is about timing, multi-step state, or anything that can't be isolated to one function, skip to step 5 and note "script step skipped — hypothesis not localisable to a single function."
+1. **Localise the suspect code.** Identify the exact **pure transform or validation function** you believe is broken. Skip to step 5 with the `skipped` reason if any of these apply:
+   - The hypothesis is about the write operation itself (the bug lives in the write handler, not in upstream logic that feeds it).
+   - The hypothesis is about timing, race condition, or multi-step state that can't be replicated in a single function call.
+   - The hypothesis can't be isolated to one function.
 2. **Fetch real data.** Use the MongoDB MCP tools (`mcp__mongodb__find`, `mcp__mongodb__aggregate`) to query the prod data that would normally reach the suspect code.
-3. **Write an inline script.** Create a small script in `/tmp/` (e.g. `/tmp/hypothesis-check.ts`) that imports the suspect function from the worktree, feeds it the fetched data, and asserts on the output or catches the expected error.
-4. **Run it.** Execute with `bun /tmp/hypothesis-check.ts` (or the appropriate runtime for the repo).
+3. **Write a script inside the worktree.** Create a small test file inside the worktree (e.g. `<worktree-path>/scripts/hypothesis-check.ts`) — NOT in `/tmp/` — so that the repo's tsconfig path aliases resolve correctly. The script should import the suspect function, feed it the fetched data, and assert on the output or catch the expected error. The script must NOT call the write-path endpoint or any function that performs a DB write or external mutation.
+4. **Run it** from inside the worktree with `bun scripts/hypothesis-check.ts` (or the appropriate runtime).
 5. **Record the result:**
-   - Output/error matches the hypothesis → `mock-run: confirmed` — paste the key line in Message 1.
-   - Script passes cleanly → `mock-run: inconclusive` — note it (the real trigger may require write-path state that can't be replicated locally; this doesn't refute the hypothesis).
-   - Script errors for a setup reason (missing env, bad import) → `mock-run: errored — <reason>` — note it as a setup issue, not hypothesis evidence.
+   - Output/error matches the hypothesis → `mock-run: confirmed` — paste the key line in Message 1. **Redact any prod PII** (user IDs, emails, personal data) before including.
+   - Script ran cleanly and returned the *correct* value with the affected user's real data → `mock-run: refuted` — this undercuts the hypothesis; re-rank before posting.
+   - Script passed but couldn't fully replicate the trigger state → `mock-run: inconclusive` — note it; doesn't refute the hypothesis.
+   - Script errored for a setup reason (missing env, bad import, unresolved alias) → `mock-run: errored — <reason>` — note as a setup issue, not hypothesis evidence.
+   - Step 1 skip condition was met → `mock-run: skipped — <reason>`.
 
 Include the mock-run result in the `verify:` column of Message 1 alongside the other evidence.
 
--- a/.claude/agents/thinker.md
+++ b/.claude/agents/thinker.md
@@ -1,7 +1,7 @@
 ---
 name: thinker
 description: Persistent thinker — generates root-cause hypotheses, verifies each, picks the most likely, and writes the fix plan. Resists anchoring on the proximate cause.
-tools: Read, Write, Edit, Bash, Grep, Glob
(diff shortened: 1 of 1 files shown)


PranavBakre (commented):
Review (gh REQUEST_CHANGES blocked: PR author == reviewer auth). Treat as request-changes — see inline comments. One blocker (the mock-run flow can fire the exact prod write the reproducer was skipped to avoid), plus three should-fix items.

.claude/agents/thinker.md
```diff
@@ -30,6 +30,23 @@ For each hypothesis, **verify with cheap evidence first**: read the suspect code
...
+### Write-path supplement: mock-run the chosen hypothesis
+
+*(Only when `reproduction.md` is absent — the bug is write-path and reproducer was skipped.)*
+
+After choosing a hypothesis, run a cheap local script to add evidence before posting Message 1:
+
+1. **Localise the suspect code.** Identify the exact function or transformation you believe is broken (e.g. `events-service.ts:handleCreate`). If the hypothesis is about timing, multi-step state, or anything that can't be isolated to one function, skip to step 5 and note "script step skipped — hypothesis not localisable to a single function."
+2. **Fetch real data.** Use the MongoDB MCP tools (`mcp__mongodb__find`, `mcp__mongodb__aggregate`) to query the prod data that would normally reach the suspect code.
+3. **Write an inline script.** Create a small script in `/tmp/` (e.g. `/tmp/hypothesis-check.ts`) that imports the suspect function from the worktree, feeds it the fetched data, and asserts on the output or catches the expected error.
+4. **Run it.** Execute with `bun /tmp/hypothesis-check.ts` (or the appropriate runtime for the repo).
```
PranavBakre: **Blocker — Product / Safety:** The mock-run flow can fire the exact prod write the reproducer was skipped to avoid.

The PR rationale is: write-path bugs skip the reproducer because invoking the flow would trigger real prod side-effects (DB writes, emails, payments — see lead.md L17, L93). But step 3 directs thinker to *import the suspect function from the worktree and feed it the fetched data*. For write-path bugs, the suspect function is very often a write handler — that is the whole class of bug. Importing and invoking it from a `/tmp/` script will execute whatever the function does, including its DB writes and side-effect calls. The MongoDB MCP being read-only does not help here: the worktree code uses its own mongoose / API SDK with whatever env it picks up.

Net effect: the supplement could undo the safety property that motivated skipping the reproducer in the first place.

**Fix:** Add an explicit guardrail that the mock-run is **pure-transform only** — only run on hypotheses where the suspect code is a pure transformation / validator / serializer that does not call out to the DB, API, or any side-effect. If the hypothesis is *about* the write itself (the create handler writes the wrong field, the mutation has a race), mock-run is not applicable and should be skipped. This is a fourth skip reason alongside step 1's "not localisable to a single function."

A related, smaller concern: even the read-only MongoDB MCP queries are against prod (per `common/runtime-environment.md` line 53). Pasting prod row contents into Slack via Message 1 has PII implications — worth a sentence telling thinker to redact/abstract the row before pasting.

.claude/agents/thinker.md
```diff
@@ -30,6 +30,23 @@ For each hypothesis, **verify with cheap evidence first**: read the suspect code
...
+5. **Record the result:**
+   - Output/error matches the hypothesis → `mock-run: confirmed` — paste the key line in Message 1.
+   - Script passes cleanly → `mock-run: inconclusive` — note it (the real trigger may require write-path state that can't be replicated locally; this doesn't refute the hypothesis).
+   - Script errors for a setup reason (missing env, bad import) → `mock-run: errored — <reason>` — note it as a setup issue, not hypothesis evidence.
```
PranavBakre: **Should-fix — Logic / Result taxonomy:** The three buckets miss two real outcomes.

- **Refuted.** Script runs cleanly with the exact prod row from the failing user and the function returns the expected/correct value with no error. That is not 'inconclusive' — it is evidence the chosen hypothesis is wrong. As written, this gets labelled `inconclusive` and the wrong hypothesis still goes into Message 1 with a (mislabelled) signal.
- **Skipped.** Step 1 instructs thinker to skip to step 5 and note 'script step skipped — hypothesis not localisable to a single function', but step 5's three buckets do not include `skipped`. The skip case is a fourth state and should appear here so the `verify:` column has a canonical value (and so it cleanly composes with the new 'hypothesis is about the write itself' skip reason raised in the blocker).

Suggest expanding to: `confirmed` / `refuted` / `inconclusive` / `errored` / `skipped`, with explicit one-liners for each.

.claude/agents/thinker.md
```diff
@@ -30,6 +30,23 @@ For each hypothesis, **verify with cheap evidence first**: read the suspect code
...
 
+### Write-path supplement: mock-run the chosen hypothesis
+
+*(Only when `reproduction.md` is absent — the bug is write-path and reproducer was skipped.)*
+
+After choosing a hypothesis, run a cheap local script to add evidence before posting Message 1:
+
+1. **Localise the suspect code.** Identify the exact function or transformation you believe is broken (e.g. `events-service.ts:handleCreate`). If the hypothesis is about timing, multi-step state, or anything that can't be isolated to one function, skip to step 5 and note "script step skipped — hypothesis not localisable to a single function."
+2. **Fetch real data.** Use the MongoDB MCP tools (`mcp__mongodb__find`, `mcp__mongodb__aggregate`) to query the prod data that would normally reach the suspect code.
+3. **Write an inline script.** Create a small script in `/tmp/` (e.g. `/tmp/hypothesis-check.ts`) that imports the suspect function from the worktree, feeds it the fetched data, and asserts on the output or catches the expected error.
```
PranavBakre: **Should-fix — Runtime / Module resolution:** Writing the script in `/tmp/` and running with `bun /tmp/hypothesis-check.ts` will silently fail to resolve the worktree's path aliases.

GX repos (and most TS repos in this stack) use tsconfig path aliases like `@/lib/...`, `@/services/...`. `bun` resolves tsconfig paths relative to the script's location, not the import target — a script outside the worktree will not pick up the worktree's `tsconfig.json` or `node_modules`. The most likely outcome on real repos: import error → `mock-run: errored — <reason>` → no signal. The supplement gets neutered exactly when it would be most useful.

**Fix:** Direct thinker to write the script *inside* the worktree (e.g. `<worktree>/scripts/hypothesis-check.ts` or `<worktree>/.tmp/hypothesis-check.ts` with a gitignore note) and run with `bun` from the worktree root, so module resolution, path aliases, and `node_modules` all work. `/tmp/` is fine only for hypotheses on standalone files with zero alias usage, which is rare.

.claude/agents/thinker.md
```diff
@@ -1,7 +1,7 @@
 ---
 name: thinker
 description: Persistent thinker — generates root-cause hypotheses, verifies each, picks the most likely, and writes the fix plan. Resists anchoring on the proximate cause.
-tools: Read, Write, Edit, Bash, Grep, Glob
+tools: Read, Write, Edit, Bash, Grep, Glob, mcp__mongodb__find, mcp__mongodb__aggregate, mcp__mongodb__count
```
PranavBakre: **Should-fix — Tool grant completeness:** Frontmatter adds `find`, `aggregate`, `count`, but `common/runtime-environment.md` line 53 lists six MongoDB MCP tools available to agents in this runtime: also `mcp__mongodb__collection-schema`, `mcp__mongodb__list-collections`, `mcp__mongodb__list-databases`.

For the mock-run flow specifically, step 2 says 'query the prod data that would normally reach the suspect code.' To do that reliably without round-tripping through code-reads, thinker often needs to discover collection names or shape first — that is exactly what `collection-schema` and `list-collections` are for. As written, thinker has the query tools but not the discovery tools that make them usable on an unfamiliar collection.

Also minor: `count` is granted but never referenced in the new prose. Either reference it ('use `count` to confirm the failing-shape row exists before fetching') or drop it from the grant for tightness.

PranavBakre:
**Review verdict: APPROVED** (gh APPROVE blocked: PR author == reviewer auth — recording verdict as a comment).

Two consecutive clean passes. All four prior-round issues are resolved at 72b98f4.

- **Blocker (write-mutation):** Step 1 now skips when the hypothesis is about the write itself; step 3 explicitly forbids calling any DB-write or mutating function and limits scope to pure transforms / validators. Belt-and-suspenders against the prod-side-effect risk that motivated skipping the reproducer in the first place.
- **Should-fix (taxonomy):** expanded to confirmed / refuted / inconclusive / errored / skipped. Refuted now properly degrades a wrongly-chosen hypothesis instead of being silently labelled inconclusive.
- **Should-fix (path aliases):** script lives at `<worktree>/scripts/hypothesis-check.ts` and runs from the worktree root, so tsconfig paths and node_modules resolve.
- **Should-fix (tool grant):** dropped `count` (unreferenced), added `list-databases` / `list-collections` / `collection-schema` (the discovery tools needed to make find/aggregate usable on unfamiliar collections).
- **Bonus:** PII redaction note added for confirmed results before pasting into Message 1.

Cleared to merge.

PR #13, opened 2026-05-06
fix(worktree): always fetch + create worktree in manager; single-arg setup hook

PranavBakre:
## Summary

Junior's worktree manager now delegates worktree creation to the per-repo `setup-worktree.sh` when `worktreeSetupCommand` is configured, passing flag-based args: `<branch> --path <abs> [--base <ref>]`. The script owns `git fetch`, `git worktree add`, env copy, install, and MCPs. When unset, manager keeps the inline `git fetch origin --prune` + `git worktree add` flow.

Junior's path namespace is preserved: `getWorktreePath` still returns `<repo>/.claude/worktrees/slack-<thread>`. The per-repo defaults (which the script computes when called manually without a target_path) live entirely in the script -- Junior never sees them.

## What changed vs. the prior commit on this branch

The first commit on this branch put worktree creation into manager.ts and reduced setup hooks to single-arg post-create helpers. After review the call was reversed: scripts retain full ownership (so manual `./scripts/setup-worktree.sh feature/foo` still works end-to-end), and manager.ts hands them the target path when it has one. The new commit on top implements that.

## Tests

- Inline path: autofetch regression guard (fresh commit on origin/main visible after createWorktree).
- Delegation: marker file proves script received `[branch, --path, abs-target-path]` exactly; manager did not pre-create the worktree.
- Inverse autofetch guard: when delegating to a non-fetching fixture script, a fresh commit on origin/main is **absent** from the resulting worktree -- proves manager doesn't fetch behind the script's back.
- Failure propagation, baseRef forwarding (inline only), branchOverride for both inline and delegated paths.
- New: baseRef forwarded as `--base` to setup script when set explicitly.
- New: `--base` is **absent** from setup-script args when baseRef is unset.

## Fetch behavior change

`git fetch origin` is now `git fetch origin --prune`. Removes stale
remote-tracking refs that the old fetch preserved. Low-risk hygiene
improvement, surfaced per review feedback.

## baseRef forwarded to setup scripts

When `baseRef` is explicitly set (e.g., via `!branch staging` driving
`session.baseRef`), Junior forwards it as `--base <ref>` to the setup
script. Setup hooks called as: `<cmd> <branch> --path <abs> [--base <ref>]`.
When `baseRef` is unset, scripts pick their per-repo default — gx-backend
and gx-admin-client default to local HEAD; gx-client-next defaults to
origin/main.

## Coordinated merge

The matching script PRs are gx-backend#3082, gx-admin-client#738, gx-client-next#5007. Land this PR with those -- mismatched directions break:

- New manager + old scripts: old gx-backend/admin scripts reject the second arg with "Usage" error; old gx-client-next interprets it as a base-branch ref and `git worktree add` fails.
- Old manager (current main) + new scripts: old manager passes `<wt-path> <branch>` (path-first); new scripts read `$1` as branch_name and `git worktree add -b /abs/path/...` rejects the leading-slash branch name.

## Updated baseRef semantics

`--base <ref>` is now ALWAYS forwarded to setup scripts, defaulting to
`repo.defaultBase` when the caller doesn't override. Rationale: Junior
worktrees should be reproducible from a known ref (typically `origin/main`),
never from whatever the local checkout happens to be on. The per-repo
script defaults (gx-backend / gx-admin-client: local HEAD; gx-client-next:
origin/main) still apply for **manual** users who omit `--base`.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -5,10 +5,11 @@ export interface RepoConfig {
   path: string;
   defaultBase: string;
   /**
-   * Optional post-create hook. Junior creates the worktree itself
-   * (`git fetch` + `git worktree add`); if set, this is invoked afterward as
-   * `<repo.path>/<command> <abs-worktree-path>`. Use for env-file copying,
-   * dependency install, MCP migration.
+   * Optional. When set, Junior delegates worktree creation to this script via
+   * `<repo.path>/<command> <branch> <abs-target-path>`. The script must
+   * `git fetch`, `git worktree add` at the given path, and perform setup
+   * (env copy, install, MCPs). When unset, Junior runs
+   * `git fetch origin --prune` and `git worktree add` inline (no setup hook).
    */
   worktreeSetupCommand?: string;
   /**
--- a/src/config.ts
+++ b/src/config.ts
@@ -6,10 +6,13 @@ export interface RepoConfig {
   defaultBase: string;
   /**
    * Optional. When set, Junior delegates worktree creation to this script via
-   * `<repo.path>/<command> <branch> <abs-target-path>`. The script must
-   * `git fetch`, `git worktree add` at the given path, and perform setup
-   * (env copy, install, MCPs). When unset, Junior runs
-   * `git fetch origin --prune` and `git worktree add` inline (no setup hook).
+   * `<repo.path>/<command> <branch> --path <abs> [--base <ref>]`. The script
+   * must `git fetch`, `git worktree add` at the given path, and perform setup
+   * (env copy, install, MCPs). `--base <ref>` is forwarded only when the
+   * caller passes an explicit `baseRef` to `createWorktree`; when unset, the
+   * script applies its own per-repo default (e.g., local HEAD or
+   * `origin/main`). When unset, Junior runs `git fetch origin --prune` and
+   * `git worktree add` inline (no setup hook).
    */
   worktreeSetupCommand?: string;
   /**
--- a/src/config.ts
+++ b/src/config.ts
@@ -4,7 +4,12 @@ export interface RepoConfig {
   name: string;
   path: string;
   defaultBase: string;
-  /** Optional setup script for worktree creation. Run as `<repo.path>/<command> <worktreePath> <branch>`. */
+  /**
+   * Optional post-create hook. Junior creates the worktree itself
+   * (`git fetch` + `git worktree add`); if set, this is invoked afterward as
+   * `<repo.path>/<command> <abs-worktree-path>`. Use for env-file copying,
+   * dependency install, MCP migration.
+   */
   worktreeSetupCommand?: string;
   /**
    * Command to start the dev server (e.g. "pnpm dev", "npm run dev").
--- a/src/config.ts
+++ b/src/config.ts
@@ -8,11 +8,13 @@ export interface RepoConfig {
    * Optional. When set, Junior delegates worktree creation to this script via
    * `<repo.path>/<command> <branch> --path <abs> [--base <ref>]`. The script
    * must `git fetch`, `git worktree add` at the given path, and perform setup
-   * (env copy, install, MCPs). `--base <ref>` is forwarded only when the
-   * caller passes an explicit `baseRef` to `createWorktree`; when unset, the
-   * script applies its own per-repo default (e.g., local HEAD or
-   * `origin/main`). When unset, Junior runs `git fetch origin --prune` and
-   * `git worktree add` inline (no setup hook).
+   * (env copy, install, MCPs). Junior always forwards `--base <ref>`,
+   * defaulting to `repo.defaultBase` when the caller doesn't override —
+   * Junior worktrees are reproducible from a known ref, never from whatever
+   * the local checkout happens to be on. Manual users can omit `--base` to
+   * fall back to the script's per-repo default. When `worktreeSetupCommand`
+   * is unset, Junior runs `git fetch origin --prune` and `git worktree add`
+   * inline (no setup hook).
    */
   worktreeSetupCommand?: string;
   /**
--- a/src/worktree/manager.test.ts
+++ b/src/worktree/manager.test.ts
@@ -209,6 +209,8 @@ describe("WorktreeManager.createWorktree", () => {
       "slack/custom-thread",
       "--path",
       join(repoRoot, ".claude/worktrees/slack-custom-thread"),
+      "--base",
+      "origin/main",
     ]);
     expect(wtPath).toBe(
       join(repoRoot, ".claude/worktrees/slack-custom-thread"),
@@ -254,7 +256,7 @@ describe("WorktreeManager.createWorktree", () => {
     await wm.removeWorktree("delegate-baseref", "delegate-baseref-thread");
   });
 
-  it("does NOT forward --base when baseRef is unset", async () => {
+  it("forwards repo.defaultBase as --base when baseRef is unset", async () => {
     const repos: RepoConfig[] = [
       {
         name: "delegate-no-baseref",
@@ -271,9 +273,9 @@ describe("WorktreeManager.createWorktree", () => {
 
     const markerText = await Bun.file(setupMarker).text();
     const args = markerText.split("\n").filter((l) => l.length > 0);
-    expect(args).not.toContain("--base");
-    // And specifically: only branch + --path + path, nothing extra.
-    expect(args.length).toBe(3);
+    expect(args).toContain("--base");
+    expect(args).toContain("origin/main");
+    expect(args.length).toBe(5);
 
     await wm.removeWorktree("delegate-no-baseref", "no-baseref-thread");
   });
@@ -421,7 +423,7 @@ describe("WorktreeManager.createWorktree", () => {
     const args = (await Bun.file(setupMarker).text())
       .split("\n")
       .filter((l) => l.length > 0);
-    expect(args).toEqual(["fix/delegated-name", "--path", wtPath]);
+    expect(args).toEqual(["fix/delegated-name", "--path", wtPath, "--base", "origin/main"]);
 
     await wm.removeWorktree("delegate-branch-override", "delegate-override-thread");
   });
--- a/src/worktree/manager.test.ts
+++ b/src/worktree/manager.test.ts
@@ -48,31 +48,56 @@ beforeAll(async () => {
   await runIn(repoRoot, ["git", "remote", "add", "origin", repoRoot]);
   await runIn(repoRoot, ["git", "fetch", "-q", "origin"]);
 
-  // Fake setup script. Two args: <branch_name> <abs_target_path>. Owns the
-  // full worktree-creation lifecycle: fetch + worktree add + marker write.
-  // This mirrors the real contract — Junior never pre-creates the worktree
-  // when delegating; it hands the script the branch and target path and the
-  // script does the rest.
+  // Fake setup script. Flag-based contract:
+  //   <branch_name> --path <abs> [--base <ref>]
+  // Owns the full worktree-creation lifecycle: fetch + worktree add + marker
+  // write. This mirrors the real contract — Junior never pre-creates the
+  // worktree when delegating; it hands the script the branch, target path,
+  // and (optionally) base ref, and the script does the rest.
+  //
+  // The marker file dumps every received arg on its own line so tests can
+  // assert exact arg ordering, presence/absence of --base, etc.
   setupMarker = join(repoRoot, "setup-marker.txt");
   const setupScript = join(repoRoot, "fake-setup.sh");
   writeFileSync(
     setupScript,
     `#!/usr/bin/env bash
 set -e
-# Validate two-arg contract.
-if [[ "$#" -ne 2 ]]; then
-  echo "fake-setup: expected 2 args (branch, abs-target-path), got: $#" >&2
+# Dump every received arg on its own line BEFORE any validation, so even
+# malformed invocations leave a marker for tests to inspect.
+: > "${setupMarker}"
+for a in "$@"; do
+  printf '%s\\n' "$a" >> "${setupMarker}"
+done
+# Parse flag-based args.
+BRANCH=""
+ABS_TARGET=""
+BASE_REF=""
+while [[ $# -gt 0 ]]; do
+  case "$1" in
+    --path) ABS_TARGET="$2"; shift 2 ;;
+    --base) BASE_REF="$2"; shift 2 ;;
+    -*) echo "fake-setup: unknown flag: $1" >&2; exit 1 ;;
+    *)
+      if [[ -z "$BRANCH" ]]; then BRANCH="$1"; shift
+      else echo "fake-setup: unexpected positional: $1" >&2; exit 1
+      fi ;;
+  esac
+done
+if [[ -z "$BRANCH" || -z "$ABS_TARGET" ]]; then
+  echo "fake-setup: branch and --path required" >&2
   exit 1
 fi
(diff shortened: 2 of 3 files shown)


src/worktree/manager.ts
```diff
@@ -34,22 +34,26 @@ export class WorktreeManager {
...
+      // while letting manual callers of the script keep their default
+      // ergonomics. Resolve the command relative to repo.path so paths like
+      // "scripts/setup-worktree.sh" work without requiring the script on PATH.
       const setupCmd = repo.worktreeSetupCommand.startsWith("/")
         ? repo.worktreeSetupCommand
         : `${repo.path}/${repo.worktreeSetupCommand}`;
       await this.runCommand(
-        [setupCmd, worktreePath, branchName],
+        [setupCmd, branchName, worktreePath],
         repo.path,
```
pranav-growthx: **[Should-fix] baseRef silently dropped in delegation path.**

When `worktreeSetupCommand` is set, the `baseRef` parameter is never forwarded to the script. The inline path correctly uses `baseRef ?? repo.defaultBase` (line 56), but the delegation path ignores it entirely -- the script decides its own base (typically `origin/main`).

This is not a regression (main also drops it), but this PR explicitly documents the two-arg contract as the canonical interface. Since `!branch <ref>` sets `session.baseRef` and `session/manager.ts:412` passes it to `createWorktree`, a user who runs `!branch staging` on a repo with a setup script will silently get an `origin/main` worktree.

Suggested fix (pick one):
1. Pass `baseRef` as a third positional arg or env var (`JUNIOR_BASE_REF`) so the script can use it.
2. Add a guard: if both `worktreeSetupCommand` and an explicit `baseRef` are present, throw an error explaining that delegated scripts don't support custom base refs yet.
3. At minimum, document the limitation in the JSDoc above (lines 36-47) so future callers know baseRef is ignored when delegating.

Option 2 is the cheapest safe fix -- it makes the failure loud instead of silent.

src/worktree/manager.ts
```diff
@@ -34,22 +34,26 @@ export class WorktreeManager {
...
+      // No setup hook configured — Junior creates the worktree inline. Fetch
+      // fresh first so the base ref is up to date, then `git worktree add`.
       const base = baseRef ?? repo.defaultBase;
-      // Fetch latest from origin so the base ref is fresh
-      await this.runGit(["fetch", "origin"], repo.path);
-
-      // Create the worktree with a new branch off the base ref
+      await this.runGit(["fetch", "origin", "--prune"], repo.path);
       await this.runGit(
         ["worktree", "add", worktreePath, "-b", branchName, base],
```
pranav-growthx, line 57: **[Nice-to-have]** The `--prune` addition is a good hygiene improvement. Worth a one-line mention in the PR description since it changes behavior (removes stale remote-tracking refs that the old `git fetch origin` preserved). Low risk but the delta should be visible to reviewers of the coordinated set.

pranav-growthx (changes requested):
Two findings -- one should-fix (silent baseRef loss in delegation path) and one nice-to-have (docs). Tests pass, contract is sound, arg-order swap is correct.

pranav-growthx (commented):
## Review: fix(worktree): setup-script delegation contract (re-review after major rework)

### Prior review findings

Both items from the previous review are addressed:
1. **baseRef forwarding** -- now forwarded as `--base <ref>` to setup scripts when explicitly set. Test coverage for both presence and absence of `--base`.
2. **`--prune` documentation** -- PR description now includes a dedicated section.

### New findings

#### No blockers found.

#### Should-fix items noted inline.

#### Cross-repo verification note

I cannot access the coordinated script PRs (gx-backend#3082, gx-admin-client#738, gx-client-next#5007) to verify the real scripts parse `--path` and `--base` with the exact same flag names. Junior's test fixture (`fake-setup.sh`) validates Junior's side of the contract, but if any target script uses different flag names (e.g. `--target` instead of `--path`), this will pass tests and break in production. Whoever merges the coordinated set should confirm the scripts match.

### Summary

The core contract change (arg order swap, flag-based `--path` and `--base`) is correct and well tested. The previous baseRef-dropped-silently bug is fixed. One behavioral side-effect on the dev-server caller warrants a quick look before merge.

src/worktree/manager.ts
```diff
@@ -34,22 +40,30 @@ export class WorktreeManager {
...
       const setupCmd = repo.worktreeSetupCommand.startsWith("/")
         ? repo.worktreeSetupCommand
         : `${repo.path}/${repo.worktreeSetupCommand}`;
-      await this.runCommand(
-        [setupCmd, worktreePath, branchName],
-        repo.path,
-      );
+      const args = [setupCmd, branchName, "--path", worktreePath];
+      if (baseRef !== undefined) {
+        args.push("--base", baseRef);
```
pranav-growthx: **[Should-fix | Behavioral side-effect on existing caller]**

This conditional forwarding is correct for the session-manager caller (`session.baseRef ?? undefined` -- null becomes undefined, `--base` is skipped). But it silently changes behavior for `lifecycle/dev-server.ts:258`, which passes `repo.defaultBase` as an **explicit** `baseRef`:

```ts
await this.worktreeManager.createWorktree(
  repo.name,
  DEV_SERVER_THREAD_ID,
  repo.defaultBase,        // <-- always truthy
  \`dev-server-slot/\${repo.name}\`,
);
```

**Before this PR (delegated path):** `baseRef` was silently dropped. The script used its own per-repo default (e.g. local HEAD for gx-backend).

**After this PR:** `repo.defaultBase` (likely `origin/main`) is forwarded as `--base origin/main` to the script. For repos whose script default differs from `repo.defaultBase`, the dev-server worktree now forks from a different starting point than before.

This might be the **desired** outcome (pin dev-server slot to a known ref). But it might not be -- the PR description says scripts should apply their own per-repo default when `--base` is absent, and this caller forces one.

Suggested: either (a) change dev-server.ts to pass `undefined` so it inherits the script's default (consistent with the new policy), or (b) add a comment in dev-server.ts clarifying that this intentionally pins to `repo.defaultBase`, so future readers know the override is deliberate.

src/worktree/manager.ts
```diff
@@ -34,22 +37,22 @@ export class WorktreeManager {
...
+      // so the script's own HEAD-based fallback is never reached).
       const setupCmd = repo.worktreeSetupCommand.startsWith("/")
         ? repo.worktreeSetupCommand
         : `${repo.path}/${repo.worktreeSetupCommand}`;
-      await this.runCommand(
-        [setupCmd, worktreePath, branchName],
-        repo.path,
-      );
+      const base = baseRef ?? repo.defaultBase;
+      const args = [setupCmd, branchName, "--path", worktreePath, "--base", base];
```
PranavBakre, line 49: **[Blocker] Manager unconditionally forwards `--base` — contradicts the PR description, the config.ts doc-comment, and breaks gx-backend's OG default.**

Lines 48-49:
```ts
const base = baseRef ?? repo.defaultBase;
const args = [setupCmd, branchName, "--path", worktreePath, "--base", base];
```

This means Junior **always** sends `--base <ref>`, defaulting to `repo.defaultBase`. The script's `if [ -n "$BASE_REF" ]` else-branch (case C, fork from local HEAD) is unreachable from Junior. That branch is the OG behavior gx-backend#3082 explicitly preserves: "OG manual behavior is preserved: when no `--base` flag is given, the new worktree is created without an explicit base ... which forks from local HEAD as it did before."

Material impact on gx-backend specifically:
- OG: `setup-worktree.sh feature/foo` forks the new branch from local HEAD.
- Manual user post-merge: same as OG (script preserves it).
- **Junior post-merge: forks from `origin/main`** because `defaultBase` is set in `.env` REPOS. Not local HEAD.

That is a silent, undocumented behavior change for gx-backend Junior worktrees. The PR description says: "`--base <ref>` is forwarded **only when `baseRef` was explicitly passed**" and "gx-backend and gx-admin-client default to local HEAD". The code does neither.

Fix: revert 89a1e7a. The previous commit (937cd84) had the contract right and was matched by a test (`does NOT forward --base when baseRef is unset`) which was deleted in 89a1e7a in favor of one that asserts the opposite.

PranavBakre (commented):
Two blockers, contract drift between code, docs, and PR description. The most recent commit on this branch (89a1e7a `always forward --base to setup scripts`) reverses the contract that the PR description, the config.ts doc-comment, and the implicit gx-backend OG behavior all assume. The tests pass because they were rewritten to match the new always-forward behavior, but the contract drift breaks gx-backend specifically: it'll now fork worktrees from `origin/main` instead of local HEAD, undocumented.

Resolution paths:
  (a) Revert 89a1e7a back to explicit-only forwarding so per-repo script defaults govern the manual ergonomics — what the PR description and config doc say. The previous commit (937cd84) had `if (baseRef !== undefined) args.push("--base", baseRef);` which was the right shape.
  (b) Keep always-forward but update PR body, config.ts doc-comment, manager.ts inline comment, and explicitly call out the gx-backend behavior change (Junior worktrees now fork from origin/main instead of local HEAD) in #3082.

My preference is (a) — that's what every other surface in this PR set assumes, and it preserves OG manual parity that gx-backend#3082 explicitly tries to keep.

src/config.ts
```diff
@@ -4,7 +4,16 @@ export interface RepoConfig {
   name: string;
   path: string;
   defaultBase: string;
-  /** Optional setup script for worktree creation. Run as `<repo.path>/<command> <worktreePath> <branch>`. */
+  /**
+   * Optional. When set, Junior delegates worktree creation to this script via
+   * `<repo.path>/<command> <branch> --path <abs> [--base <ref>]`. The script
+   * must `git fetch`, `git worktree add` at the given path, and perform setup
+   * (env copy, install, MCPs). `--base <ref>` is forwarded only when the
+   * caller passes an explicit `baseRef` to `createWorktree`; when unset, the
```
PranavBakre: **[Blocker] Doc-comment is wrong about `--base` forwarding behavior.**

The comment says:
> `--base <ref>` is forwarded only when the caller passes an explicit `baseRef` to `createWorktree`; when unset, the script applies its own per-repo default (e.g., local HEAD or `origin/main`).

But manager.ts (89a1e7a) actually does `const base = baseRef ?? repo.defaultBase` and **always** appends `--base base` to the args. The script's per-repo default is never consulted from Junior.

Fix together with the manager.ts blocker: pick one behavior and align all three sites (PR body, doc-comment, code).

src/worktree/manager.ts
```diff
@@ -34,22 +37,22 @@ export class WorktreeManager {
     const branchName = branchOverride ?? `slack/${threadId}`;
 
     if (repo.worktreeSetupCommand) {
-      // Delegate to the repo's own setup script: `<repo.path>/<command> <worktreePath> <branch>`
-      // Resolve the command relative to repo.path so paths like "bin/setup-worktree.sh"
-      // work without requiring the script to be on PATH.
+      // Delegate worktree creation to the repo's setup script. The script
+      // owns `git fetch`, `git worktree add`, env-file copying, dependency
+      // install, and MCP migration. Junior hands it the branch, the absolute
+      // target path, and the base ref (always — defaulting to repo.defaultBase
```
PranavBakre, line 43: **[Should-fix] Inline comment also lies.**

> Junior hands it the branch, the absolute target path, and the base ref (always — defaulting to repo.defaultBase so the script's own HEAD-based fallback is never reached).

This is what the code actually does, but it's the opposite of what config.ts line 11-12 claims, and the opposite of what the PR description claims. Whichever path the blocker resolution goes, this comment and the doc-comment must agree.

src/worktree/manager.test.ts
```diff
@@ -95,17 +196,118 @@ describe("WorktreeManager.createWorktree", () => {
...
+    ]);
+
+    // Sanity: the script honored --base, so FEATURE.md (only on
+    // feature/seeded) is present.
+    expect(existsSync(join(wtPath, "FEATURE.md"))).toBe(true);
+
+    await wm.removeWorktree("delegate-baseref", "delegate-baseref-thread");
+  });
+
+  it("forwards repo.defaultBase as --base when baseRef is unset", async () => {
```
PranavBakre, line 259: **[Should-fix] Test pin says one thing, PR description says another.**

This test (`forwards repo.defaultBase as --base when baseRef is unset`) is the right pin for the actual code on 89a1e7a — but the PR description says explicitly: "`--base <ref>` is forwarded **only when `baseRef` was explicitly passed**." The test is correct about the code; the description is correct about the intended contract. They don't match.

Whichever way the blocker is resolved, this test (and the earlier `does NOT forward --base when baseRef is unset` test the prior commit had) need to align with whatever wins.

src/worktree/manager.ts
```diff
@@ -11,12 +11,15 @@ export class WorktreeManager {
...
    * - `baseRef` is the starting point (a git ref/commit like `origin/main`).
-   *   Defaults to `repo.defaultBase`. This becomes the worktree's HEAD.
+   *   Defaults to `repo.defaultBase` on both the inline and delegated paths.
    * - `branchOverride` renames the new branch the worktree tracks. Defaults
    *   to `slack/<threadId>`. The two are independent — pass `branchOverride`
    *   to name the branch differently from the default thread-keyed slug,
    *   pass `baseRef` to fork from a non-main starting point.
    *
+   * Setup-script delegation contract:
+   *   `<repo.path>/<command> <branch> --path <abs> [--base <ref>]`
```
PranavBakre: **[Should-fix | Documentation] JSDoc contract signature implies `--base` is optional; code always sends it.**

Line 21:
```
 *   `<repo.path>/<command> <branch> --path <abs> [--base <ref>]`
```

Square brackets conventionally mean optional, but the code on lines 48-49 unconditionally appends `--base`:
```ts
const base = baseRef ?? repo.defaultBase;
const args = [setupCmd, branchName, "--path", worktreePath, "--base", base];
```

The config.ts JSDoc (line 8: "Junior always forwards `--base <ref>`") and the inline comment (line 43-44: "always -- defaulting to repo.defaultBase") are both correct. This JSDoc contract signature is the only surface that still implies optional.

Fix: drop the brackets to `--base <ref>` (no `[...]`), matching the always-forward behavior.

PranavBakre (commented):
## Review: fix(worktree): setup-script delegation contract (round 3)

### Previous findings status

1. **baseRef silently dropped in delegation path** (round 1 blocker) -- Fixed. The delegation path now always forwards `--base`, defaulting to `repo.defaultBase`. Code, config.ts JSDoc, inline comments, and tests are aligned on the always-forward behavior.
2. **dev-server.ts behavioral side-effect** (round 2 should-fix) -- Non-issue confirmed. Under always-forward, `dev-server.ts` passing `repo.defaultBase` explicitly is functionally identical to passing `undefined` (both resolve to `repo.defaultBase`). The explicit pass documents intent.
3. **Contract drift between code and docs** (round 2 blocker on config.ts) -- Fixed in `aa1c817`. config.ts JSDoc now matches the always-forward code.

### New findings

One should-fix found. See inline comment on `manager.ts:21`.

### Cross-repo note (not a finding)

Same as round 2: the coordinated scripts (gx-backend#3082, gx-admin-client#738, gx-client-next#5007) must parse `--base` with the understanding it is **always** present from Junior. Scripts that branch on `if [ -n "$BASE_REF" ]` to pick a different default will never hit their else-branch when called by Junior. This is intentional per the PR description's "Updated baseRef semantics" section, but whoever merges the set should confirm.

### Summary

The code logic is correct and well-tested. One documentation inconsistency remains in manager.ts's JSDoc contract signature: `[--base <ref>]` (brackets imply optional) vs the code that always sends it. Should-fix before merge so the contract signature matches what consumers and the coordinated scripts should expect.

PR #14, opened 2026-05-06
feat: HTTP dashboard for session + dev-server status

anmolm-growthx:
## Motivation

You asked Friday for this — port her HTTP dashboard over so Junior has a way to inspect live state without grepping logs or poking the SQLite DB by hand. This is that.

## What's in it

A Bun HTTP server that binds **127.0.0.1 only**, off by default. Enable with `HTTP_DASHBOARD_PORT=8787 bun run dev` and visit `http://127.0.0.1:8787`.

The dashboard polls every 2s and renders three panels:

- **Threads** — every `ThreadSession` with status, target repo / base ref, and a nested per-agent breakdown (lead / reproducer / thinker / review / echo) showing each agent's `sessionId`, status, last activity, and pending-message count.
- **Dev Servers** — for each repo with `devCommand` configured: holder thread, branch, idle-TTL countdown, waiter list. Reads `DevServerManager.status()` + `DevServerQueue.readQueueDepth()` directly.
- **Health** — uptime, version, session/agent totals.

## Endpoints

| Path | Description |
| --- | --- |
| `GET /` | Static dashboard (`public/index.html`) |
| `GET /api/health` | Uptime + session/agent totals |
| `GET /api/sessions` | All threads with their agent sessions |
| `GET /api/sessions/:threadId` | Full session detail |
| `GET /api/dev-server` | Per-repo manager state + queue depth |
| `GET /api/logs?date=&tail=&tag=&level=` | Tail of structured logs |
| `GET /api/memory` / `GET /api/memory/<path>` | Browse `docs/**/*.md` |

## What changed

- New `src/http/` (server + 5 route modules; **no SSE chat** — intentionally dropped).
- `src/index.ts` — gated `startHttpServer({...})` block at the end of boot.
- `src/config.ts` — adds `config.http.{enabled,port}`. `enabled` is `process.env.HTTP_DASHBOARD_PORT != null`.
- `src/session/manager.test.ts` — adds the new `http` field to its mock `Config`.
- `public/index.html` — vanilla JS, dark theme, no build step.
- `README.md` — was missing entirely; added with a Dashboard section.

## Constraints honored

- **Localhost only.** `Bun.serve({ hostname: "127.0.0.1" })`. No token auth.
- **Off by default.** Only enabled if `HTTP_DASHBOARD_PORT` is set.
- **No SSE chat.** Friday's `chat-manager.ts` and `routes/chat.ts` were not ported.

## Verification

- `bun test` — 234 pass / 2 fail. Both failures are pre-existing on `main` (`createSession returns all expected keys` and the `DevServerManager (integration) > restarts the server when branch changes` integration test). Verified by running `bun test` on `main` with my changes stashed — same two failures.
- Smoke test: started the server with a seeded SQLite store, hit `/api/health`, `/api/sessions`, `/api/dev-server` — all return the expected shapes including the per-agent join.

## Credit

Ported from Friday's `src/http/`. Junior-specific extensions are the `/api/dev-server` route (which Friday doesn't have) and the agent-sessions join in `/api/sessions` (Friday's threads are single-agent).

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -59,6 +59,11 @@ export interface Config {
     defaultVerbosity: SessionVerbosity;
   };
   channelDefaults: Record<string, { agentType: string }>;
+  http: {
+    /** Localhost dashboard. Off unless HTTP_DASHBOARD_PORT is set. */
+    enabled: boolean;
+    port: number;
+  };
 }
 
 function required(name: string): string {
@@ -105,6 +110,10 @@ export function loadConfig(): Config {
         '{"C05557KKV37":{"agentType":"lead"}}',
       ),
     ),
+    http: {
+      enabled: process.env.HTTP_DASHBOARD_PORT != null,
+      port: Number(optional("HTTP_DASHBOARD_PORT", "0")),
+    },
   };
 }
 
--- a/src/config.ts
+++ b/src/config.ts
@@ -110,11 +110,31 @@ export function loadConfig(): Config {
         '{"C05557KKV37":{"agentType":"lead"}}',
       ),
     ),
-    http: {
-      enabled: process.env.HTTP_DASHBOARD_PORT != null,
-      port: Number(optional("HTTP_DASHBOARD_PORT", "0")),
-    },
+    http: parseHttpDashboard(process.env.HTTP_DASHBOARD_PORT),
   };
+}
+
+/**
+ * Parse HTTP_DASHBOARD_PORT strictly.
+ *
+ * - Unset or empty → dashboard disabled (port 0).
+ * - Positive integer string → enabled on that port.
+ * - Anything else (NaN, 0, negative, decimal) → throw at config load. The
+ *   prior `Number(...)` fallback would silently let `Bun.serve` bind to a
+ *   random port for non-numeric values, which is surprising and makes the
+ *   dashboard unreachable at the configured URL.
+ */
+function parseHttpDashboard(raw: string | undefined): { enabled: boolean; port: number } {
+  if (raw == null || raw === "") {
+    return { enabled: false, port: 0 };
+  }
+  const port = Number(raw);
+  if (!Number.isInteger(port) || port <= 0 || port > 65535) {
+    throw new Error(
+      `Invalid HTTP_DASHBOARD_PORT: ${JSON.stringify(raw)} (expected positive integer 1-65535, or unset to disable)`,
+    );
+  }
+  return { enabled: true, port };
 }
 
 function parseStoreKind(value: string): SessionStoreKind {
--- a/src/http/routes/dev-server.ts
+++ b/src/http/routes/dev-server.ts
@@ -2,14 +2,13 @@ import type { RepoConfig } from "../../config.ts";
 import type { DevServerManager, DevServerState } from "../../lifecycle/dev-server.ts";
 import type { DevServerQueue } from "../../lifecycle/dev-server-queue.ts";
 
-const IDLE_TTL_MS = 20 * 60 * 1_000;
-
 export async function handleDevServers(
   manager: DevServerManager,
   queue: DevServerQueue,
   repos: RepoConfig[],
 ): Promise<Response> {
   const statusMap = manager.status() as Map<string, DevServerState>;
+  const idleTtlMs = manager.getIdleTtlMs();
   const now = Date.now();
 
   const items = await Promise.all(
@@ -21,7 +20,7 @@ export async function handleDevServers(
 
         const idleMsRemaining =
           state?.lastUsedAt != null
-            ? Math.max(0, IDLE_TTL_MS - (now - state.lastUsedAt))
+            ? Math.max(0, idleTtlMs - (now - state.lastUsedAt))
             : null;
 
         return {
@@ -41,5 +40,5 @@ export async function handleDevServers(
       }),
   );
 
-  return Response.json({ devServers: items, idleTtlMs: IDLE_TTL_MS });
+  return Response.json({ devServers: items, idleTtlMs });
 }
--- a/src/http/routes/dev-server.ts
+++ b/src/http/routes/dev-server.ts
@@ -1,0 +1,45 @@
+import type { RepoConfig } from "../../config.ts";
+import type { DevServerManager, DevServerState } from "../../lifecycle/dev-server.ts";
+import type { DevServerQueue } from "../../lifecycle/dev-server-queue.ts";
+
+const IDLE_TTL_MS = 20 * 60 * 1_000;
+
+export async function handleDevServers(
+  manager: DevServerManager,
+  queue: DevServerQueue,
+  repos: RepoConfig[],
+): Promise<Response> {
+  const statusMap = manager.status() as Map<string, DevServerState>;
+  const now = Date.now();
+
+  const items = await Promise.all(
+    repos
+      .filter((r) => r.devCommand)
+      .map(async (repo) => {
+        const state = statusMap.get(repo.name);
+        const queueDepth = await queue.readQueueDepth(repo.name);
+
+        const idleMsRemaining =
+          state?.lastUsedAt != null
+            ? Math.max(0, IDLE_TTL_MS - (now - state.lastUsedAt))
+            : null;
+
+        return {
+          repo: repo.name,
+          devCommand: repo.devCommand ?? null,
+          devPort: repo.devPort ?? null,
+          readyUrl: repo.readyUrl ?? (repo.devPort ? `http://localhost:${repo.devPort}` : null),
+          running: state?.pid != null,
+          pid: state?.pid ?? null,
+          branch: state?.branch ?? queueDepth.holder?.branch ?? null,
+          startedAt: state?.startedAt ?? null,
+          lastUsedAt: state?.lastUsedAt ?? null,
+          idleMsRemaining,
+          holder: queueDepth.holder,
+          waiters: queueDepth.waiters,
+        };
+      }),
+  );
+
+  return Response.json({ devServers: items, idleTtlMs: IDLE_TTL_MS });
+}
--- a/src/http/routes/logs.ts
+++ b/src/http/routes/logs.ts
@@ -1,0 +1,48 @@
+import path from "node:path";
+
+const LOG_DIR = path.resolve(import.meta.dir, "../../../logs");
+
+interface LogEntry {
+  timestamp: string;
+  level: string;
+  tag: string;
+  message: string;
+}
+
+const LOG_LINE_RE = /^(.+?) \[(\w+)\] \[(.+?)\] (.*)$/;
+
+function parseLogLine(line: string): LogEntry | null {
+  const match = line.match(LOG_LINE_RE);
+  if (!match) return null;
+  return {
+    timestamp: match[1],
+    level: match[2],
+    tag: match[3],
+    message: match[4],
+  };
+}
+
+export async function handleLogs(searchParams: URLSearchParams): Promise<Response> {
+  const date = searchParams.get("date") ?? new Date().toISOString().split("T")[0];
+  const tail = Number(searchParams.get("tail") ?? "50");
+  const tag = searchParams.get("tag");
+  const level = searchParams.get("level");
+
+  const logFile = Bun.file(path.join(LOG_DIR, `${date}.log`));
+  if (!(await logFile.exists())) {
+    return Response.json({ date, entries: [] });
+  }
+
+  const content = await logFile.text();
+  const lines = content.trim().split("\n").filter(Boolean);
+  let entries = lines
+    .map(parseLogLine)
+    .filter((e): e is LogEntry => e !== null);
+
+  if (tag) entries = entries.filter((e) => e.tag === tag);
+  if (level) entries = entries.filter((e) => e.level === level);
+
+  const sliced = tail > 0 ? entries.slice(-tail) : entries;
+
+  return Response.json({ date, entries: sliced });
+}
--- a/src/http/routes/logs.ts
+++ b/src/http/routes/logs.ts
@@ -11,6 +11,15 @@ interface LogEntry {
 
 const LOG_LINE_RE = /^(.+?) \[(\w+)\] \[(.+?)\] (.*)$/;
 
+/**
+ * Strict ISO date matcher for the `?date=` query param. The value is concatenated
+ * into a filesystem path (`<LOG_DIR>/<date>.log`), so anything other than a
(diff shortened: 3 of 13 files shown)


src/http/routes/logs.ts
```diff
@@ -0,0 +1,48 @@
...
+  };
+}
+
+export async function handleLogs(searchParams: URLSearchParams): Promise<Response> {
+  const date = searchParams.get("date") ?? new Date().toISOString().split("T")[0];
+  const tail = Number(searchParams.get("tail") ?? "50");
+  const tag = searchParams.get("tag");
+  const level = searchParams.get("level");
+
+  const logFile = Bun.file(path.join(LOG_DIR, `${date}.log`));
```
PranavBakre, line 46: 🔴 **Blocker — path traversal.**

`date` is taken straight from a query param and concatenated into a filesystem path with no validation. `path.join(LOG_DIR, "${date}.log")` happily resolves `../` segments:

```
GET /api/logs?date=../../../etc/passwd
→ reads <repo-parent>/etc/passwd.log
```

Verified locally:
```
LOG_DIR: .../worktrees/agent-…/logs
Joined:  .../worktrees/.claude/etc/passwd.log
```

The `.log` suffix limits impact to files that end in `.log`, but combined with the CORS-wildcard issue (separate comment on `server.ts`), any website the operator visits can scan for and exfiltrate `*.log` files anywhere the bot user has read access.

**Fix:** validate `date` matches `/^\d{4}-\d{2}-\d{2}$/` before using it. Same shape `memory.ts` uses — reject on `..` or `/`, then `path.resolve` and re-check `startsWith(LOG_DIR + path.sep)`.

src/http/server.ts
```diff
@@ -0,0 +1,108 @@
...
+import { handleLogs } from "./routes/logs.ts";
+import { handleMemoryList, handleMemoryRead } from "./routes/memory.ts";
+import { handleDevServers } from "./routes/dev-server.ts";
+import { log } from "../logger.ts";
+
+const PUBLIC_DIR = path.resolve(import.meta.dir, "../../public");
+const startedAt = new Date().toISOString();
+
+function cors(res: Response): Response {
+  res.headers.set("Access-Control-Allow-Origin", "*");
```
PranavBakre: 🔴 **Blocker — CORS wildcard breaks the loopback-only threat model.**

The docstring at the top of this file says "localhost-only by design (binds 127.0.0.1, no auth)". Loopback binding stops *remote* attackers, but `Access-Control-Allow-Origin: *` lets *any website the operator visits* execute `fetch("http://127.0.0.1:8787/api/sessions")` from JavaScript and read the response. Same-origin policy normally blocks that read; the wildcard explicitly opts out of it.

What leaks: thread IDs, channel IDs, repo names, branch names, target users, lead `systemPrompt` (via `/api/sessions/:threadId`), log lines (via `/api/logs`), and any `docs/**/*.md` content (via `/api/memory/<path>`). For Junior specifically, `systemPrompt` may include operator notes / channel context that shouldn't go to evil.example.com.

**Fix:** drop the `cors()` helper entirely — the dashboard is same-origin, so CORS adds nothing but attack surface. If you want belt-and-suspenders against DNS rebinding, also reject requests whose `Host:` header isn't `127.0.0.1:<port>` or `localhost:<port>`.

src/http/routes/dev-server.ts
```diff
@@ -0,0 +1,45 @@
+import type { RepoConfig } from "../../config.ts";
+import type { DevServerManager, DevServerState } from "../../lifecycle/dev-server.ts";
+import type { DevServerQueue } from "../../lifecycle/dev-server-queue.ts";
+
+const IDLE_TTL_MS = 20 * 60 * 1_000;
```
PranavBakre: 🟡 **Should-fix — duplicated default, ignores configured override.**

`IDLE_TTL_MS = 20 * 60 * 1_000` is the *default* in `DevServerManager` (src/lifecycle/dev-server.ts:43), but the manager accepts an `idleTtlMs` option that overrides it. This route hardcodes the default, so if a future caller passes a different TTL, the dashboard's countdown will be wrong without anyone noticing.

**Fix:** expose the TTL via `manager.idleTtlMs` (add a getter) and read from there, or include `idleTtlMs` in the per-state object returned by `manager.status()`. Single source of truth.

src/config.ts
```diff
@@ -94,6 +99,10 @@ export function loadConfig(): Config {
         '{"C05557KKV37":{"agentType":"lead"}}',
       ),
     ),
+    http: {
+      enabled: process.env.HTTP_DASHBOARD_PORT != null,
+      port: Number(optional("HTTP_DASHBOARD_PORT", "0")),
```
PranavBakre: 🟡 **Should-fix — silent fallback to a random port on bad input.**

`Number(optional("HTTP_DASHBOARD_PORT", "0"))` returns `NaN` for non-numeric values and `0` for an empty string. `Bun.serve({ port: NaN })` and `{ port: 0 }` both bind to a *random available port* without erroring (verified locally — `HTTP_DASHBOARD_PORT=abc` ended up on port 62641). Operator types `=8788x` or `=`, the dashboard "runs", but on a port they can't predict.

**Fix:** validate before constructing the config:
```ts
const rawPort = process.env.HTTP_DASHBOARD_PORT;
if (rawPort != null) {
  const port = Number(rawPort);
  if (!Number.isInteger(port) || port < 1 || port > 65535) {
    throw new Error(`Invalid HTTP_DASHBOARD_PORT: ${rawPort}`);
  }
}
```
Matches the strictness pattern of `parseStoreKind` / `parseChannelDefaults` already in this file.

PranavBakre (changes requested):
**Retraction note:** the prior version of this review claimed a "silent revert" blocker. That was wrong — the reviewer compared against a stale local `main` (one commit unpushed-ahead of remote `main`), so the "missing" commits were artifacts of the local-only state, not anything this branch reverted. Even when a branch is genuinely behind main, a normal merge preserves main's commits via the merge base — "behind main" is a routine state, not a revert. Apologies for the noise.

Re-reviewing now against current remote `main`. The four inline findings below stand on their own merits:

- 🔴 `/api/logs?date=…` path traversal — see inline comment on `routes/logs.ts`.
- 🔴 `Access-Control-Allow-Origin: *` defeats the loopback-only threat model — see inline comment on `server.ts`.
- 🟡 Hardcoded `IDLE_TTL_MS` in dev-server route — see inline comment.
- 🟡 Silent random-port fallback on bad `HTTP_DASHBOARD_PORT` — see inline comment.

A new review will follow once it's run against current `main`.

src/http/routes/sessions.ts
```diff
@@ -0,0 +1,48 @@
...
+      activeAgentName: s.activeAgentName ?? null,
+      targetRepo: s.targetRepo,
+      baseRef: s.baseRef,
+      agentType: s.agentType,
+      verbosity: s.verbosity,
+      muted: s.muted,
+      model: s.model,
+      lastActivity: s.lastActivity,
+      createdAt: s.createdAt,
+      pendingMessages: s.pendingMessages.length,
```
PranavBakre: 🟡 **Should-fix (post-rebase) — `/api/sessions` will silently drop the new `defaultAgent` field.**

Against this PR's base (`bfaa440`), `ThreadSession` didn't have a `defaultAgent` field — so this route's hand-written field projection was complete at the time it was written. Not your fault.

But current `personal/main` (commit `f01b3df`, the `!agent` thread-default-agent feature) added:

```ts
// src/session/types.ts
/** Thread-specific default agent override. When set, overrides channel default... */
defaultAgent?: "junior" | "lead" | null;
```

This route enumerates fields explicitly (`threadId`, `channel`, `status`, …) so once you rebase, `defaultAgent` won't appear in the response. Operators using the dashboard to introspect why a thread is routing to `lead` instead of `junior` (or vice-versa) won't see the override.

I ran `git rebase personal/main` in a worktree — clean rebase, no conflicts, but this is the silent gap.

**Fix:** add `defaultAgent: s.defaultAgent ?? null,` to the projection in this map after rebasing. Consider whether `handleSessionDetail` (which spreads the whole session) is enough alternatively, but the list view is what most operators look at.

src/index.ts
```diff
@@ -181,5 +181,16 @@ setInterval(() => {
...
 
+  if (config.http.enabled) {
+    const { startHttpServer } = await import("./http/server.ts");
+    startHttpServer({
+      store,
+      config,
+      devServerManager,
+      devServerQueue,
+      repos: config.repos,
+    });
```
PranavBakre: 🟡 **Should-fix — dashboard boot failure crashes the bot.**

If `Bun.serve({ port: 8787 })` throws synchronously (port already in use, e.g. another Junior instance running, or a leftover process on the same port), the call here propagates out of the async IIFE that wraps the boot block. Bun treats unhandled promise rejections as fatal → the bot exits before the `"Junior is running"` log line, and Slack connectivity is silently lost.

The dashboard is documented as optional / off-by-default. A failure to start it shouldn't take Slack down with it.

**Fix:**

```ts
if (config.http.enabled) {
  try {
    const { startHttpServer } = await import("./http/server.ts");
    startHttpServer({ store, config, devServerManager, devServerQueue, repos: config.repos });
  } catch (err) {
    log.error("boot", `HTTP dashboard failed to start: ${err instanceof Error ? err.message : String(err)}`);
  }
}
```

Matches the same defensive pattern already used a few lines up for `devServerManager.bootstrap()` (line 152).

src/config.ts
PranavBakre: Reconfirming this still applies on the current branch HEAD (`79d2dea`). One additional edge case for the same fix:

`HTTP_DASHBOARD_PORT=` (empty string) sets `enabled: process.env.HTTP_DASHBOARD_PORT != null` → **true**, then `Number(optional("HTTP_DASHBOARD_PORT", "0"))` → `Number("") = 0` → silent random port. Empty string in env files is a common operator mistake (commenting out a value but leaving the `=`). When you address this finding, prefer parsing the value once, treating empty string and `NaN` the same as unset, and only enable the dashboard when an explicit valid integer is provided.

PranavBakre (changes requested):
## Re-review (post-retraction)

Acknowledging the prior review's retraction at the top of #4235268691 — the "silent revert" claim was wrong, caused by comparing against a stale local `main`. Not re-flagging it.

This pass verified the four still-standing inline findings against current branch HEAD (`79d2dea`), and added two new ones grounded in interactions with main commits that landed after the PR was opened.

### Verified — prior findings still apply (commits unchanged, no fixes pushed)

- 🔴 `src/http/routes/logs.ts:31` — path traversal on `?date=`
- 🔴 `src/http/server.ts:26` — `Access-Control-Allow-Origin: *` breaks loopback-only threat model
- 🟡 `src/http/routes/dev-server.ts:5` — hardcoded `IDLE_TTL_MS`, ignores `DevServerManager` config
- 🟡 `src/config.ts:104` — silent random-port fallback on bad input (replied on that thread with an additional empty-string edge case)

### New findings (post-rebase / second-pass)

- 🟡 `src/http/routes/sessions.ts:22` — after rebasing onto `personal/main` (`f01b3df`), the new `ThreadSession.defaultAgent` field will silently be missing from `/api/sessions`. The route hand-projects fields, so it needs to be added. Rebase itself is clean (no conflicts), I tried it.
- 🟡 `src/index.ts:192` — `startHttpServer` is not wrapped in try/catch. `Bun.serve` failures (port in use) propagate out of the async IIFE → unhandled rejection → bot exits before completing boot. The dashboard is optional; it shouldn't be able to take Slack down.

### Cross-check items the prior review didn't cover

- **Mock Configs on main.** Only one `Config = { … }` literal exists across the repo (`src/session/manager.test.ts`), and the branch updated it. No others to chase.
- **Path traversal on `/api/memory`.** The `..` literal substring check + `path.resolve` + `startsWith(DOCS_DIR)` is sufficient (`..` blocks the sibling-dir bypass that `startsWith` alone would miss). OK.
- **Polling cadence.** Single `setInterval` doing `Promise.all` over 3 endpoints — coordinated, not racing. OK.
- **Hardcoded agent list.** Dashboard renders agents dynamically from the response; not hardcoded. OK.
- **Error response leaks.** Errors logged with `${err}` (good for debugging) but responses are generic. OK.
- **Rate limiting.** None, but loopback-only — fine for the threat model.
- **Rebase conflicts.** `git rebase personal/main` succeeds with zero conflicts.

### Test results on branch HEAD

`bun test` → **235 pass / 1 fail**. Failing test is `createSession > returns all expected keys` in `src/session/types.test.ts`, which the PR did not touch — pre-existing failure on main, post-`!agent` change. PR description claims "234 pass / 2 fail" — minor drift, not a blocker.

`bun run typecheck` → clean.

### Verdict

**Request changes.** Two unresolved 🔴 (path traversal, CORS), two unresolved 🟡 from prior review (idle TTL, port fallback), two new 🟡 from this pass (`defaultAgent` missing, dashboard crash). Six total open. No clean passes yet.

PranavBakre:
@anmolm-growthx — pushed review fixes to your `feat/http-dashboard` branch (rebased onto current `main`). Six commits, one per finding:

- `bd18bfc` fix(http): reject path traversal in `/api/logs` date param — strict `^\d{4}-\d{2}-\d{2}$` regex, returns 400 otherwise. Adds `src/http/routes/logs.test.ts`.
- `fc16daf` fix(http): drop CORS wildcard — the dashboard is same-origin, so no CORS headers are needed at all. Removed the `cors()` helper.
- `a67eb27` fix(http): read idle TTL from `DevServerManager` instead of hardcoding — added `getIdleTtlMs()` accessor on the manager and have the route use it.
- `22e7566` fix(http): require positive integer `HTTP_DASHBOARD_PORT` or disable — silently parsing `NaN` to a random port is now a hard error at config-load.
- `0c669cd` fix(http): surface `defaultAgent` in `/api/sessions` — switched from explicit field list to a spread-based projection that omits internal fields (`worktreePath{,s}`, `cwd`, `pid`, `systemPrompt`, `slackIdentity`) so future `ThreadSession` additions flow through automatically.
- `011c44b` fix(http): handle `Bun.serve` startup failures without crashing the bot — wrapped `startHttpServer` in a try/catch matching the `devServerManager.bootstrap()` pattern.

Verification:
- `bun run typecheck` clean.
- `bun test` — 243 pass, 1 fail. The single failure (`createSession > returns all expected keys` in `src/session/types.test.ts`) is pre-existing on `main` and unrelated to this PR (test missing the `muted` key from when verbosity/mute landed).

Inline review threads can be marked resolved once you confirm the fixes look right.

PranavBakre (approved):
All six findings fixed in bd18bfc..011c44b. Pre-existing muted-key test failure fixed on main in dbed947. Approving.

PR #15, opened 2026-05-07
fix(worktree): move worktree path out of .claude/ to a sibling directory

PranavBakre:
## Summary
- Move worktree path from `<repo>/.claude/worktrees/slack-<thread>` to `<repo>.junior-worktrees/slack-<thread>` (sibling to the repo, not under `.claude/`).
- Update the duplicate path-construction site in `DevServerQueue.getLockDir` so locks land at the same place as the worktree.
- Tests + doc comments updated; `afterAll` teardown extended to clean the new sibling directory.

## Why
Target repos' `setup-worktree.sh` recursively copies `.claude/` into a freshly-creating worktree (e.g., gx-client-next did `cp -R .claude/. <new-worktree>/.claude/`). Because junior's worktrees lived under `.claude/worktrees/`, the copy source contained the destination plus every sibling thread's full source tree — and `cp -R` recursed into the destination itself. Observed runaway: 100 GB of nested duplicates after 7 minutes at 83% CPU on gx-client-next before npm install ever started.

Two clashing decisions were producing the bug together: (1) junior nested its worktrees inside an in-repo dir, (2) the setup scripts did unfiltered recursive copies of that dir. Either decision alone was fine; together they pathologically interacted. The setup scripts in all three GX repos have been patched separately (`rsync -a --exclude='worktrees' .claude/ dst/.claude/`) as defense-in-depth. This PR is the durable architectural fix on junior's side: worktrees now live outside any in-repo dir a setup script could walk.

## Side effects
- Existing worktrees at the old `<repo>/.claude/worktrees/slack-*` paths become orphans. Junior won't find them on next boot.
  - Dev-server bootstrap recreates fresh at the new path — no continuity needed.
  - Active thread sessions persisted in SQLite reference old paths — `--resume` continuity breaks for those threads until they're re-bootstrapped.
- The pre-existing flaky test `DevServerManager (integration) > ensure() restarts the server when branch changes` still fails on `main` without these changes (the failing test relies on a `fake-server.js` file that's not on the `fix/test` branch in the test fixture). Not addressed in this PR.

## Test plan
- [x] `bun test` — 244/245 pass; the one failure is the pre-existing flake unrelated to this change
- [ ] Manual: stop junior, delete any old `.claude/worktrees/slack-*` dirs in target repos, restart junior, confirm new worktrees are created at `<repo>.junior-worktrees/slack-dev-server` and that the setup script's `rsync` copy completes cleanly
- [ ] Manual: confirm `npm install` succeeds against `npm.pkg.github.com` (depends on the direnv-export fix in target repos' setup-worktree.sh, also pushed to `main` in each repo)

### DIFF
--- a/src/worktree/manager.ts
+++ b/src/worktree/manager.ts
@@ -147,7 +147,12 @@ export class WorktreeManager {
     if (!repo) {
       throw new Error(`Unknown repo: ${repoName}`);
     }
-    return `${repo.path}.junior-worktrees/slack-${threadId}`;
+    // Strip trailing slashes — without this, a config with `path: "/r/"` would
+    // resolve to `/r/.junior-worktrees/...`, a hidden subdir INSIDE the repo
+    // rather than a sibling, recreating the recursive-copy bug. Belt-and-
+    // suspenders with the same normalization at config load.
+    const base = repo.path.replace(/\/+$/, "");
+    return `${base}.junior-worktrees/slack-${threadId}`;
   }
 
   getBranchName(threadId: string): string {
--- a/src/worktree/manager.ts
+++ b/src/worktree/manager.ts
@@ -135,13 +135,19 @@ export class WorktreeManager {
 
   /**
    * Get the worktree path for a thread (without creating it).
+   *
+   * Worktrees live in a sibling directory to the repo, NOT under `.claude/`.
+   * Setup scripts that recursively copy `.claude/` (e.g. `cp -R .claude/.`)
+   * would otherwise pull every sibling thread's worktree — and the destination
+   * itself — into a freshly-creating worktree, producing a recursive copy
+   * that loops on its own destination.
    */
   getWorktreePath(repoName: string, threadId: string): string {
     const repo = this.getRepo(repoName);
     if (!repo) {
       throw new Error(`Unknown repo: ${repoName}`);
     }
-    return `${repo.path}/.claude/worktrees/slack-${threadId}`;
+    return `${repo.path}.junior-worktrees/slack-${threadId}`;
   }
 
   getBranchName(threadId: string): string {
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ Implementation: [`src/lifecycle/dev-server.ts`](src/lifecycle/dev-server.ts), [`
 
 ### Worktrees per (repo, thread)
 
-Each thread that touches a target repo gets its own git worktree under `<repo>/.claude/worktrees/slack-<threadId>`. A thread can have worktrees in multiple repos at once (full-stack bug → backend + frontend). See [`src/worktree/manager.ts`](src/worktree/manager.ts) and the `worktreePaths` field on `ThreadSession`.
+Each thread that touches a target repo gets its own git worktree at `<repo>.junior-worktrees/slack-<threadId>` (sibling to the repo, deliberately outside `.claude/` so target-repo setup scripts that recursively copy `.claude/` don't pull every sibling thread's source tree into a fresh worktree). A thread can have worktrees in multiple repos at once (full-stack bug → backend + frontend). See [`src/worktree/manager.ts`](src/worktree/manager.ts) and the `worktreePaths` field on `ThreadSession`.
 
 ### SQLite persistence
 
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -185,7 +185,7 @@ Slack message ("!build fix auth")
   → Worktree Manager: create worktree in example-backend
   → Agent Router: load example-backend/.claude/agents/build.md → compose systemPrompt
   → Claude Spawner: spawn claude -p "fix auth" --resume --append-system-prompt "..." 
-      cwd=example-backend/.claude/worktrees/slack-<threadId>
+      cwd=example-backend.junior-worktrees/slack-<threadId>
   → Stream Parser: parse stdout events
       → init event: extract sessionId, store in session
       → tool_use events: emit to Stream-to-Slack → edit status message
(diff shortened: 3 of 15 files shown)


PranavBakre (commented):
Path fix is architecturally sound and the duplicate-construction site in `DevServerQueue.getLockDir` is correctly updated. Tests pass (244/245 — confirmed the one failure is the pre-existing flake unrelated to this PR). Two should-fix items below.

---

## 🟡 Should-fix — stale docs describe the old path shape

Several user-facing docs still describe `<repo>/.claude/worktrees/slack-<threadId>` and now contradict the source:

- `README.md:59` — top-level architecture description.
- `docs/architecture.md:188` — `cwd=example-backend/.claude/worktrees/slack-<threadId>`.
- `docs/features/worktree-manager.md:52` — `createWorktree` description.
- `docs/features/process-lifecycle.md:130, :141` — dev-server worktree + lock dir.
- `docs/features/thread-commands.md:60` — `!status` output sample.
- `docs/features/project-setup.md:161` — `.gitignore` mention (also irrelevant now since the new path is a sibling, not in-repo).

CLAUDE.md emphasizes "Read docs first — don't explore the codebase when a doc answers the question." Future readers (human or Claude itself) following these docs will look in the wrong place. Worth a one-shot sweep while the change is fresh.

## 🟢 Nit — dead `lockDir` in compromise-check test

`src/lifecycle/dev-server-queue.test.ts:441` still uses `join(repoRoot, ".claude", "worktrees", "slack-dev-server-compromise-check")`. The lockDir is created and never used — the real lock is acquired at `realLockDir` (line 468). Harmless (dir lands inside the tmpdir which is wiped in `afterAll`), but the leftover `mkdirSync` is dead code that the path-literal sweep missed. Not a blocker — flagging only because the harmlessness depends on the dead-code property holding.

## Side effects acknowledged in PR description

The PR notes that existing on-disk worktrees at `<repo>/.claude/worktrees/slack-*` become orphans (junior won't find them) and SQLite-persisted thread sessions reference old paths so `--resume` continuity breaks. Since `removeWorktree` is only invoked from tests in the current code (no production callsite — pre-existing), there's no automatic on-disk cleanup either way. Operator will need to manually `rm -rf <repo>/.claude/worktrees/slack-*` post-deploy. Worth mentioning explicitly in the deploy runbook / release notes if those exist.

src/worktree/manager.ts
```diff
@@ -135,13 +135,19 @@ export class WorktreeManager {
...
+   * itself — into a freshly-creating worktree, producing a recursive copy
+   * that loops on its own destination.
    */
   getWorktreePath(repoName: string, threadId: string): string {
     const repo = this.getRepo(repoName);
     if (!repo) {
       throw new Error(`Unknown repo: ${repoName}`);
     }
-    return `${repo.path}/.claude/worktrees/slack-${threadId}`;
+    return `${repo.path}.junior-worktrees/slack-${threadId}`;
```
PranavBakre: 🟡 **Should-fix — trailing-slash recurrence vector.** `${repo.path}.junior-worktrees/...` produces a sibling dir only when `repo.path` has no trailing slash. With a trailing slash (e.g. `"/Users/x/gx-backend/"` from a tab-completed shell paste), the result becomes `"/Users/x/gx-backend/.junior-worktrees/slack-..."` — INSIDE the repo, recreating exactly the bug this PR fixes (any `setup-worktree.sh` that recursively copies the repo dir would walk straight into it).

Current operator config in `.env` and `.env.example` happen to omit the trailing slash so it doesn't bite today — but the regression vector is one config edit away, and the failure mode is the 100GB blowup that motivated this PR. Cheap fix at the source:

```ts
return `${repo.path.replace(/\/+$/, "")}.junior-worktrees/slack-${threadId}`;
```

Or better, normalize once at config load in `loadConfig()` so every consumer of `repo.path` is safe (the matching site `dev-server-queue.ts:333` has the same vulnerability and would need the same fix here).

PranavBakre:
## Re-review verdict: approved (two consecutive clean passes)

GitHub blocks self-approval, but per protocol this is the approval signal.

### Prior items — verified fixed

- 🟡 **Trailing-slash defense** — belt-and-suspenders: normalized once at config load (`src/config.ts:94-101`) AND at both path-construction sites (`src/worktree/manager.ts:154`, `src/lifecycle/dev-server-queue.ts:334`). New regression test at `src/worktree/manager.test.ts:428` directly asserts the failure mode (`path: "/r/"` → `/r.junior-worktrees/...`, NOT the in-repo `/r/.junior-worktrees/...` shape).
- 🟡 **Doc sweep** — all 8 current-state references updated. Correctly skipped the 2026-04-08 audit snapshot (historical document). The `docs/features/project-setup.md:161` clarification (noting the `.gitignore` entry now serves OTHER tooling like ultrareview) goes beyond literal find-and-replace.
- 🟢 **Dead `lockDir` mkdir** in compromise-check test — removed.

### Second clean pass — fresh look

- **Logic.** Regex `/\/+$/` correctly handles 0, 1, or N trailing slashes. The only degenerate case is `path: "/"` → `""`, but that config is nonsensical and would fail elsewhere first.
- **Cross-file consistency.** `getLockDir` doc comment now cross-references `getWorktreePath` so the duplicate-site relationship is documented for future readers.
- **Other `repo.path` consumers** (`src/agents/router.ts:24,47`, the various `runGit` cwds) either interpose a literal `/` (where trailing slashes collapse harmlessly) or use it as a cwd (where POSIX accepts trailing slashes). Not affected.
- **Test coverage.** Regression guard tests the exact failure mode the fix targets — not just a generic "trailing slash works" assertion.

### Test verification

`bun run typecheck` clean. Targeted runs of `src/worktree/manager.test.ts` and `src/lifecycle/dev-server-queue.test.ts` — 38/38 pass, including the new regression guard and the previously-flagged onCompromised test (now without the dead `lockDir` mkdir). The one failing test on a full `bun test` run is the pre-existing `DevServerManager > ensure() restarts the server when branch changes` flake noted in the PR description — reproduces on `main` without these changes, not introduced here.

No new findings on the second pass. Two consecutive clean passes — approved.

PR #16, opened 2026-05-07
fix(dev-server): write lock/queue ignores to info/exclude, not .gitignore

PranavBakre:
## Summary
- Bootstrap was overwriting the dev-server worktree's `.gitignore` with `.lock*\n.queue*\n`, clobbering the upstream-tracked file. Every previously-ignored file (node_modules/, .env*, .envrc, .mcp.json, *.pem, .DS_Store) surfaced as untracked, plus `M .gitignore` itself.
- Lock/queue files are operational artifacts, not source-tree concerns. They belong in per-worktree `info/exclude` (resolved via `git rev-parse --git-dir` — for `git worktree add`, the gitdir is at `<main>/.git/worktrees/<name>/`).
- `.gitignore` is shared-owner (upstream repo + user + tools all contribute); never overwrite. `info/exclude` is single-owner (this clone); overwrite is fine.

## Test plan
- [x] `bun test` — 245/246 pass (the one failure is a pre-existing flake on `main`, unrelated)
- [x] New bootstrap assertions: `info/exclude` contains `.lock*` and `.queue*`, AND `.gitignore` is NOT modified relative to HEAD after bootstrap (regression guard for the bug being fixed)
- [ ] Manual: stop junior, restart, confirm new dev-server worktrees boot with `git status` clean (zero `M .gitignore`, zero spurious untracked from previously-ignored upstream patterns)

## Operator note
Existing dev-server worktrees that already ran the buggy bootstrap need a one-time `git -C <wt> checkout -- .gitignore` to restore upstream content. `info/exclude` is created fresh on next boot under the new code path.

### DIFF
--- a/src/lifecycle/dev-server.test.ts
+++ b/src/lifecycle/dev-server.test.ts
@@ -23,7 +23,7 @@ import {
   readFileSync,
 } from "node:fs";
 import { tmpdir } from "node:os";
-import { join } from "node:path";
+import { isAbsolute, join } from "node:path";
 import type { RepoConfig } from "../config.ts";
 import { WorktreeManager } from "../worktree/manager.ts";
 import { DevServerManager, resolveReadyUrl } from "./dev-server.ts";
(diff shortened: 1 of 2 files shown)


PranavBakre (commented):
Two consecutive clean passes — would approve, but GitHub blocks self-approval. Leaving as a comment.

All prior-pass items addressed in `b699eeb`:
- 🟡 `runGit` swap — clean, error path consistent with the existing helper.
- 🟢 Self-heal with exact-match fingerprint — narrow, safe, idempotent. Removes the operator-action requirement noted in the PR description.
- 🟢 Test split — single-concern `it()` blocks; idempotency now also asserts `.gitignore` cleanliness on the second bootstrap; new self-heal test plants the bug fingerprint and verifies recovery.
- 🟢 Mirror of `isAbsolute` normalization in tests — production parity.

Typecheck clean. All 5 bootstrap tests pass.

Two minor 🟢 nice-to-haves (non-blocking, can defer):
- `src/lifecycle/dev-server.ts:319-323` — inner `catch {}` on `git checkout -- .gitignore` swallows all errors silently. A `log.debug` inside the inner catch would leave a breadcrumb if a transient (non-"untracked") failure ever masks recovery.
- `src/lifecycle/dev-server.test.ts:226-255` — the self-heal test fixture doesn't commit an upstream `.gitignore`, so the test only exercises the "unlink + checkout fails harmlessly" path. The production-common path (upstream tracks `.gitignore`, checkout restores from HEAD) isn't directly verified. A variant that commits a non-bug `.gitignore` upstream before planting the corrupt content would close the gap.

Verdict: ready to merge.

PR #17, opened 2026-05-08
Admin-gate destructive commands + filter sibling-bot ✽ noise

PranavBakre:
## Summary

Two related-but-distinct hardening changes to message handling, in separate commits:

1. **`fix(events): drop messages starting with ✽`** — Sibling Claude bots (Friday, Doraemon) post their streaming "thinking" updates with a leading ✽ glyph. Junior was treating those as user input. New guard at the event boundary drops them; foreign bots' real messages still flow through.
2. **`feat(commands): admin-gate !mute/!unmute/!reset and require !reset arg`**
   - New env var `ADMIN_SLACK_USER_ID` (single Slack user). When set, only that user can run `!mute`, `!unmute`, `!reset`. Non-admin → ❌ reaction, no thread reply, no log noise. Unset → gate is open (local dev).
   - `!reset` now requires an arg. `!reset all` is the old full-session wipe. `!reset <agent>` clears just that agent's slice (lead lives on the parent `ThreadSession`; others in `agentSessions`). Bare `!reset` is a usage error so admins don't accidentally nuke an in-flight reproducer/thinker turn.

Per-agent `!mute` and a richer thread-owner-based access model were considered and explicitly scoped to v2 — see `docs/features/v2-backlog.md`.

## Test plan
- [ ] Set `ADMIN_SLACK_USER_ID=<your-id>`, restart bot. As admin: `!mute`, `!unmute`, `!reset all` work. As non-admin: ❌ reacts, no reply, session unchanged.
- [ ] Unset `ADMIN_SLACK_USER_ID`: any user can run all three (local-dev fallback).
- [ ] Bare `!reset` → reply with usage, no session change.
- [ ] `!reset lead` clears lead-side state (`sessionId`, `leadSessionId`, pending) but leaves `agentSessions` entries intact.
- [ ] `!reset reproducer` deletes `agentSessions.reproducer`, leaves lead state intact.
- [ ] `!reset nonexistent` → "Nothing to reset", no state change.
- [ ] Paste a message starting with `✽` from another bot in a watched channel — Junior logs `dropped reason=foreign-bot-thinking` and ignores it.

### DIFF
--- a/src/session/manager.test.ts
+++ b/src/session/manager.test.ts
@@ -97,6 +97,7 @@ const testConfig: Config = {
     defaultVerbosity: "quiet",
   },
   channelDefaults: {},
+  adminSlackUserId: null,
   http: { enabled: false, port: 0 },
 };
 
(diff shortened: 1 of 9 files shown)


src/session/manager.ts
```diff
@@ -188,8 +243,29 @@ export class SessionManager {
...
-        this.onCommandResponse?.(event, "Session reset.");
+        const arg = event.text.trim();
+        if (!arg) {
+          this.onCommandResponse?.(
+            event,
+            "Usage: `!reset <agent>` (e.g. `!reset lead`, `!reset reproducer`) or `!reset all` to clear the entire thread session.",
+          );
+          return true;
+        }
+        if (this.denyIfNotAdmin(event)) return true;
```
PranavBakre: **Should-fix (Product Thinking / Authorization):** `!cancel` (line 221, unchanged in this PR) is not admin-gated, but it kills running processes and clears `pendingMessages` for all agents in the thread. The PR's stated goal is to prevent non-admins from accidentally nuking in-flight reproducer/thinker turns -- `!cancel` does exactly that.

Granted, `!cancel` doesn't delete state (no sessionId wipe, no worktree cleanup) so it's weaker than `!reset`. But a non-admin can still disrupt an active agent turn and discard queued messages. The PR description doesn't mention `!cancel` as an explicit scoping decision.

Suggested fix: add `if (this.denyIfNotAdmin(event)) return true;` at the top of the `cancel` case, or document why it's intentionally left open (e.g. "anyone can interrupt, only admins can wipe").

PranavBakre (commented):
Four findings -- one should-fix, three nice-to-have. The code is solid overall but the `!cancel` / admin-gate gap deserves attention before merge.

src/session/manager.ts
```diff
@@ -158,6 +159,60 @@ export class SessionManager {
...
+      if (session.sessionId || session.leadSessionId || session.pendingMessages.length > 0 || session.status !== "idle") {
+        touched = true;
+      }
+      session.sessionId = null;
+      session.leadSessionId = null;
+      session.pendingMessages = [];
+      session.status = "idle";
+      session.pid = null;
+    } else if (session.agentSessions?.[agentName]) {
+      delete session.agentSessions[agentName];
```
PranavBakre, line 192: **Nice-to-have (Data Integrity / Race condition):** `resetAgent` deletes `agentSessions[agentName]` here, but the killed process's `onRunComplete` fires asynchronously and calls `getOrCreateAgentSession(fresh, agentName)` (line 688, unchanged) -- which **recreates** the deleted entry with default state plus the old `sessionId`. The reset is silently undone.

Trace:
1. `!reset reproducer` -> `handle.kill()`, `delete agentSessions.reproducer`, persist.
2. Killed process resolves -> `onRunComplete` -> refetches session (no reproducer entry) -> `getOrCreateAgentSession` recreates it -> sets `agentSession.sessionId = result.sessionId` -> persists.
3. Next `!reproducer` call `--resume`s the conversation the user just tried to reset.

This is a pre-existing issue (the `resetSession` path has a similar concern, acknowledged at line 684), so it's not a blocker for this PR. But since `resetAgent` is new code, consider guarding in `onRunComplete`: check whether the agent entry still exists after refetch, and skip the write if it was deleted between spawn and completion.

src/session/manager.ts
```diff
@@ -188,8 +243,29 @@ export class SessionManager {
       }
 
       case "reset": {
```
PranavBakre, line 245: **Nice-to-have (Product Thinking):** The empty-arg check at line 247 runs *before* `denyIfNotAdmin` at line 254. A non-admin typing bare `!reset` gets the usage help reply instead of the silent x-reaction that `!reset all` or `!reset lead` would give them.

The PR description says non-admin invocations produce "no thread reply, no log noise" -- this path contradicts that claim. Consider swapping the order: check admin first, then validate the arg.

src/session/manager.test.ts
```diff
@@ -297,28 +298,132 @@ describe("SessionManager", () => {
...
+        await manager.handleMessage(makeEvent({ user: "U-ANYONE", text: "go" }));
+
+        await manager.handleMessage(
+          makeEvent({ user: "U-ANYONE", command: "mute", text: "", ts: "ts-mute" }),
+        );
+
+        expect(onReaction).not.toHaveBeenCalled();
+        expect((await store.get("thread-1"))!.muted).toBe(true);
+      });
+    });
```
PranavBakre, line 487: **Nice-to-have (Test coverage):** Two gaps vs. the PR's test plan:

1. No test for `!unmute` admin gating (only `!mute` is tested at line 398). Symmetric but easy to add.
2. `!reset lead` clearing parent state (`sessionId`, `leadSessionId`, `pendingMessages`, `status`) isn't tested. The test at line 328 uses `"echo"`, which goes through the `agentSessions` branch, not the parent-state branch at `resetAgent` line 182.

src/session/manager.ts
```diff
@@ -188,8 +243,29 @@ export class SessionManager {
...
-        await this.resetSession(session.threadId);
-        this.onCommandResponse?.(event, "Session reset.");
+        const arg = event.text.trim();
+        if (!arg) {
+          this.onCommandResponse?.(
+            event,
+            "Usage: `!reset <agent>` (e.g. `!reset lead`, `!reset reproducer`) or `!reset all` to clear the entire thread session.",
+          );
+          return true;
```
pranav-growthx, line 257: 🟡 **Should-fix — admin gate leaks usage info to non-admins** (contradicts documented behavior).

The empty-arg branch posts a thread reply *before* `denyIfNotAdmin` runs. A non-admin who types `!reset` (no arg) gets the full Usage hint posted to the thread — even though `docs/features/thread-commands.md` explicitly promises:

> Non-admin invocations are silently rejected with a ❌ reaction on the trigger message — no thread reply, no log noise.

This also lets a non-admin probe the gating model (they learn `!reset all` and `!reset <agent>` exist) before being denied on the next attempt.

**Fix:** swap the two checks — admin check first, then arg check:

```ts
case "reset": {
  if (this.denyIfNotAdmin(event)) return true;
  const arg = event.text.trim();
  if (!arg) {
    this.onCommandResponse?.(event, "Usage: ...");
    return true;
  }
  // ...
}
```

src/session/manager.ts
```diff
@@ -158,6 +159,60 @@ export class SessionManager {
...
+      session.pendingMessages = [];
+      session.status = "idle";
+      session.pid = null;
+    } else if (session.agentSessions?.[agentName]) {
+      delete session.agentSessions[agentName];
+      touched = true;
+    }
+
+    await this.store.set(threadId, session);
+    return touched;
```
pranav-growthx, line 197: 🟡 **Should-fix — `onRunComplete` resurrects the agent slice you just deleted** (and re-persists a stale `sessionId`).

Trace `!reset reproducer` while the reproducer is mid-turn:

1. `resetAgent` calls `handle.kill()`, deletes `session.agentSessions.reproducer`, persists. ✅
2. Killed Claude process exits → `handle.result` resolves → `onRunComplete` fires.
3. Line 685: `const fresh = (await this.store.get(...)) ?? session;` — reads the row that was just modified.
4. Line 688: `getOrCreateAgentSession(fresh, agentName)` — **recreates the slice you just deleted** with `status: idle`.
5. Lines 696–711: writes `agentSession.sessionId = result.sessionId` (often the killed run's sessionId, captured from the `init` event before the kill) and sets `status: "failed"` (or `"done"` if no error). Persists.

Net effect: a few seconds after a successful-looking `!reset reproducer`, the slice **reappears** with a stale `sessionId` and `status: failed`. The next user message to that agent would `--resume` a half-dead conversation.

The `!reset lead` path is worse — `resetAgent` zeroes `session.sessionId` and `session.leadSessionId`, but `onRunComplete` (lines 696–702) writes `fresh.sessionId = result.sessionId; fresh.leadSessionId = result.sessionId;`, fully defeating the reset.

**Fix options:**
- Track "recently reset" agent keys in a small `Set<string>` on the manager. `resetAgent` adds the key; `onRunComplete` checks it and bails out of the post-completion writes (with cleanup of the marker) when the run was cancelled by a reset.
- Or: have `resetAgent` swap the `handle` reference for a sentinel before calling `kill()`, and have `onRunComplete` short-circuit when the handle in `this.handles` no longer matches the one it was started with.

(Adjacent context: `!reset all` has a similar pre-existing race against `resetSession` — `store.get` returns undefined, falls back to the in-memory snapshot, and `store.set` resurrects the deleted row. That predates this PR; flagging only because the new path inherits the same shape.)

src/slack/events.ts
```diff
@@ -71,6 +71,13 @@ export function registerEventHandlers(
       return;
     }
 
+    // Sibling Claude bots (Friday, Doraemon) post their streaming "thinking"
+    // updates with a leading ✽ glyph. Drop those — they're not user input.
+    if (text.trimStart().startsWith("✽")) {
```
pranav-growthx: 🟡 **Should-fix — no automated test for the ✽ filter.**

This filter is the primary defense against infinite loops between Junior, Friday, and Doraemon — exactly the kind of regression that's silent until production goes wrong. The PR adds 117 lines of tests, none of which cover the new line.

There's no `src/slack/events.test.ts` file at all. A minimal test would:
- Construct a fake Bolt `app` with `app.event` capturing the handler.
- Invoke the handler with `{ text: "✽ Thinking…", channel, ts, user }` and assert `onMessage` was NOT called.
- Invoke with `{ text: "hello from friday", ... }` and assert `onMessage` WAS called (proves the filter doesn't over-match foreign bots' real messages).
- Edge: leading whitespace + ✽ should still drop (covers `text.trimStart()`).

Given this is commit 1's whole purpose, the test is cheap insurance.

src/slack/events.ts
```diff
@@ -71,6 +71,13 @@ export function registerEventHandlers(
...
+      logDrop("foreign-bot-thinking", evMeta);
+      return;
+    }
```
pranav-growthx, line 86: 🟢 **Nice-to-have — `app_mention` doesn't apply the ✽ filter.**

In practice sibling bots' streaming "thinking" lines won't `<@Junior>` so this is theoretical. But if a tool ever emits an @-mention that happens to start with ✽ (e.g. someone wires up a sibling bot to pipe its thinking into a thread where Junior is mentioned), the filter would silently miss it. Either:
- Move the filter into a small `shouldDropText(text)` helper used by both `message` and `app_mention` handlers, or
- Add a comment explaining why `app_mention` is exempt.

Not blocking.

pranav-growthx (changes requested):
Two-pass review, first pass found blocking issues. Two 🟡 should-fix items (one race/data-integrity, one admin-gate ordering bug that contradicts the documented behavior) plus a 🟡 missing-test and a 🟢 nice-to-have. Details inline.

**Verdict: REQUEST CHANGES** — fix the bare-`!reset` gate ordering and the `onRunComplete` resurrection of reset slices, and add a test for the ✽ filter.

pranav-growthx (approved):
Re-reviewed b835846. All three previous findings addressed:

1. **Admin gate on !reset** — `denyIfNotAdmin` now runs before the empty-arg branch (manager.ts:250). Non-admin bare `!reset` gets a silent ❌ with no usage hint posted. Test at events.test.ts is replicated in manager.test.ts:426 (`expect(onCmd).not.toHaveBeenCalled()`).

2. **Reset-during-flight race** — `onRunComplete` now accepts `ownHandle` and bails if `this.handles.get(handleKey) !== ownHandle` (manager.ts:690). Verified the bail also skips `onResponse`. Both `!reset <agent>` and `!reset all` mid-flight paths are tested (manager.test.ts:350, 374). Identity check correctly handles drain path (new handle overwrites map entry, prior run's onRunComplete already done), error path (both arms of `result.then` pass handle), and timeout path (wrapped handle from `withTimeout` is what's stored and what `.then` is attached to).

3. **✽ filter + tests** — `isForeignBotThinking` extracted, applied in both `message` (events.ts:83) and `app_mention` (events.ts:148) handlers. New events.test.ts covers leading-whitespace edge cases, mid-text false negatives, and passthrough.

Note (not blocking): a pre-existing race remains where a `!reset` arriving in the async gap between `runClaudeWithAgent` invocation and `handles.set(handleKey, handle)` would miss the kill loop entirely. The new handle then registers post-reset, its onRunComplete sees its own handle in the map, and it writes back state. This pre-dates the PR and is narrow (sub-millisecond gap with no awaits before withTimeout/handles.set on the synchronous path); flagging only for v2-backlog.

PR #22, opened 2026-05-12
fix(onboard-member): mandatory membership pre-check + correct verify path

PranavBakre:
## Summary

Two bugs from the first live use of \`!onboard-member\` ([thread](https://teamgrowthx.slack.com/archives/C0338BCK1UL/p1778524598500099)):

1. **Worker re-onboarded an already-active member.** Phase 1 had a "stop if already a member" guard, but it was checking \`subscriptions.MEMBERSHIP_PRO.is_completed\` on the users doc — a path that doesn't exist on either collection. The check always returned falsy → worker proceeded as if onboarding was needed.

2. **Phase 3 verify had the same wrong path.** Last round's review caught \`user\` → \`user_id\` (Mongoose virtual), but the completion-flag path remained stale.

## Fix (in submodule \`junior-private-agents\` \`21d8440\`, pulled in here)

- **Phase 1 split into 1a + 1b.** 1a finds the user in \`users\` (unchanged). 1b is a new mandatory step that queries \`memberdatas\` by \`user_id\` for \`onboarding.subscription_plans.MEMBERSHIP_PRO.is_completed: true\`. If 1b returns a doc → stop, surface "already a member."
- **Phase 3 verify** uses the same correct shape — same collection, same path. Two field-path traps now documented inline (the \`user\` virtual + the \`subscriptions\` vs \`onboarding.subscription_plans\` confusion) so a future reader doesn't fall into either.

Path verified against \`gx-backend/packages/database/models/member_data.ts\`:
- L213-216: \`memberOnboardingSubscriptionPlanSchema\` defines \`is_completed: Boolean\`
- L337-338: \`onboarding.subscription_plans\` is a \`Map\` of that schema, keyed by plan name (\`MEMBERSHIP_PRO\`)

## Test plan

- [ ] After merge + bot restart + submodule refresh: re-test with a known already-onboarded user (e.g. Ruta Bhatt, \`_id\` 64a4f06a0b03d63335433ee8). Phase 1 should return "already a member" and stop before Phase 2.
- [ ] Re-test with a never-onboarded user. Phase 1 → Phase 2 plan → Phase 3 execute. Verify in step 5 should return exactly one doc and not falsely report failure.

## Note on routing (the *other* mistake from the live thread)

The user also reported "junior didn't route to the agent." That's almost certainly a deployment issue — the previous PR (#21) was merged but the bot needs a restart + submodule pull to actually run the new dispatch wiring + load the new agent doc. Not addressable in this PR; surfacing separately.


PranavBakre:
Pushed `7b00c8b` adding the second fix from the live thread.

**Intent recognition (the *other* mistake from p1778524598500099):** Updated `common/orchestrator-dispatch.md` to trigger on **data shape**, not the verb.

Harpreet's ask was "can you check if there is a member called Raul John Aju, here is the phone number: 8943344344 and email raulthereal1@gmail.com". Verb was "check" → default Junior treated it as a lookup, stopped at "no match." Real intent was onboarding — the name+phone+email triplet is the giveaway.

Updated rules in the preamble:
- **Strongest trigger is the shape: name + phone + email together.** Verb is incidental.
- Canonical miss listed verbatim ("can you check if there is a member called X…") so future Junior can pattern-match it.
- Partial contact info (name + one of phone/email) → do the lookup inline, but if "no match" → flip to dispatch.
- Pure curiosity with no contact info ("is X in the DB?") stays inline.

Now this PR has two fixes from the live use:
1. Mandatory membership pre-check in Phase 1 (uses correct `onboarding.subscription_plans.MEMBERSHIP_PRO.is_completed` path on `memberdatas`) — covers "agent didn't check for membership status before onboarding."
2. Broader trigger for the orchestrator-dispatch preamble — covers "Junior didn't realise this was a user onboard issue."

Both fixes are in the submodule (`junior-private-agents` at `6a14420`); junior side is just two pointer bumps.

PranavBakre (commented):
## Review: fix(onboard-member) — mandatory membership pre-check + correct verify path

(Posting as COMMENT because GitHub blocks REQUEST_CHANGES on one's own PR. Treat as "do-not-merge until 🟡 is addressed.")

### Verification of the headline claims

**Field paths — confirmed correct.** Cross-checked against `gx-backend/packages/database/models/member_data.ts`:
- L305 — `user_id: { type: ObjectId, ref: 'User', required: true, unique: true }` — query returns 0 or 1 doc
- L213-216 — `is_completed: { type: Boolean, default: false }` on `memberOnboardingSubscriptionPlanSchema`
- L337-338 — `onboarding.subscription_plans` is a `Map` of that subdoc
- L454 — `schema.index({ 'onboarding.subscription_plans.MEMBERSHIP_PRO': 1 })` — exact field is indexed
- `packages/database/utils/user.ts:9` reads `memberData?.onboarding?.subscription_plans?.MEMBERSHIP_PRO` — same canonical access pattern used in production code

Both the Phase 1 1b query and Phase 3 verify use the exact correct shape. The two field-path traps (the `user`-virtual trap and the `subscriptions`-vs-`onboarding.subscription_plans` trap) are now documented inline.

**PR scope** is purely submodule pointer bumps (`22de59b → 6a14420`). PR #21's lead/default identity split and `isOrchestratorAgent` helper are untouched. Confirmed via `gh pr diff`.

### Findings

#### 🟡 Should-fix

- **`common/orchestrator-dispatch.md` L35 (partial-info clause).** The clause is scoped to "name + ONE of phone/email." A literal read leaves a gap: ONE piece of contact info with NO name (e.g. "is `bob@example.com` one of our members?"). It doesn't satisfy the "name + one piece" condition, but an email is contact info, so it also doesn't cleanly hit the "no contact info" curiosity branch on L43. In practice the orchestrator will probably treat it as curiosity, but the rule should say so explicitly. Suggest adding one sentence after L35: *"If you have one piece of contact info but no name, treat as a pure curiosity lookup — answer inline, do not dispatch."*

#### 🟢 Nice-to-have

- **`onboard-member.md` Phase 1 1b query collapse.** The query combines `user_id` + `is_completed: true` in a single filter. This hides the intermediate state where a memberdatas doc exists with `is_completed: false` (half-finished or abandoned onboarding). Worker can't currently surface "user exists in memberdatas but onboarding wasn't completed" as a distinct branch — it looks identical to "no memberdata doc." Since `user_id` is unique on the collection, lifting 1b to `{ user_id: ObjectId(...) }` with an `is_completed` projection costs nothing and lets you branch on three outcomes: no doc → onboard; doc with `false` → flag to Captain that this is recovery / investigate first; doc with `true` → already a member. Optional — depends on whether the migration scripts are idempotent over half-onboarded users.

- **`common/orchestrator-dispatch.md` intent-shape rule.** "name + phone + email together = dispatch, regardless of verb" is a strong heuristic, but it false-positives in admin/moderation contexts: "this user is sketchy, flag them: name + phone + email", "remove X from the system: name + phone + email", "did X (phone, email) attend our event?". The cost of over-trigger is bounded (Phase 1 1b will stop on already-members, Phase 2's human gate lets Captain veto), but it still costs a worker spawn and a clarifying round-trip. Consider adding to the NOT-for list: *"Moderation / removal / analytics asks that happen to include contact info — verbs like 'flag', 'remove', 'delete', 'abuse', 'duplicate', 'attended' should prompt one clarifying question, not auto-dispatch."* Optional; the current rule is defensible if you'd rather err on the side of dispatch.

### Summary

Headline fixes are correct and well-documented. The schema-path correction is exact, with the right canonical reference and an explicit failure-mode note. Field-path traps are now inline so future readers don't repeat them. Holding (do-not-merge) on one 🟡 (the partial-info-with-no-name gap on L35 of orchestrator-dispatch.md) plus two 🟢 heuristic concerns that are worth a thought but not required.

PranavBakre:
Pushed `a8db924` addressing the review.

**🟡 Should-fix — single-piece-no-name gap:** Fixed. Ambiguity rules now explicitly cover four cases:
- name + phone + email → dispatch
- name + ONE of phone/email → inline lookup, flip to dispatch on no-match
- ONE contact piece with NO name → inline curiosity lookup (the previously-uncovered case)
- no contact info → inline curiosity lookup

Also expanded the NOT-for list with moderation/safety, analytics/attendance, and verification asks — the Phase 2 human gate would have caught these anyway but pre-filtering means fewer cycles burned and clearer worker intent.

**🟢 Nice-to-have #1 — Phase 1 1b query collapse:** Fixed. Query now filters by `user_id` only and branches on three outcomes:
- No doc → proceed with `onboard_users.ts`
- Doc with `is_completed: true` → already a member, stop
- Doc with flag falsy/missing → half-onboard state (account stub exists, payment never landed or script crashed mid-run). Stop and ask Captain — re-running may or may not be safe depending on side-effects.

The half-onboard surface is the real value here: previously the worker would have silently treated it the same as "no doc" and re-run the script, potentially corrupting state. Now it gates on Captain explicitly.

**🟢 Nice-to-have #2 — moderation/analytics false positives:** Covered by the expanded NOT-for list in the orchestrator-dispatch preamble.

Both refinements in submodule (`junior-private-agents` `beabdc8`); junior side is just the pointer bump.

Ready for re-review.

PranavBakre (commented):
## Re-review: APPROVED (two consecutive clean passes)

(Posting as COMMENT — GitHub blocks self-approval. Treat as **ship-it** from a review standpoint; merge is your call.)

Re-reviewed `a8db924` (submodule `beabdc8`). All three findings from the prior review are addressed cleanly.

### Verification of the three fixes

**1. Single-piece-no-name (orchestrator-dispatch L35-40)** — The ambiguity ladder now enumerates a complete partition over (has-name × contact-pieces):
- name + phone + email → dispatch (no exceptions)
- name + ONE → ambiguous, lookup-first
- one piece, no name → curiosity inline
- no info → curiosity inline

The L42 "don't stop at no-match when the ask had a full contact shape" caveat scopes cleanly to the first case only. No remaining gap.

**2. Phase 1 1b half-onboard branch (onboard-member L52-63)** — Filter is now `user_id` only with the completion flag projected. Three outcomes:
- no doc → proceed (`onboard_users.ts`)
- `is_completed === true` → stop, already a member
- flag falsy / missing → "**Stop** and ask Captain... DO NOT silently proceed to re-run the script... Wait for instruction"

The downstream combined-result block at L67-73 mirrors all three 1b outcomes (plus the two 1a-only outcomes for zero/multiple matches). No contradiction between the inline 1b prose and the combined block.

**3. NOT-for list expansion (L48-50)** — Three new entries (moderation/safety, analytics/attendance, verification) with realistic examples. Combined with the pre-existing entries (bulk / already-a-member / account-without-onboard), the list covers the realistic surface where name+phone+email shape could appear without onboarding intent.

### Pass 6 surface check

Read both files end-to-end. No stale references, no contradictions between new and old sections, no typos. The "shape over verb" framing at L25 stays consistent with the new "shape ladder" at L35-40 (shape > verb when shape is *present*; absence of shape falls back to the L17 "ask one clarifying question" general rule).

### Scope

Pure submodule pointer bump (1 file, +1/-1). No source code changed. Test posture unchanged (303/304, pre-existing dev-server flake).

PR #24, opened 2026-05-12
refactor(agents): move private agent identities to frontmatter

PranavBakre:
## Summary

Removes the last leak of an org-specific agent name from the public junior repo.

**Before:** `AGENT_IDENTITIES` in `src/support/agents.ts` had `"onboard-member": { username: "Onboarder", iconEmoji: ":handshake:" }` hardcoded. Every fork of junior inherited this entry even though the worker's actual playbook lives in the private agents-org overlay — leaking both the existence of the org-specific workflow and its slack identity.

**After:** private/overlay agents declare their slack identity in their own `.md` frontmatter. Junior loads the overlay directory at startup, parses the new `username` and `iconEmoji` fields, and registers the identities into `AGENT_IDENTITIES` dynamically. The public AGENT_IDENTITIES literal contains only the six core architectural agents.

## What changed

| File | Change |
|---|---|
| `src/agents/loader.ts` | Parses optional `username` + `iconEmoji` from frontmatter; adds to `AgentDefinition` interface |
| `src/support/agents.ts` | Removes `onboard-member` from `AGENT_IDENTITIES` literal. Adds `registerAgentIdentity(name, identity)` (refuses to overwrite core agents) and `loadOverlayIdentities(dir)` (scans .md files, registers identities for any agent declaring both fields). |
| `src/index.ts` | Calls `loadOverlayIdentities(".claude/agents-org")` before `app.start()`. Non-fatal on failure (overlay is optional). |
| `docs/features/persistent-agents.md` | Example AGENT_IDENTITIES block matches the new shape; comment explains the overlay mechanism. |
| `.claude/agents-org/onboard-member.md` | Declares `username: Onboarder` + `iconEmoji: ":handshake:"` in frontmatter (in the submodule). |

## Backward compat

- Agents without `username`/`iconEmoji` frontmatter keep working — those fields are optional on `AgentDefinition`.
- Existing core agents are untouched; their identities stay in code.
- `registerAgentIdentity` refuses to overwrite core agents — overlay can't silently re-skin lead, thinker, etc.

## Architectural intent (the user's original question)

The junior/agents-org split now holds **both** structurally and content-wise:

- **Junior public:** core agent names + generic dispatch architecture. No `gx-backend`, `MEMBERSHIP_PRO`, `growthx`, or any org-specific name reaches the public repo.
- **Agents-org private:** full playbooks AND identities for org-specific workers. Self-contained — adding a new private agent is one .md file with the right frontmatter, no code change required in junior.

## Test plan

- [x] `bun test` — 311/312 (one pre-existing dev-server integration flake, unrelated)
- [x] New tests:
  - `loader.test.ts`: parses `username`/`iconEmoji` from frontmatter; both default to `null` when absent.
  - `agents.test.ts`: `registerAgentIdentity` registers + refuses to overwrite core; `loadOverlayIdentities` registers from a temp dir, skips half-declared identities, doesn't throw on missing dir; overlay-registered agent appears in `dispatchableAgentsFor("lead")` after registration.

## Supersedes #23

PR #23 was a small PII sanitization (replacing real names in test fixtures with "Test Member"). This refactor changes those same fixtures more thoroughly — they now use core agent names (\`!review\`, \`!thinker\`) because the refactor's tests don't depend on the overlay being loaded. Closing #23 as superseded.

### DIFF
--- a/src/support/agents.ts
+++ b/src/support/agents.ts
@@ -10,9 +10,9 @@ import type { AgentIdentity } from "../session/types.ts";
  * `isOrchestratorAgent` set, attribution-suffix logic, etc.) — they're part
  * of the contract any fork of junior inherits.
  *
- * Private / org-specific workers (e.g. `onboard-member`) register their
- * identities via `username` + `iconEmoji` frontmatter on their `.md` files,
- * loaded at startup by `loadOverlayIdentities` from the org overlay
+ * Private / org-specific workers (e.g. an org-specific worker) register
+ * their identities via `username` + `iconEmoji` frontmatter on their `.md`
+ * files, loaded at startup by `loadOverlayIdentities` from the org overlay
  * directory. This keeps the public repo free of org-specific names.
  */
 export const AGENT_IDENTITIES: Record<string, AgentIdentity> = {
@@ -70,11 +70,32 @@ export async function loadOverlayIdentities(dirPath: string): Promise<void> {
     }
     for (const entry of entries) {
       const def = await loadAgentDefinition(`${dirPath}/${entry}`);
-      if (!def || !def.name) continue;
-      if (!def.username || !def.iconEmoji) continue;
+      if (!def) continue;
+      const hasUsername = !!def.username;
+      const hasIcon = !!def.iconEmoji;
+      // Genuine "no slack identity declared" — fine. Many overlay files are
+      // pure prompt with no slack-posting role (Task-tool sub-agents, etc.).
+      if (!hasUsername && !hasIcon) continue;
+      // Identity declared but incomplete — likely a typo. Warn so the author
+      // can fix it; without this signal the agent silently has no identity
+      // and the next observable failure is at dispatch time.
+      if (!hasUsername || !hasIcon) {
+        log.warn(
+          "agents",
+          `loadOverlayIdentities: ${entry} declares ${hasUsername ? "username" : "iconEmoji"} but not ${hasUsername ? "iconEmoji" : "username"} — both fields are required; skipping`,
+        );
+        continue;
+      }
+      if (!def.name) {
+        log.warn(
+          "agents",
+          `loadOverlayIdentities: ${entry} declares username + iconEmoji but no name — skipping; add a 'name:' field in frontmatter`,
+        );
+        continue;
+      }
       registerAgentIdentity(def.name, {
-        username: def.username,
-        iconEmoji: def.iconEmoji,
+        username: def.username!,
+        iconEmoji: def.iconEmoji!,
       });
     }
   } catch (err) {
--- a/src/support/agents.ts
+++ b/src/support/agents.ts
@@ -1,5 +1,20 @@
+import { loadAgentDefinition } from "../agents/loader.ts";
+import { log } from "../logger.ts";
 import type { AgentIdentity } from "../session/types.ts";
 
+/**
+ * Core agents — part of the open-source architecture. Orchestrators (lead,
+ * default) and bug-pipeline workers (reproducer, thinker, review) plus the
+ * echo debug agent. Identities live in code because these names are
+ * referenced from junior's source (special-cases in `manager.ts`, the
+ * `isOrchestratorAgent` set, attribution-suffix logic, etc.) — they're part
+ * of the contract any fork of junior inherits.
+ *
+ * Private / org-specific workers (e.g. `onboard-member`) register their
+ * identities via `username` + `iconEmoji` frontmatter on their `.md` files,
+ * loaded at startup by `loadOverlayIdentities` from the org overlay
+ * directory. This keeps the public repo free of org-specific names.
+ */
 export const AGENT_IDENTITIES: Record<string, AgentIdentity> = {
   // Default Junior — the bot's main face, responds to @mentions in any channel.
   default: { username: "Junior", iconEmoji: ":face_with_cowboy_hat:" },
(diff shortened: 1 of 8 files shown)


src/support/agents.ts
```diff
@@ -1,5 +1,20 @@
...
 
+/**
+ * Core agents — part of the open-source architecture. Orchestrators (lead,
+ * default) and bug-pipeline workers (reproducer, thinker, review) plus the
+ * echo debug agent. Identities live in code because these names are
+ * referenced from junior's source (special-cases in `manager.ts`, the
+ * `isOrchestratorAgent` set, attribution-suffix logic, etc.) — they're part
+ * of the contract any fork of junior inherits.
+ *
+ * Private / org-specific workers (e.g. `onboard-member`) register their
```
pranav-growthx: **🟡 Should-fix — org-specific name in this docstring.**

```
 * Private / org-specific workers (e.g. `onboard-member`) register their
```

This still reveals the existence of an org-specific workflow named `onboard-member` to anyone reading public junior's source. Less egregious than the test (this is explaining a mechanism, not asserting on the name), but it's still grep-visible. The same grep that found the original leak finds this.

**Fix:** replace `` (e.g. `onboard-member`) `` with a neutral placeholder, e.g. `(e.g. an org-specific worker)`. The mechanism description doesn't need to name a real agent.

src/support/agents.ts
```diff
@@ -11,9 +26,66 @@ export const AGENT_IDENTITIES: Record<string, AgentIdentity> = {
...
+export async function loadOverlayIdentities(dirPath: string): Promise<void> {
+  try {
+    const glob = new Bun.Glob("*.md");
+    const entries: string[] = [];
+    for await (const entry of glob.scan({ cwd: dirPath })) {
+      entries.push(entry);
+    }
+    for (const entry of entries) {
+      const def = await loadAgentDefinition(`${dirPath}/${entry}`);
+      if (!def || !def.name) continue;
```
pranav-growthx: **🟢 Nice-to-have — silent skip when overlay file lacks `name:`.**

```ts
if (!def || !def.name) continue;
if (!def.username || !def.iconEmoji) continue;
```

If an overlay author writes a `.md` file that declares `username` + `iconEmoji` but forgets `name:` (or typos it), the file is silently skipped — both inside `loadAgentDefinition` (which returns `name: ""`) and here. They won't see a warning at boot; they'll only discover it when a Slack post comes through with the wrong identity (or no identity gets registered at all).

Consider differentiating the cases:
```ts
if (!def) continue;
if (!def.username || !def.iconEmoji) continue; // no identity declared — skip silently
if (!def.name) {
  log.warn("agents", `loadOverlayIdentities: ${entry} declared username/iconEmoji but no name; skipped`);
  continue;
}
```

Order matters: only warn when the file *clearly intended* to register an identity (both fields present) but is missing the key. Don't warn on agents that just lack identity fields entirely — those are legitimately silent.

pranav-growthx (changes requested):
## Review: PR #24 — refactor(agents): move private agent identities to frontmatter

The architecture is right: parser, registry, and startup ordering all check out. `loadOverlayIdentities` runs before `app.start()` and before any event handler is registered, so no session can spawn against a half-loaded `AGENT_IDENTITIES`. All non-test readers (`identityForAgent`, `agentForUsername`, `isPersistentAgent`, `dispatchableAgentsFor`) read at call time. The frontmatter parser handles quoted and unquoted colon-containing values correctly. PR #21 (lead/default split) and PR #22 (Phase 1 pre-check) survive untouched.

But by the PR's own success criterion — "No `gx-backend`, `MEMBERSHIP_PRO`, `growthx`, or any org-specific name reaches the public repo" — the leak isn't fully closed. Two stale references to `onboard-member` remain in the public tree. The whole point of the refactor was to remove that name from public; leaving it in tests and code comments after the move undercuts the goal.

---

### 🔴 Blocker — org-specific name still in public test

**`src/support/agents.test.ts:86`** (line not touched by this PR — it's a stale survivor from PR #21/22):
```ts
expect(isOrchestratorAgent("onboard-member")).toBe(false);
```

The PR removed 3 of the 4 `onboard-member` references from this test file but missed this one. The literal string serves no test-value tied to the specific name — any non-orchestrator string (`"test-worker"`, `"private-worker"`) would make the same assertion. A `git grep onboard-member src/` from a fresh clone of public junior still lights up here.

The entire architectural point of the refactor — moving the identity into the private overlay — is undone slightly by leaving the name as a literal in a public test.

**Fix:** replace `"onboard-member"` with `"test-worker"` (or any neutral name) on that line.

---

### 🟡 Should-fix — org-specific name in new docstring (introduced by this PR)

**`src/support/agents.ts:13`** — see inline comment.

---

### 🟢 Nice-to-have — silent skip when overlay file lacks `name:`

**`src/support/agents.ts:73`** — see inline comment.

---

### Test posture confirmed
- `bun test src/support/agents.test.ts src/agents/loader.test.ts src/slack/events.test.ts src/support/router.test.ts` — 71/71 pass.

### What was verified
- Startup ordering: `loadOverlayIdentities` awaited before `app.start()` and before `registerEventHandlers`. Module-level constants (`AgentRouter`, `SessionManager`, `AgentDispatcher`) constructed before the IIFE do not eagerly read `AGENT_IDENTITIES`; all consumers read at call time.
- `Bun.Glob("*.md")` is non-recursive — overlay `common/` subdirectory is correctly skipped by `loadOverlayIdentities`.
- Frontmatter parser handles quoted/unquoted colon-containing values (`iconEmoji: ":handshake:"` and `iconEmoji: :handshake:` both parse correctly because `indexOf(":")` splits on the FIRST colon).
- `registerAgentIdentity` refuse-on-overwrite log message names the offending key and is descriptive.
- Tests' `afterEach` cleanup is scoped per-describe-block and shouldn't leak across blocks.

pranav-growthx:
Pushed \`ac043cc\` addressing the review.

**🔴 Blocker — stale \`onboard-member\` reference in \`agents.test.ts:86\`:** Fixed. The \`isOrchestratorAgent\` test now uses \`review\` (also a worker, also exercises the rejects-workers branch). The PR's own success criterion now holds:

\`\`\`
$ grep -rn "onboard-member|Onboarder|handshake" src/ docs/
(no output)
\`\`\`

That grep is exactly the leak check I should have run *before* claiming the refactor was clean. Painful lesson — re-learning the principle I just added to learnings.md this morning ("when claiming clean separation, run the negation audit, not the default audit").

**🟡 Should-fix — docstring example:** Fixed. The new docstring at \`src/support/agents.ts:13\` now says \`(e.g. an org-specific worker)\` instead of naming \`onboard-member\` specifically.

**🟢 Nice-to-have — silent skip on no-name:** Fixed and slightly extended. The loader now warns in two distinct cases:
- Identity fields declared but one of the two is missing → \`"declares username but not iconEmoji — both fields are required; skipping"\` (likely typo)
- \`username + iconEmoji\` declared but no \`name:\` → \`"declares username + iconEmoji but no name — skipping; add a 'name:' field in frontmatter"\`

The genuinely-no-identity case (a pure-prompt overlay file with neither field) stays silent, since that's a valid shape for Task-tool sub-agents that never post to slack.

Added a test for the no-name case (\`skips .md files that have identity fields but no name\`); existing half-identity test still covers the partial-declaration branch.

71/71 tests pass on the affected files. Ready for re-review.

pranav-growthx (approved):
Re-review on `ac043cc`. Two consecutive clean passes — approved.

### Verification (all three prior findings addressed)

**🔴 Blocker resolved** — `src/support/agents.test.ts:86` now reads `expect(isOrchestratorAgent("review")).toBe(false)`. The leak grep `git grep -nE "onboard-member|Onboarder|handshake" src/ docs/` returns zero matches. `review` is a real worker, still exercises the same `false` branch.

**🟡 Should-fix resolved** — `src/support/agents.ts:13` now reads `(e.g. an org-specific worker)`. The mechanism description is intact; no org-specific name surfaces.

**🟢 Nice-to-have resolved** — `loadOverlayIdentities` now has three distinct paths in `src/support/agents.ts:73-95`:
- No identity fields at all → silent continue (legitimate Task-tool sub-agents)
- Partial identity (one of two fields) → warn naming the missing field
- Full identity but no `name:` → warn telling the author to add a `name:` field
- The new test `skips .md files that have identity fields but no name` (lines 241-258) writes a file with `username` + `iconEmoji` but no `name:`, exercising the `!def.name` branch on line 89 — relies on `loader.ts:62` (`name: frontmatter["name"] ?? ""`) returning the empty-string fallback that falsy-tests to true.

### Pass 6 (surface) re-sweep
- Repo-wide grep for `onboard-member|Onboarder|handshake` in non-overlay paths: zero hits.
- Three remaining matches for "onboard" are generic English ("onboarding flows", "onboarding experience") in `.claude/agents/lead.md` and `.claude/agents/pm.md` — not leaks of the agent name.
- Warn message templates are clear and actionable (name the missing field by name).
- `events.test.ts:153` fixture (`!review take a look at PR 21`) still resolves correctly — `review` is in `AGENT_IDENTITIES` as a core agent.

### Test posture
- `bun run typecheck` clean (on the real repo with `node_modules` present; the worktree's typecheck failure was missing deps, not a real error).
- `bun test src/support/agents.test.ts src/agents/loader.test.ts src/slack/events.test.ts src/support/router.test.ts` → 72/72 pass.
- PR #21 (lead/default split) and PR #22 (Phase 1 pre-check) behavior intact — `dispatchableAgentsFor`, `agentForUsername`, `isOrchestratorAgent` and `workerMayDispatch` tests all pass.

### What was newly verified on this pass
- Three-branch control flow in `loadOverlayIdentities` is sound for every input shape: null def, no-fields, partial-fields, full-fields-no-name, full.
- Non-null assertions `def.username!` / `def.iconEmoji!` on lines 97-98 are safe — guarded by the short-circuit on lines 78 + 82.
- afterEach cleanup in the new test correctly omits `cleanupKeys.push(...)` since nothing was registered.
- Empty-string username/iconEmoji from quoted-empty frontmatter degrades to "not declared" (silent skip) — reasonable handling.

Approved.

PR #25, opened 2026-05-13
feat(admins): support additional admins via sqlite alongside env bootstrap

PranavBakre:
## Summary
- Adds an `admins` table to the sqlite session DB so additional admins can be granted via direct SQL, while `ADMIN_SLACK_USER_ID` remains the bootstrap admin (keeps the system bootable on a fresh DB).
- New `SessionStore.extraAdmins()` method — memory store returns empty, sqlite store reads the `admins` table on each `isAdmin` check (no cache, so SQL inserts take effect on the next command without a restart).
- `SessionManager.isAdmin` is now async and checks env first, then DB. Permissions stay flat — anyone listed in either tier gets mute/unmute/reset/cancel.

## Admin lifecycle
New admins are added by direct SQL — no Slack command, no CLI script. Example:

```sql
INSERT INTO admins (slack_user_id, added_at) VALUES ('U0XXXX', strftime('%s','now') * 1000);
```

## Test plan
- [x] `bun test src/session/manager.test.ts` — 35/35 pass
- [x] Full `bun test` — only pre-existing DevServer integration failure, unrelated
- [ ] Confirm on deploy that the env-var admin still passes the gate
- [ ] Confirm an admin added via direct SQL can run `!mute` without a restart

### DIFF
--- a/src/session/manager.ts
+++ b/src/session/manager.ts
@@ -198,18 +198,21 @@ export class SessionManager {
     return touched;
   }
 
-  isAdmin(userId: string): boolean {
-    const adminId = this.config.adminSlackUserId;
-    if (!adminId) return true;
-    return userId === adminId;
+  async isAdmin(userId: string): Promise<boolean> {
+    const envAdmin = this.config.adminSlackUserId;
+    // No env admin configured = open mode (local dev). Skip DB lookup.
+    if (!envAdmin) return true;
+    if (userId === envAdmin) return true;
+    const extras = await this.store.extraAdmins();
+    return extras.has(userId);
   }
 
   /**
    * Returns true if the user is denied (and a ❌ reaction was queued).
    * Caller should `return true` from the command handler to short-circuit.
    */
-  private denyIfNotAdmin(event: SlackMessageEvent): boolean {
-    if (this.isAdmin(event.user)) return false;
+  private async denyIfNotAdmin(event: SlackMessageEvent): Promise<boolean> {
+    if (await this.isAdmin(event.user)) return false;
     this.onReaction?.(event, "x");
     return true;
   }
@@ -248,7 +251,7 @@ export class SessionManager {
         // bare `!reset` gets the usage hint posted to the thread, contradicting
         // the documented "silent ❌ reaction, no thread reply" promise — and
         // leaks the gating model to anyone probing.
-        if (this.denyIfNotAdmin(event)) return true;
+        if (await this.denyIfNotAdmin(event)) return true;
         const arg = event.text.trim();
         if (!arg) {
           this.onCommandResponse?.(
@@ -435,7 +438,7 @@ export class SessionManager {
       }
 
       case "mute": {
-        if (this.denyIfNotAdmin(event)) return true;
+        if (await this.denyIfNotAdmin(event)) return true;
         session.muted = true;
         await this.store.set(session.threadId, session);
         this.onCommandResponse?.(event, "Muted. I'll stop responding until you `!unmute`.");
@@ -443,7 +446,7 @@ export class SessionManager {
       }
 
       case "unmute": {
-        if (this.denyIfNotAdmin(event)) return true;
+        if (await this.denyIfNotAdmin(event)) return true;
         session.muted = false;
         await this.store.set(session.threadId, session);
         this.onCommandResponse?.(event, "Unmuted.");
--- a/src/session/manager.ts
+++ b/src/session/manager.ts
@@ -200,11 +200,13 @@ export class SessionManager {
 
   async isAdmin(userId: string): Promise<boolean> {
     const envAdmin = this.config.adminSlackUserId;
-    // No env admin configured = open mode (local dev). Skip DB lookup.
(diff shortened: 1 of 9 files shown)


src/session/manager.ts
```diff
@@ -198,18 +198,21 @@ export class SessionManager {
...
   }
 
-  isAdmin(userId: string): boolean {
-    const adminId = this.config.adminSlackUserId;
-    if (!adminId) return true;
-    return userId === adminId;
+  async isAdmin(userId: string): Promise<boolean> {
+    const envAdmin = this.config.adminSlackUserId;
+    // No env admin configured = open mode (local dev). Skip DB lookup.
+    if (!envAdmin) return true;
```
PranavBakre: 🟡 **Should-fix — Authorization / Fail-open risk:** `if (!envAdmin) return true;` short-circuits *before* consulting the `admins` table. If `ADMIN_SLACK_USER_ID` is ever unset in prod (env reload bug, secret-manager misfire, accidentally cleared), every Slack user becomes admin **and** the populated `admins` table is silently ignored. The design contract you stated in the PR description ("one in env, rest in DB") is violated with no signal — `!reset all` / `!mute` are suddenly open to anyone.

The local-dev convenience of open-mode is fine when *nothing* is configured, but it shouldn't override an explicitly populated DB.

Fix options (pick one):

1. **Treat a populated `admins` table as opt-in to gated mode.** Always consult the DB; only return `true` (open) when both env is unset *and* the table is empty.
   ```ts
   const extras = await this.store.extraAdmins();
   if (!envAdmin && extras.size === 0) return true; // open dev mode
   if (userId === envAdmin) return true;
   return extras.has(userId);
   ```
2. **Require the env admin to be set if gating is desired.** When env is unset, treat that as a misconfiguration and `return false` for everyone — fail closed.

Option 1 preserves your stated UX (DB inserts take effect without restart) and closes the fail-open hole.

PranavBakre (commented):
Review for PR #25 — `feat/admin-storage-sqlite`. One should-fix around a fail-open hole when `ADMIN_SLACK_USER_ID` is unset with a populated `admins` table, one self-contradicting comment, and one missing test for the new code path. See inline.

src/session/store/sqlite.ts
```diff
@@ -37,6 +37,15 @@ export class SqliteSessionStore implements SessionStore {
         FOREIGN KEY (thread_id) REFERENCES sessions(thread_id)
       )
     `);
+    // Extra admins beyond the env-var bootstrap. Added by direct SQL —
+    // process restart picks up changes; isAdmin() reads from this table each
+    // call so no cache invalidation is needed.
```
PranavBakre: 🟢 **Nit — self-contradicting comment:** the first sentence says *"process restart picks up changes"* but the very next sentence says *"isAdmin() reads from this table each call so no cache invalidation is needed"* — the second is correct, the first is the opposite of what the code does. Suggest:

```ts
// Extra admins beyond the env-var bootstrap. Added by direct SQL —
// isAdmin() reads from this table on every check so inserts take effect
// on the next command without a restart.
```

src/session/manager.ts
```diff
@@ -198,18 +198,21 @@ export class SessionManager {
...
+    if (userId === envAdmin) return true;
+    const extras = await this.store.extraAdmins();
+    return extras.has(userId);
```
PranavBakre: 🟡 **Should-fix — missing test for the new code path:** `manager.test.ts` covers the env-admin allow, non-admin deny, and open-mode cases — but nothing asserts that a user listed in `extraAdmins()` is admitted. That's the behavior this PR is introducing; without a test, a future refactor (e.g. accidentally changing precedence, or breaking the `Set.has()` check) silently regresses the feature.

Minimal addition: stub `store.extraAdmins` on the existing `InMemorySessionStore` (or attach a one-shot override on the instance) and assert that `U-EXTRA` can run `!mute` while `U-OTHER` still gets ❌.

PranavBakre (commented):
Second pass clean — approving (posting as comment since self-approval is blocked by GitHub).

Verified on HEAD `076db45`:

- **Fail-open hole closed.** Walked every branch of `isAdmin` truth table. The only `return true` paths are env-match, DB-match, or BOTH tiers empty. The earlier attack (env unset + DB populated → unlisted user is admin) now correctly returns false.
- **Async migration correct.** `isAdmin` is now async; only caller (`denyIfNotAdmin` at manager.ts:217) properly awaits.
- **Error path is fail-closed.** If `store.extraAdmins()` throws, the rejection propagates through `denyIfNotAdmin` → command handler; no silent promotion.
- **Comment nit fixed.** sqlite.ts:40-42 no longer contradicts itself.
- **Tests cover the new code path.** Both the DB-admin-admitted case and the env-unset-DB-populated-closed regression guard are asserted. Monkey-patching `store.extraAdmins` on the instance is valid (TS allows property reassign of an async method; runtime shadows the prototype).
- Typecheck clean. `bun test` → 314/315 pass (one pre-existing unrelated DevServer integration failure).

PR #28, opened 2026-05-14
Thread attention gate: !aside, !listen, auto-dormant on human sidebar

PranavBakre:
## Summary

Junior used to keep replying to every message in a thread that had ever addressed it, even after the conversation turned into a human-to-human sidebar. `!mute` exists but is admin-only and silences the bot everywhere. This adds a thread-local control surface:

- **`!aside [text]`** — drops just that message (👀 reaction, no state change). For an aside inside an active thread.
- **`!listen`** — clears auto-dormant (👂 reaction). Anyone in the thread.
- **Auto-dormant** — when a second human posts without `@`-mentioning Junior, Junior posts one notice and goes silent. `@mention` or `!listen` wakes it.

## The trigger rule

The signal that matters is *"addressed to me,"* not number of participants — counting misfires both ways. Decision rule: a non-Junior human reply that doesn't `@mention` Junior, in a thread where another human has already participated.

Bots — Junior, its agents (shared `bot_id`), foreign bots like Friday/Doraemon/CI webhooks — are never counted as conversation partners.

Auto-trigger channels like `#bugs-backlog` are **exempt**: humans-talking-to-humans is the intended input there, not a sidebar.

## One-shot announcement

Sticky `dormantAnnounced` flag prevents re-trigger thrashing. Once the human takes manual action (`!listen`), trust them; if they want to re-silence, that's `!aside` or `!mute`, not another auto-trigger.

## Persistence

`dormant`, `dormantAnnounced`, `humanParticipants` live on `ThreadSession` and persist via SQLite (json blob, normalized on read; pre-existing sessions default cleanly).

## Where the gate lives

`SessionManager.gateAttention(event)` runs in `index.ts` **before** `supportRouter.handleMessage`. Returns `true` when the message is consumed — no agent dispatch, no Claude spawn.

## Test plan

- [ ] Thread with one human + Junior — chats freely, never goes dormant.
- [ ] Second human joins and `@`-mentions Junior — wakes (no-op, wasn't dormant). Subsequent posts from either human route normally.
- [ ] Second human posts without `@` while first human is present — dormant, notice posted **once**. Subsequent messages dropped silently.
- [ ] `!listen` from anyone — wakes; next message routes normally; auto-dormant does **not** re-trigger.
- [ ] `!aside fyi I have to step out` — dropped with 👀, thread state unchanged.
- [ ] `#bugs-backlog`-style channel — second reporter posts a related bug, Junior still picks it up (gate exempted via `channelDefaults`).
- [ ] CI bot posts into a 1-human thread — does not register as a 2nd participant; trigger doesn't fire.

12 unit tests added covering each branch of the gate; full `bun test` is green (one pre-existing failing dev-server integration test on `main`, unrelated).


PranavBakre (commented):
## Review: PR #28 — feat/thread-aside-dormant

(Posting as **REQUEST_CHANGES**. Single commit, so per-commit and cumulative diff are the same — verified.)

### What I verified

- **Per-commit posture.** Single commit `a80e006`. Per-commit = cumulative. Stated explicitly per the repo's review-per-commit rule.
- **Test posture.** `bun test` runs **278 pass / 1 fail**. The single failure is `DevServerManager > ensure() restarts the server when branch changes` — confirmed pre-existing flake on `main`, called out in the PR description. New gate logic adds 10 unit tests + 3 type-shape tests (13 total, vs the "12" claim — close enough). All gate tests pass.
- **Persistence migration.** `normalizeSession` in `sqlite.ts` is applied on every read path (`get`, `getAll`, `getRecent`) and on `set`. `updateActivity` doesn't read the new fields, so the missing normalize call there is fine. Pre-existing rows default cleanly: `dormant=false`, `dormantAnnounced=false`, `humanParticipants=[]`.
- **Wake idempotence.** Slack delivers both `message` and `app_mention` for `<@junior>` in a thread. `gateAttention` runs *before* `seenMessages` dedup, so it fires twice. Walked through — both fires are idempotent: the second wake-on-mention is a no-op (`dormant` already false), no second notice posts. Good.
- **Notice goes to the thread, not the channel root.** `onCommandResponse` is wired to `responder.postResponse(event.channel, event.threadId, response)` (index.ts:114-117). `responder.postResponse` passes `thread_ts: threadTs` (responder.ts:40). The "side conversation" notice lands in the thread, not the channel.
- **Sticky `dormantAnnounced`.** Confirmed never reset in code. `!listen` clears `dormant` only. Tested in `does not re-trigger after manual !listen`.
- **Auto-trigger channel exemption.** `channelDefaults[event.channel]` early-returns false before the trigger block. Tested in `exempts auto-trigger channels`.
- **Race window between participant write and gate read.** `getOrCreateSession` adds the user and writes synchronously. The window where two events from different users could both miss each other in `humanParticipants` is bounded by the gap between the gate's `store.get` and `getOrCreateSession`'s `store.set` for the *previous* event — i.e. genuinely concurrent first-time posts. Pre-existing concurrency posture (`supportRouter.handleMessage` is fire-and-forget at index.ts:210, unchanged in this PR), not a PR-introduced regression. Not flagging.

### Findings

#### 🟡 Should-fix

- **app_mention handler doesn't populate `botId` / `isSelfBot`** — see inline comment on `src/slack/events.ts`. The gate now reads these flags to classify humans vs bots; only the `message` handler sets them. A foreign bot's `<@junior>` mention enters via `app_mention` with no `botId`, gets classified as human, and pollutes `humanParticipants`. A later real human's non-@-mention reply would then falsely trigger the sidebar gate (the foreign bot's user_id is in `humanParticipants` as an "other human").

#### 🟢 Nice-to-have

- **`humanParticipants` unbounded** — see inline on `src/session/types.ts`. Threads that accumulate many distinct human posters grow this array forever. No cap, no rotation. Bounded by Slack channel membership in practice (small), but worth a one-line comment.
- **`!aside` skips participant tracking** — see inline on `src/session/manager.ts`. The `!aside` early-return at gateAttention:87-90 returns before `getOrCreateSession` ever runs, so the aside-sender isn't added to `humanParticipants`. Real-world impact is trivial (people don't use threads solely for asides) but worth a thought.

### Summary

The design is well-scoped: "addressed to me, not headcount" is the right framing and the sticky `dormantAnnounced` flag correctly prevents trigger thrashing. State persistence and migration are clean. Test coverage walks every branch of the gate.

The one real should-fix is the asymmetric flag population between `message` and `app_mention` event handlers — this is a half-applied change that puts the gate's bot-classification at risk for any @mention from a foreign bot. Easy fix (extract `bot_id` symmetrically in the `app_mention` handler, mirroring the `message` handler at events.ts:161-164).

Not approving on this pass. Re-review once the `app_mention` handler is symmetric.

src/slack/events.ts
```diff
@@ -196,10 +213,16 @@ export function registerEventHandlers(
       ts: event.ts,
       command: parsed.command,
       files: files.length > 0 ? files : undefined,
+      mentionsJunior,
     });
```
PranavBakre, line 230: 🟡 **Should-fix — asymmetric flag population between event handlers.**

The `message` handler at lines 161-176 extracts `botId` and sets `isSelfBot` before calling `onMessage`. The `app_mention` handler here (lines 208-217) emits the event with **neither** field populated.

The gate now relies on these flags to decide who counts as a human (`manager.ts:126`: `const isHuman = !event.isSelfBot && !event.botId`). When a foreign bot posts `<@junior> something`, Slack fires `app_mention` for that mention. The bot's `user` field is present (so the event isn't dropped at line 185), but `botId` is undefined here → `isHuman` evaluates to `true` in the gate.

Concrete failure:
1. Foreign bot (CI webhook, Friday, Doraemon, etc.) @-mentions Junior in a thread.
2. `app_mention` event reaches `gateAttention` and ultimately `getOrCreateSession` with `botId=undefined`.
3. The foreign bot's `user` is added to `session.humanParticipants` as if it were a human.
4. A real human later posts a non-@-mention reply.
5. The gate's `otherHumans` filter sees the foreign bot's user_id as an "other human" → fires the sidebar trigger → posts "side conversation" notice → goes dormant.

The PR description explicitly states: *"Bots — Junior, its agents (shared bot_id), foreign bots like Friday/Doraemon/CI webhooks — are never counted as conversation partners."* That contract is broken via the `app_mention` path.

**Fix:** mirror the `message` handler's botId extraction symmetrically here. Suggest:

```ts
const botId =
  "bot_id" in event && typeof (event as { bot_id?: unknown }).bot_id === "string"
    ? (event as { bot_id: string }).bot_id
    : undefined;
const isSelfBot = !!(botId && selfBotId && botId === selfBotId);

onMessage({
  threadId,
  channel: event.channel,
  user: event.user,
  text: parsed.text,
  ts: event.ts,
  command: parsed.command,
  files: files.length > 0 ? files : undefined,
  isSelfBot,
  botId,
  mentionsJunior,
});
```

`selfBotId` is already in scope (parameter to `registerEventHandlers`).

Test to add: a foreign-bot `app_mention` arrives and is **not** added to `humanParticipants`. The existing `does not count Junior or foreign bots as humans for the trigger` test only exercises the `message` path.

src/session/types.ts
```diff
@@ -44,6 +44,25 @@ export interface ThreadSession {
...
+   * Sticky: set once when the attention gate fires its one-shot announcement
+   * for this thread, and never reset. Suppresses re-trigger thrashing after
+   * the human takes manual action (`!listen` / @mention). If they want to
+   * re-silence, that's `!aside` or `!mute`, not another auto-trigger.
+   */
+  dormantAnnounced: boolean;
+  /**
+   * Slack user IDs of humans who have posted in this thread (order of first
+   * post). Bots — Junior, its agents, foreign bots — are never recorded. Used
+   * by the attention gate to detect human-to-human sidebar.
```
PranavBakre, line 63: 🟢 **Nice-to-have — unbounded growth.**

`humanParticipants` accumulates Slack user IDs forever for the life of the thread. Bounded by channel membership in practice (small), but a chatty long-lived thread (a permanent support channel, a standup thread reused for weeks) accumulates everyone who ever posted.

Not worth changing the data shape over, but worth a comment that the gate only reads `length > 0` + `filter(u => u !== currentUser).length > 0` — i.e. it really only needs to know "have any humans other than the current sender posted?" — so an `O(1)` flag (`hasOtherHuman: boolean`) would do the same work with bounded storage. Optional; the array carries no real cost at expected thread sizes.

src/session/manager.ts
```diff
@@ -62,6 +62,95 @@ export class SessionManager {
...
+   * Auto-trigger channels (any channel with a `channelDefaults` entry — e.g.
+   * `#bugs-backlog`) are exempt: humans-talking-to-humans is the intended
+   * input there, not a sidebar.
+   */
+  async gateAttention(event: SlackMessageEvent): Promise<boolean> {
+    // !aside: drop the message, no state change. Anyone in the thread.
+    if (event.command === "aside") {
+      this.onReaction?.(event, "eyes");
+      return true;
+    }
```
PranavBakre, line 102: 🟢 **Nice-to-have — `!aside` skips participant tracking.**

The `!aside` early-return at lines 87-90 returns before `getOrCreateSession` runs, so a user who *only* sends asides into a thread is never added to `humanParticipants`. The trigger rule depends on `humanParticipants` being populated — so a thread where U-A posts a real message, then U-B posts only `!aside`s, then U-B posts a real non-@-mention message would *not* fire the trigger on U-B's first real post (because U-B was never added to `humanParticipants` by the aside-only path).

Real-world impact is small — people typically don't use threads exclusively for asides — but mentally cleaner to add U-B to participants on `!aside` too, so the gate's bookkeeping matches what humans observe. A one-liner inside the aside branch would do it.

Not a blocker; the user's *next* non-aside message would add them and the trigger would fire on whoever-comes-after that. Worth a thought, no strong call.

pranav-growthx (approved):
## Re-review: commit a58e6e7 addresses all prior findings

Per re-review protocol, scoped to the new commit and verification of the three findings from the prior review (PR #28#pullrequestreview-4291298083).

### Fixes verified

- **🟡 app_mention asymmetric flag population** — fixed at `src/slack/events.ts:213-217`. The handler now extracts `botId` symmetrically with the `message` handler (`"bot_id" in event && typeof ... === "string" ? ... : undefined`) and derives `isSelfBot = !!(botId && selfBotId && botId === selfBotId)`. New test at `src/session/manager.test.ts:962-997` (`does not add foreign bots to humanParticipants even on @mention`) covers the foreign-bot app_mention path that was previously untested.

- **🟢 humanParticipants unbounded** — addressed at `src/session/types.ts:60-71` with an inline comment explaining the design tradeoff (`O(1)` boolean would suffice but the array carries debug observability via `!status`, room for future per-user policy, bounded in practice by channel membership). Reasonable — not worth a data-shape change.

- **🟢 !aside skips participant tracking** — fixed at `src/session/manager.ts:91-102`. The `aside` branch now resolves the session and pushes the sender to `humanParticipants` (guarded by `isHuman` and dedup via `includes`) before reacting and returning. Matches the bookkeeping invariant.

### New commit additions (commit a58e6e7)

- Defensive `case "aside"` and `case "listen"` in `handleCommand` (`src/session/manager.ts:550-562`) prevent fall-through to Claude if the gate is bypassed (e.g. direct `handleMessage` invocation from tests). Mirrors the gate's behavior — no state divergence.

- One new test (`does not add foreign bots to humanParticipants even on @mention`) — verifies the exact failure mode the prior review flagged. Assertion is correctly framed against the contract ("B-FOREIGN is not in humanParticipants regardless of which path it took").

### Verified

- Per-commit posture: commit 1 (`a80e006`) + commit 2 (`a58e6e7`). Both reviewed.
- `bun test` — 328 pass / 1 fail (pre-existing DevServerManager flake on main, called out in PR description).
- All 13 gate-related tests pass; one new test added in commit 2.

### Side-observation (not blocking)

The `app_mention` handler still doesn't set `botUsername`, while the `message` handler does (`src/slack/events.ts:177-180`). With `isSelfBot=true` now correctly set in the app_mention path, `supportRouter.handleMessage` will resolve `sourceAgent = agentForUsername(undefined) = null` for self-bot app_mention events, and short-circuit at `router.ts:183` (`event.isSelfBot && sourceAgent === null`). Before this PR that path went through unfiltered because `isSelfBot` was undefined. This is most likely an improvement (no self-mention loops), but it's a behavior change for self-mentions in non-auto-trigger channels — worth a glance if anyone runs into "Junior stopped responding to its own continuations." Not introduced by this PR's intent (the asymmetric `botUsername` was already there); flagging for future awareness, not blocking.

### Verdict

Two clean re-review passes on the deltas. The prior findings are addressed precisely as suggested, no new issues introduced. Approved.

PR #29, opened 2026-05-14
feat(driver): tmux driver, swappable with headless

PranavBakre:
## Summary

Implements the interactive driver scoped in #26. \`claude -p\` (headless) moves to API credits under the new Anthropic terms; the persistent TUI under tmux stays on Max. This PR adds a second driver path so threads can run either, gated by \`DEFAULT_CLAUDE_DRIVER\` and \`!driver headless|tmux\`.

**Iter 0** — \`scripts/tmux-smoke-test.sh\` validates tmux >= 3.4, paste-buffer -p, transcript JSONL appearance, \`system.turn_duration\` event, --resume continuity. The smoke test informed a design shift: the original Stop-hook + sentinel-file design was replaced by reading \`system.subtype: \"turn_duration\"\` straight off the transcript. Single signal channel.

**Iter 1+2** — \`ClaudeDriver\` interface + \`HeadlessDriver\` (wraps the existing spawner with zero behavior change) + \`TmuxDriver\` (new). Manager picks via \`session.driverMode\`. Per-(threadId, agentName) keying for tmux session names.

**Iter 3+4** — Boot-time reconciliation (\`tmux has-session\` for each persisted tmux row, adopt or downgrade) + idle eviction sweep (\`TMUX_IDLE_TTL_MS\` default 4h, \`TMUX_SWEEP_INTERVAL_MS\` default 15min).

**Iter 5** — \`!stop\` sends Escape; \`!driver headless|tmux\` switches mode. \`!reset\` tears down the tmux session.

**Iter 6** — \`transcript-adapter.ts\` translates JSONL → \`StreamEvent\` and flags \`turn_duration\` as \`turnDone\`. Unit tests for adapter + driver (multi-agent isolation).

**Iter 7** — Config-only wiring through \`createDrivers\`, factory, and updated CLAUDE.md / project-setup.md.

**Review fixes (final commit)** —
- \`tmux-reconcile\` / \`tmux-evict\` previously hardcoded \`agentName: \"lead\"\`. Default-Junior tmux threads (non-support channels use \"default\") would orphan their session on reconcile and silently skip eviction. Now persists \`topLevelTmuxAgent\` on the session row.
- \`fs.watch\` on macOS misses some create events for files appearing in a watched directory. Added a 250ms poll fallback alongside the watcher that stops once the transcript path is located (or after ~30s).

## Risks / known gaps

- **Iter 0 was not executed against live \`tmux\` + \`claude\`** during this branch — the smoke-test script exists and the design was informed by inspecting existing transcript JSONL files. The first real soak under \`DEFAULT_CLAUDE_DRIVER=tmux\` may surface issues the unit tests don't catch.
- Iter 7 \"flip the default to tmux\" is a config change pending a real soak; the code paths are in place but \`DEFAULT_CLAUDE_DRIVER\` still defaults to \`headless\`.

## Test plan

- [x] \`bun run typecheck\` — clean
- [x] \`bun test\` — 330 pass (1 pre-existing flaky integration test unrelated to driver)
- [ ] Run \`scripts/tmux-smoke-test.sh\` once with a live \`claude\` binary
- [ ] Flip a single test thread to \`!driver tmux\`, send a few turns, restart bot, confirm reconciliation adopts the live session
- [ ] Verify \`!stop\` interrupts mid-turn without killing tmux
- [ ] Verify eviction sweep kills an idle tmux session past TTL


src/session/manager.ts
```diff
@@ -156,6 +173,16 @@ export class SessionManager {
         handle.kill();
         this.handles.delete(key);
       }
+      // Tear down any tmux state for this thread before deleting the row.
+      // The row is the only place we track tmuxSessionName, so close drivers
+      // first or we leak detached tmux sessions.
+      if (session.driverMode === "tmux") {
+        const tmux = this.driverFor("tmux");
+        await tmux.close(threadId, "lead").catch(() => undefined);
```
pranav-growthx: 🔴 **Blocker — Bug / Resource Leak.** `resetSession` hardcodes `agentName: "lead"` when closing the tmux session, but the top-level agent is `"default"` on non-support channels (set at line 720, `runSingleSession` passes `"default"` at line 79). For a default-Junior thread on tmux, `tmux.close(threadId, "lead")` looks up key `(threadId, "lead")` in the driver's `sessions` map, finds nothing, and silently returns at tmux-driver.ts:145 — the live tmux session for `(threadId, "default")` is **leaked** when `!reset all` runs.

This is the exact symmetric bug the final review commit (fe0d0f9) fixed in `tmux-reconcile.ts` and `tmux-evict.ts`, but `resetSession` was missed.

**Fix:** read `session.topLevelTmuxAgent ?? "lead"` instead of the literal:
```ts
const topAgent = session.topLevelTmuxAgent ?? "lead";
await tmux.close(threadId, topAgent).catch(() => undefined);
```

src/session/manager.ts
```diff
@@ -439,6 +468,61 @@ export class SessionManager {
...
+          this.onCommandResponse?.(event, `Already on *${arg}*.`);
+          return true;
+        }
+        // Tear down the outgoing driver's per-(thread,agent) state.
+        const outgoing = this.driverFor(session.driverMode);
+        for (const key of this.handles.keys()) {
+          if (!key.startsWith(`${session.threadId}:`)) continue;
+          const agentName = key.slice(session.threadId.length + 1);
+          await outgoing.close(session.threadId, agentName).catch(() => undefined);
+        }
```
pranav-growthx, line 657: 🔴 **Blocker — Bug / Resource Leak.** The `!driver` command only walks `this.handles` (the in-flight handle map) to tear down the outgoing driver. But the typical case is an admin running `!driver tmux` → `!driver headless` between turns, when no message is in-flight — `this.handles` is empty (entries are deleted in `onRunComplete`, line 816). The tmux session persists between turns by design (that's the whole point of the driver), so the iteration closes nothing. Then line 519–520 sets `session.driverMode = "headless"` and `session.tmuxSessionName = null`.

Result: the live tmux session is unreachable from sqlite (`tmuxSessionName` is nulled), and the eviction sweep skips it (`tmux-evict.ts:30` filters `driverMode !== "tmux"`). **The tmux session leaks until manual `tmux kill-server` or host reboot.**

**Fix:** when switching away from tmux, also close the top-level + per-agent tmux sessions the way `tmux-evict.ts` does — walk `session.topLevelTmuxAgent` and `Object.values(session.agentSessions)`, not just `this.handles`:
```ts
if (session.driverMode === "tmux" && arg !== "tmux") {
  const tmux = this.driverFor("tmux");
  if (session.topLevelTmuxAgent) {
    await tmux.close(session.threadId, session.topLevelTmuxAgent).catch(() => undefined);
  }
  for (const agentSession of Object.values(session.agentSessions ?? {})) {
    if (agentSession.tmuxSessionName) {
      await tmux.close(session.threadId, agentSession.agentName).catch(() => undefined);
    }
  }
  session.topLevelTmuxAgent = null;
}
```

src/claude/tmux-driver.ts
```diff
@@ -0,0 +1,507 @@
...
+      watcher: null,
+      pollTimer: null,
+      activeTurn: null,
+      startedAt: Date.now() - 86_400_000, // Adopt anything that already exists.
+    };
+    this.sessions.set(key, sess);
+    if (transcriptPath && existsSync(transcriptPath)) {
+      sess.transcriptOffset = statSync(transcriptPath).size;
+      this.startTranscriptWatch(sess);
+    }
```
pranav-growthx, line 237: 🔴 **Blocker — Bug.** `adoptExistingSession` only starts the transcript watcher when `transcriptPath` is non-null AND `existsSync(transcriptPath)` is true. But on bot boot, an alive-tmux session whose `sessionId` is null on the sqlite row (e.g., the bot crashed before the first `system.init` was drained), OR whose transcript path on disk doesn't exist yet, leaves the adopted session **without a watcher and without a poll timer**.

The next `send()` calls `ensureSession`, which sees the entry already in `this.sessions` map AND `tmuxHasSession` returns true → returns at line 215 without calling `startTranscriptWatch`. The next turn fires, claude writes to the transcript, but the driver never observes `system.turn_duration` → the turn never resolves and hangs until `timeoutMs` (default 5 min).

**Fix:** always call `startTranscriptWatch(sess)` in `adoptExistingSession` (the watcher + poll fallback are designed to handle a not-yet-existent transcript dir/file):
```ts
this.sessions.set(key, sess);
if (transcriptPath && existsSync(transcriptPath)) {
  sess.transcriptOffset = statSync(transcriptPath).size;
}
this.startTranscriptWatch(sess); // unconditional
```

src/lifecycle/tmux-reconcile.ts
```diff
@@ -0,0 +1,87 @@
...
+      const live = await driver.tmuxHasSession(session.tmuxSessionName);
+      if (live) {
+        await driver.adoptExistingSession({
+          threadId,
+          agentName: session.topLevelTmuxAgent ?? "lead",
+          cwd,
+          tmuxSessionName: session.tmuxSessionName,
+          sessionId: session.sessionId,
+        });
+        adopted++;
```
pranav-growthx, line 56: 🟡 **Should-fix — Bug / Stuck State.** When the bot crashes mid-turn with `status: "busy"` and tmux survives, this branch adopts the session but never resets `status` from `busy` → `idle`. The symmetric branch (lines 48–55) already handles this for the dead-tmux case.

After adoption, there's no `activeTurn` on the driver-side `TmuxSession` (line 188 sets `activeTurn: null`), so when the in-flight turn's `system.turn_duration` event lands on disk and `drainTranscript` reads it → `processTranscriptLine` checks `const turn = sess.activeTurn; if (!turn) return;` (line 385) → silently drops it. The sqlite row stays `busy` forever; new messages get buffered into `pendingMessages` (manager.ts:148–155) and never drain.

**Fix:** mirror the downgraded branch in the live branch:
```ts
if (live) {
  await driver.adoptExistingSession({ ... });
  if (session.status === "busy") {
    session.status = "idle";
    await store.set(threadId, session);
  }
  adopted++;
}
```
Same for the per-agent loop at lines 62–82.

src/claude/tmux-driver.ts
```diff
@@ -0,0 +1,507 @@
...
+      // Stop polling once the transcript is located, or after ~30s.
+      if ((sess.transcriptPath && existsSync(sess.transcriptPath)) || ticks >= 120) {
+        if (sess.pollTimer) {
+          clearInterval(sess.pollTimer);
+          sess.pollTimer = null;
+        }
+      }
+    }, 250);
+    // Kick off an initial drain in case events were already on disk.
+    onChange();
```
pranav-growthx, line 412: 🟡 **Should-fix — Concurrency.** `startTranscriptWatch` wires both `fs.watch` (line 308) and a 250ms `setInterval` poll (line 314) into the same `onChange` → `drainTranscript`. `drainTranscript` is async: it reads `sess.transcriptOffset`, awaits `file.open`, awaits `file.stat`, reads bytes, then writes the new offset back.

Two concurrent invocations (e.g., poll tick and watcher event firing within the same loop) both observe the old offset, both decode the same byte range, both invoke `processTranscriptLine` for every line in that range → duplicate listener fires → duplicate Slack status updates streamed to the channel. The double-resolve of `turn_duration` is guarded by `turn.rejected`, but duplicate `assistant` events aren't.

**Fix:** serialize with an in-progress flag on `TmuxSession`:
```ts
if (sess.draining) return;
sess.draining = true;
try { /* existing body */ } finally { sess.draining = false; }
```

src/claude/tmux-driver.ts
```diff
@@ -0,0 +1,507 @@
...
+    const dir = join(this.projectsRoot, encodeCwd(sess.cwd));
+    mkdirSync(dir, { recursive: true });
+    const onChange = () => {
+      this.drainTranscript(sess).catch((err) => {
+        console.warn("[tmux-driver] transcript drain error:", err);
+      });
+    };
+    // Watch the directory — when the transcript file is created or appended,
+    // fs.watch fires. On macOS this requires recursive: false (default).
+    const watcher = fsWatch(dir, { persistent: false }, onChange);
```
pranav-growthx, line 389: 🟢 **Nice-to-have.** `fs.watch` returns an `FSWatcher` that can emit `'error'` events — e.g., if the watched directory is removed (worktree teardown while a tmux session is being reconciled). Without an `error` listener, an emitted error becomes an unhandled `'error'` event on the EventEmitter and **crashes the process**.

Add `watcher.on('error', (err) => { console.warn('[tmux-driver] watcher error:', err); })` after line 308.

pranav-growthx (changes requested):
Reviewed in detail with 6-pass methodology. Found 3 blockers, 3 should-fix items, 1 nice-to-have — all in the manager.ts/tmux-driver.ts/tmux-reconcile.ts code paths. Inline comments below. Most blockers are symmetric to the bug the final review commit (fe0d0f9) caught (`topLevelTmuxAgent` hardcoded as "lead") — same class, different sites.

pranav-growthx:
One additional finding that can't be posted inline (the affected line is not in the diff, but the bug is induced by this PR's tmux state on the session row):

🟡 **Should-fix — Resource Leak in `resetAgent` (src/session/manager.ts:198–226).** When `agentName` is `"lead"` or `"default"` (lines 210–218), the handler clears sessionId / pendingMessages / status but doesn't tear down the tmux session for the top-level agent and doesn't clear `tmuxSessionName` / `topLevelTmuxAgent`. The `handle.kill()` at line 205 invokes TmuxDriver's `kill()`, which sends Escape but does NOT call `tmux kill-session` (by design — `kill` is interrupt, `close` is teardown).

Result: `!reset lead` on a tmux thread leaves a live tmux session whose pointer (`tmuxSessionName`) is still on the row. The next message will reuse it — and now claude's POV is "fresh conversation" (sessionId cleared) but the TUI still has the previous turn's artifacts in scroll/state, which is confusing. Worse, if `resetAgent` is followed by a `!driver headless` switch (which has the leak called out in the inline review on manager.ts:518), the row loses `tmuxSessionName` and the tmux session leaks permanently.

**Fix:** when `agentName` is the top-level agent and `driverMode === "tmux"`, call `tmux.close(threadId, agentName)` and clear `tmuxSessionName` + `topLevelTmuxAgent`. Same pattern as the suggested fix for `resetSession`.

src/session/manager.ts
```diff
@@ -439,6 +497,82 @@ export class SessionManager {
...
```
pranav-growthx, line 657: 🟡 **Should-fix — `!driver` mid-turn now triggers a junk partial Slack post + auto-drain on the new driver.**

The new symmetric resolve in `TmuxDriver.close()` (tmux-driver.ts:154-164) is correct, but it exposes a pre-existing asymmetry: this loop calls `outgoing.close(...)` without `this.handles.delete(key)`. Trace when an admin runs `!driver headless` while a tmux turn is in flight:

1. `outgoing.close()` → TmuxDriver.close synchronously runs `turn.resolve({ error: 'driver-closed', response: lastAssistantText, ... })` at tmux-driver.ts:163, queuing the `.then` microtask registered at manager.ts:815.
2. Next line in close is `await this.killTmuxSession(...)` — that yield drains microtasks, so `onRunComplete` starts. The guard at manager.ts:863 (`this.handles.get(handleKey) !== ownHandle`) does **not** trip — the handle is still in the map — so `onRunComplete` proceeds.
3. `onRunComplete` fires `onResponse` with the partial `lastAssistantText` (Slack post), fires `onError` with `"driver-closed"`, and then auto-drains buffered messages by re-entering `runClaudeWithAgent` on the new driver.

Result: contrary to the user-facing message `"Driver set to *headless*. Next message starts on the new path."` (line 572), the next message gets auto-sent on the new driver and the partial mid-turn output is posted as if it were complete.

The other teardown paths handle this correctly:
- `resetSession` (line 174): `this.handles.delete(key)` before tmux.close → guard trips, onRunComplete bails.
- `resetAgent` (line 212): same — delete before close.
- `!stop` (line 512): same — delete before tmux ops.

Fix is one line — mirror the others:
```ts
for (const key of this.handles.keys()) {
  if (!key.startsWith(`${session.threadId}:`)) continue;
  const agentName = key.slice(session.threadId.length + 1);
  await outgoing.close(session.threadId, agentName).catch(() => undefined);
  this.handles.delete(key);  // ← add this
}
```

This is a should-fix, not a blocker — it only manifests when an admin runs `!driver` mid-turn, which is an edge case. The primary deadlock fix is needed and correct.

pranav-growthx (changes requested):
Re-review of fix commit 65f3e24.

Six of seven prior findings are correctly addressed. The new symmetric `TmuxDriver.close()` resolve is necessary to avoid the deadlock, and the regression tests pin the (threadId, topLevelTmuxAgent) close call across three teardown paths. However, the symmetric resolve unmasks a pre-existing inconsistency in `!driver`: unlike `!reset`/`!stop`/`resetAgent`, the `!driver` command does **not** delete entries from `this.handles` before calling `close()`. Trace below.

One should-fix; rest is clean. Not approving on this pass.

PranavBakre (commented):
**LGTM — two consecutive clean passes on f130e7a.**

(Posting as comment since PranavBakre authored this PR and can't self-approve.)

Verified for f130e7a:
- Handle is deleted before `outgoing.close()` in the !driver loop, so when tmux's symmetric close-resolve fires the in-flight activeTurn, `onRunComplete`'s `handles.get(key) !== ownHandle` guard trips and the discarded turn bails without posting partials or auto-draining.
- `[...this.handles.keys()]` snapshot is correct — required because the loop body now mutates `this.handles` via `delete` before each `await`.
- Other reset paths (`resetSession`, `resetAgent`, `!stop`) maintain the delete-before-close invariant; !stop uses interrupt+kill (Escape only), which doesn't resolve activeTurn, so no race there.
- Headless→tmux direction unaffected: HeadlessDriver.close is a no-op, so the reordering is harmless when outgoing is headless.

src/session/manager.ts
```diff
@@ -543,6 +601,88 @@ export class SessionManager {
...
+              agentSession.tmuxSessionName = null;
+            }
+          }
+        }
+        session.driverMode = arg;
+        if (arg !== "tmux") {
+          session.tmuxSessionName = null;
+          session.topLevelTmuxAgent = null;
+        }
+        await this.store.set(session.threadId, session);
```
anmolm-growthx, line 691: **Bug: `!driver` during an in-flight turn leaves the thread stuck `busy` forever.**

`handleCommand` runs BEFORE the `status === "busy"` gate in `runSingleSession` (lines 242–245), so `!driver` is reachable mid-turn. The loop above deletes the handle, then `outgoing.close()` resolves the activeTurn with `driver-closed` — but `onRunComplete` then bails on the `handles.get(key) !== ownHandle` guard (line 991). Net result: `session.status` stays `busy`, no drain fires, every subsequent message gets buffered. Only `!cancel`/`!reset` recovers.

Compare `!stop` (line 619) and `!cancel` (line 392) — both explicitly flip `status = "idle"` after killing handles. Same flip is missing here. Suggest:

```ts
session.driverMode = arg;
if (arg !== "tmux") {
  session.tmuxSessionName = null;
  session.topLevelTmuxAgent = null;
}
session.status = "idle";
session.pendingMessages = [];  // optional, depending on desired UX
for (const agentSession of Object.values(session.agentSessions ?? {})) {
  if (agentSession.status === "busy") agentSession.status = "idle";
}
```

The seeded test at `manager.test.ts:835` doesn't catch this because it doesn't put an in-flight handle in the map. Worth adding a regression test.

src/claude/tmux-driver.ts
```diff
@@ -0,0 +1,558 @@
...
+
+  // ─────────────────────────────────────────────────────────────────────
+  // Internals
+  // ─────────────────────────────────────────────────────────────────────
+
+  private async ensureSession(input: DriverSendInput): Promise<void> {
+    const key = handleKey(input.threadId, input.agentName);
+    const existing = this.sessions.get(key);
+    if (existing && (await this.tmuxHasSession(existing.name))) {
+      return;
```
anmolm-growthx, line 291: **Wedged-pane risk: `tmuxHasSession` succeeds even if `claude` exited inside the pane.**

If the interactive `claude` process crashes (OOM, SIGSEGV, manual kill from inside the TUI), tmux keeps the session alive — just dropped back to a shell prompt. `ensureSession` only checks `tmuxHasSession(existing.name)` and reuses it. Subsequent `paste-buffer + Enter` then goes to a dead shell. Each turn times out at `CLAUDE_TIMEOUT_MS` (5min default) and resolves with a timeout error; the wedge persists until someone runs `!reset`.

Worth either (a) checking that the active pane's PID is `claude` (e.g., `tmux display -t <name> -p '#{pane_current_command}'`), or (b) treating any timeout in a tmux thread as a signal to tear down + relaunch with `--resume`. The `withTimeout` callback at `manager.ts:917` is the natural hook for the latter.

src/lifecycle/tmux-evict.ts
```diff
@@ -0,0 +1,73 @@
...
+  const evicted: string[] = [];
+
+  for (const [threadId, session] of sessions) {
+    if (session.driverMode !== "tmux") continue;
+
+    if (session.tmuxSessionName && now - session.lastActivity > opts.ttlMs) {
+      if (opts.skipBusy && session.status === "busy") continue;
+      const topAgent = session.topLevelTmuxAgent ?? "lead";
+      try {
+        await driver.close(threadId, topAgent);
```
anmolm-growthx, line 44: **Race: eviction sweep can kill a freshly-started turn.**

The `skipBusy` check reads `session.status`, then `driver.close()` runs at `await`-point microtasks later. Between the read and the kill, a new Slack message can pass the manager's idle check (line 249) and start a turn via `runClaudeWithAgent` — which then gets its tmux session ripped out from underneath it. close() resolves the new activeTurn with `error: "driver-closed"`, user sees an error, conversation continues on a cold-start.

Probability is low (15-min sweep cadence vs. a millisecond-wide window) but visible when it hits. Cheapest mitigation: re-read the row inside the loop body right before `close()` and skip if status flipped to busy. Could also use a per-thread mutex shared with `runClaudeWithAgent`.

src/claude/tmux-driver.ts
```diff
@@ -0,0 +1,558 @@
...
+
+    if (adapted.turnDone) {
+      const result: SpawnResult = {
+        sessionId: sess.sessionId,
+        response: turn.resultText || turn.lastAssistantText,
+        events: turn.events,
+        exitCode: 0,
+        error: null,
+      };
+      sess.activeTurn = null;
```
anmolm-growthx: **Stale `turn_duration` can clobber the next turn's `activeTurn`.**

`sess.activeTurn = null` runs unconditionally — even when `turn.rejected` is true (because `kill()` / `close()` already resolved it). Scenario:

1. Turn 1 in flight, claude is writing to transcript.
2. `!cancel` → `handle.kill()` → `turn1.rejected = true`, Escape sent. Tmux session survives.
3. User sends new message → `runClaudeWithAgent` → `send()` → `ensureSession` reuses, sets `sess.activeTurn = turn2`.
4. Meanwhile claude was still flushing the cancelled turn1's transcript; `turn_duration` line arrives.
5. `drainTranscript` reads it, captures `turn = sess.activeTurn` = `turn2`. Hits `turnDone` branch.
6. `sess.activeTurn = null` nukes turn2's binding. `turn.resolve(result)` runs against turn2 (turn2.rejected is false) — turn2 resolves prematurely with the empty result of turn1.

In practice the transcript-offset advances past turn 1's lines synchronously, so the realistic failure mode is narrower (delayed disk flush or the rare case where turn1's turn_duration sits in the same chunk as a turn2 kickoff event). Still cheap to harden: capture `const turn = sess.activeTurn` and only null out if `sess.activeTurn === turn`, or check `turn.rejected` before nulling.

src/lifecycle/tmux-reconcile.ts
```diff
@@ -0,0 +1,104 @@
...
+        }
+        await store.set(threadId, session);
+        downgraded++;
+      }
+    }
+  }
+
+  log.info("reconcile", `tmux adopted=${adopted} downgraded=${downgraded}`);
+  return { adopted, downgraded };
+}
```
anmolm-growthx, line 104: **Missing tests for the lifecycle paths.**

The review-fix bug (`topLevelTmuxAgent` defaulting to literal `"lead"`) lived right here — and there's no test covering reconciliation or eviction. The manager tests added in this PR cover the `!reset`/`!driver` close paths, but not boot-time adopt/downgrade or the idle sweep. Given how easy these are to test with a fake `TmuxDriver` (the manager.test.ts `makeFakeDrivers` pattern is right there), a few smoke tests would have caught the original bug and would protect against regressions like the migration-fallback edge cases.

At minimum: (a) reconcile picks up rows where `topLevelTmuxAgent` is `null` (pre-migration shape) without crashing; (b) reconcile flips `status: busy` → `idle` on adopt; (c) evict skips a row that flipped to busy mid-sweep (mitigates the race I flagged separately).

Non-blocking — code looks right, just under-tested.

src/claude/tmux-driver.ts
```diff
@@ -0,0 +1,558 @@
...
+      },
+      pid: null,
+    };
+  }
+
+  async interrupt(threadId: string, agentName: string): Promise<void> {
+    const sess = this.sessions.get(handleKey(threadId, agentName));
+    if (!sess) return;
+    await this.sendKeys(sess.name, "Escape", "Escape");
+  }
```
anmolm-growthx, line 162: **`interrupt()` is asymmetric with `kill()` — doesn't resolve the active turn.**

`kill()` (line 118-135) sends Escape AND resolves the active turn's promise. `interrupt()` only sends Escape. The result is that any caller invoking `interrupt()` alone (currently only `!stop` at `manager.ts:614`, which then also calls `handle.kill()` immediately after, so it's fine in practice) would leave the turn promise pending forever.

The mismatch is brittle — if a future caller uses `interrupt()` without the kill follow-up (e.g., an admin slash-command, or wired into a Slack reaction), the thread will wedge until `withTimeout` fires 5min later. Either document that `interrupt()` MUST be paired with `handle.kill()`, or have `interrupt()` resolve the active turn too (symmetric with `close()` which does).

anmolm-growthx (commented):
**Verdict:** ❌ Blocking (1 issue)

The driver abstraction is well-shaped — clean interface, factory swap, headless preserved as a thin wrapper, transcript-adapter is a tidy translation layer. The review-fix on `topLevelTmuxAgent` shows the right instinct (root-cause the literal "lead" assumption, not patch the call sites). Worth landing once the stuck-busy state on `!driver` is addressed.

### Blocking

1. **`!driver` mid-turn leaves `session.status = "busy"` permanently** (`src/session/manager.ts:631-684`). `handleCommand` runs before the busy gate in `runSingleSession`, so the command is reachable while a turn is in flight. The teardown loop intentionally bails out of `onRunComplete` via the `handles.get(key) !== ownHandle` guard, but nothing flips status back to idle — every subsequent message gets buffered with no drain to release them. `!stop` and `!cancel` both do this flip; `!driver` should too.

### Non-blocking

1. **Wedged-pane risk** (`src/claude/tmux-driver.ts:236`). `tmuxHasSession` succeeds even if `claude` exited inside the pane (process crash, internal `/exit`, etc.). The bot keeps pasting into a dead shell until 5min timeout each turn. Suggest checking `#{pane_current_command}` or treating tmux-mode timeouts as a teardown signal.
2. **Eviction race** (`src/lifecycle/tmux-evict.ts:32-46`). `skipBusy` check is stale by the time `driver.close()` runs; a turn that starts in the gap gets its tmux killed mid-flight. Re-read the row right before close, or share a per-thread mutex.
3. **Stale `turn_duration` clobbers next turn** (`src/claude/tmux-driver.ts:461`). `sess.activeTurn = null` runs unconditionally, including when the line arrives after kill+next-send. Narrow window in practice but cheap to harden.
4. **`interrupt()` ≠ `kill()`** (`src/claude/tmux-driver.ts:140`). `interrupt()` only sends Escape; doesn't resolve the active turn. Today only `!stop` uses it, and it pairs with `handle.kill()`. Brittle if a future caller skips the kill.
5. **No tests for `tmux-reconcile.ts` / `tmux-evict.ts`** — the lifecycle code where the `topLevelTmuxAgent` regression actually lived has zero coverage. `manager.test.ts:makeFakeDrivers` is the pattern.

### What's solid

- The headless wrapper is genuinely zero-behavior-change — minimal seam, existing tests pass unchanged.
- `topLevelTmuxAgent` persistence + migration default is the right shape; sqlite's `normalizeSession` handles pre-PR rows cleanly.
- `drainTranscript` serialization (`drainInFlight` / `drainPending`) catches the fs.watch-vs-poll race that would otherwise double-dispatch assistant events.

PR #30, opened 2026-05-15
feat: add runner provider abstraction

PranavBakre:
## Summary
- Introduce a normalized runner provider boundary so session management and Slack formatting no longer depend on Claude-native events.
- Add OpenCode args, config, parser, and spawner support behind the new provider selection path.
- Persist provider state across sessions and expose provider selection/status in Slack commands, home output, and docs.

## Test plan
- [x] bun run typecheck
- [x] git diff --check
- [x] bun test (full suite: 361 pass, 0 fail)
- [x] bun test (full suite, second clean run: 361 pass, 0 fail)


src/session/store/sqlite.ts
```diff
@@ -202,11 +216,25 @@ export class SqliteSessionStore implements SessionStore {
...
+      .all();
+    if (columns.some((column) => column.name === "provider")) return;
+    this.db.run(
+      "ALTER TABLE agent_sessions ADD COLUMN provider TEXT DEFAULT 'claude'",
+    );
+  }
 }
 
 function normalizeSession(session: ThreadSession): ThreadSession {
+  session.provider = normalizeRunnerProvider(session.provider);
```
anmolm-growthx: **🟡 Should-fix — codex defense-in-depth on persisted load.**

`normalizeRunnerProvider` accepts `"codex"` (it's in `RunnerProvider`) and round-trips it through the store unchanged. The PR description claims rejection at every entry point, but persisted-state load is the third boundary the contract names. A row with `provider: "codex"` (or `agent_sessions.provider = "codex"`) loads cleanly today — and `sqlite.test.ts:69` even asserts that round-trip — so the next spawn throws "Codex runner provider is not implemented yet" inside `runRunnerWithAgent`'s catch with the cause buried in a generic "setup" `lastError`.

In practice nothing writes `codex` via normal paths today (parseRunnerProvider rejects, !provider rejects, init events only emit `claude|opencode`), so this is theoretical. But it matches exactly the deferred-fix pattern Pranav watches for: the rejection is at the user-facing boundary, the storage layer is permissive. Suggest either (a) `normalizeRunnerProvider` collapses unimplemented planned providers to `"claude"` (with a one-shot warn log on conversion), or (b) the sqlite load path runs `isImplementedRunnerProvider` and demotes/warns symmetrically with config.

src/opencode/parser.ts
```diff
@@ -0,0 +1,200 @@
...
+          provider: "opencode",
+          text: event.part.text,
+        });
+      } else if (event.type === "step_finish" && event.part?.tokens) {
+        events.push({
+          type: "done",
+          provider: "opencode",
+          usage: event.part.tokens,
+        });
+      }
```
anmolm-growthx, line 203: **🟡 Should-fix — `done` is gated on `step_finish.part.tokens`.**

The Claude mapper unconditionally emits `done` on every `result` event (spawner.ts:158). The OpenCode mapper only emits `done` when `event.part?.tokens` is truthy. If OpenCode ever emits a `step_finish` without tokens (early termination, rate-limited turn, future protocol drift), no normalized `done` event fires — downstream code that listens for `done` (and any future "post-turn" hook) silently misses it. Tests only cover the happy-path with tokens.

Safer: always emit `done` on `step_finish`, with `usage` optional:
```ts
else if (event.type === "step_finish") {
  const usageEvent: RunnerEventDone = { type: "done", provider: "opencode" };
  if (event.part?.tokens) usageEvent.usage = event.part.tokens;
  events.push(usageEvent);
}
```

src/session/manager.ts
```diff
@@ -393,6 +401,58 @@ export class SessionManager {
...
+        // (reproducer/thinker/lead/etc). Switching provider while ANY of
+        // these is set leaves mixed-provider state in one thread — the new
+        // provider drives fresh dispatches while existing agents continue
+        // resuming against the old provider. Block all three.
+        const hasAgentSession = Object.values(session.agentSessions ?? {}).some(
+          (a) => a.sessionId,
+        );
+        const hasNativeSession = Boolean(
+          session.sessionId || session.leadSessionId || hasAgentSession,
+        );
```
anmolm-growthx, line 615: **🟡 Should-fix — `!provider` guard misses busy worker agents.**

The busy check on line 421 only considers `session.status !== "idle"`. But `handleAgentMessage` for a persistent agent (reproducer/thinker/echo/etc.) sets `agentSession.status = "busy"` and *does not* touch `session.status` (manager.ts:105–107). So `!provider` succeeds while a worker is mid-turn.

The new `hasAgentSession` check on line 436 catches it via `a.sessionId`, but only after that worker's first `init` event has landed. A first-turn-in-flight worker (sessionId still null, status busy) slips through and you end up flipping `session.provider` while a child run is alive — the next dispatch goes to the new provider while the in-flight worker continues against the old one.

Fix: extend the busy check to also cover agent sessions:
```ts
const busyAgent = Object.values(session.agentSessions ?? {}).some(
  (a) => a.status === "busy",
);
if (session.status !== "idle" || busyAgent) {
  this.onCommandResponse?.(event, "Cannot change provider while a runner is active. Use `!cancel` first.");
  return true;
}
```

src/opencode/parser.ts
```diff
@@ -0,0 +1,200 @@
...
+    },
+    map(event: OpenCodeEvent): RunnerEvent[] {
+      const events = maybeInit(event);
+
+      if (event.type === "text") {
+        response += event.part.text;
+        events.push({
+          type: "message",
+          provider: "opencode",
+          text: event.part.text,
```
anmolm-growthx: **🟡 Should-fix — OpenCode `message` events are per-fragment, not per-turn.**

Claude's mapper aggregates an entire assistant turn's text into a single `message` runner event (spawner.ts:142–155). OpenCode's mapper here emits one `message` event per text part as it streams in. Two consequences:

1. **Slack rate-limit risk.** `index.ts:96` calls `responder.updateStatus` for every `message` event. For a long OpenCode response that streams in N chunks, that's N updates per turn vs Claude's 1.
2. **Live status only shows the latest fragment.** Each updateStatus overwrites the prior status line, so the user sees the trailing chunk instead of the accumulated text. With Claude they see the whole assistant message preview.

`SpawnResult.response` is correctly the full concatenated text (mapper accumulates), so the *final* posted reply is fine — it's only the in-flight live status that's degraded.

Fix options: (a) coalesce text fragments in the mapper and emit one `message` event per `step_finish` (or some bounded buffer), or (b) document this as expected OpenCode behavior in `runner-providers.md` so operators aren't surprised. Option (a) keeps the contract symmetric across providers.

src/config.ts
```diff
@@ -97,6 +113,20 @@ export function loadConfig(): Config {
...
+    opencode: {
+      model: process.env.OPENCODE_MODEL ?? null,
+      timeoutMs: Number(optional("OPENCODE_TIMEOUT_MS", "300000")),
+      permission: optional("JUNIOR_OPENCODE_PERMISSION", "allow"),
+      mcpEnabled: parseBooleanEnv("OPENCODE_MCP_ENABLED", true),
+      slackMcpEnabled: parseBooleanEnv("OPENCODE_SLACK_MCP_ENABLED", true),
+      playwrightMcpEnabled: parseBooleanEnv(
+        "OPENCODE_PLAYWRIGHT_MCP_ENABLED",
+        true,
+      ),
```
anmolm-growthx, line 142: **🟡 Should-fix — `OPENCODE_PLAYWRIGHT_MCP_ENABLED` defaults to true.**

When this is on, every OpenCode spawn that has worktree-backed MCP wiring includes `{type: "local", command: ["npx", "@playwright/mcp", "--headless"]}` in the generated config. `npx @playwright/mcp` will install Playwright on first run if it isn't cached, and even after warm cache adds non-trivial cold-start cost on every spawn (because it's a fresh `npx` invocation per child). Combined with the per-turn spawn model (rule 2 in CLAUDE.md), this multiplies per-message latency.

The doc itself argues "Junior should be explicit" about MCP/permissions; defaulting Playwright on is the opposite. Suggest defaulting to `false` and letting threads opt in (or making it a `RUNNER_PROVIDER=opencode`-only default once you've measured cost). At minimum, README should call this out.

src/config.ts
```diff
@@ -97,6 +113,20 @@ export function loadConfig(): Config {
       permissionMode: optional("CLAUDE_PERMISSION_MODE", "bypassPermissions"),
       defaultModel: process.env.CLAUDE_MODEL ?? null,
     },
+    runner: {
+      provider: parseRunnerProvider(optional("RUNNER_PROVIDER", "claude")),
+    },
+    opencode: {
+      model: process.env.OPENCODE_MODEL ?? null,
+      timeoutMs: Number(optional("OPENCODE_TIMEOUT_MS", "300000")),
+      permission: optional("JUNIOR_OPENCODE_PERMISSION", "allow"),
```
anmolm-growthx, line 136: **🟢 Nit — env var prefix asymmetry.**

`JUNIOR_OPENCODE_PERMISSION` uses the `JUNIOR_` prefix; every other OpenCode knob uses `OPENCODE_*` (`OPENCODE_MODEL`, `OPENCODE_TIMEOUT_MS`, `OPENCODE_MCP_ENABLED`, ...). Easy footgun for operators sweeping their env. Pick one — `OPENCODE_PERMISSION` matches the rest.

src/lifecycle/dev-server.ts
```diff
@@ -49,6 +49,9 @@ const IDLE_SWEEP_INTERVAL_MS = 60_000;
 /** Grace period for SIGINT before escalating to SIGKILL (ms). */
 const SIGINT_GRACE_MS = 5_000;
 
+/** After a kill, wait briefly for the listener to release its TCP port. */
+const PORT_RELEASE_TIMEOUT_MS = 5_000;
```
anmolm-growthx, line 53: **🟢 Nit — unrelated dev-server changes.**

The new `PORT_RELEASE_TIMEOUT_MS` / `waitForPortRelease` plumbing and the `fix/test`-branch fake-server commit in `dev-server.test.ts` are not part of the runner-provider story. They're reasonable on their own merits but they belong in their own PR — bundling them here makes it harder to revert one without the other and inflates the diff Pranav has to review.

src/runners/index.ts
```diff
@@ -0,0 +1,89 @@
...
+      },
+      targetRepoCwd,
+      botToken,
+      agentIdentity,
+      imagePaths,
+    );
+  }
+
+  if (provider === "codex") {
+    throw new Error("Codex runner provider is not implemented yet");
```
anmolm-growthx, line 58: **🟢 Nit — unreachable codex throw.**

With config-load + `!provider` rejection in place, this `throw` is now unreachable from any normal code path (the only way to land here is direct mutation of a persisted `provider: "codex"` row — see related comment on `sqlite.ts`). Worth a one-line `// defense-in-depth: every user-facing entry rejects codex; if you hit this, persisted state is corrupt` so a future reader doesn't mistake it for partially wired codex spawning.

anmolm-growthx:
**Verdict: ⚠️ Ship after addressing**

Reviewed all three commits walked individually, then reconciled against the final state. The provider boundary is well-shaped — app code (manager / formatter / home / timeout) consumes only `RunnerEvent` / `SpawnHandle` / `SpawnResult`, and the centralized `buildRunnerRuntime` keeps cwd/MCP/env policy from drifting between adapters. Commit 2's review-followup fixed the right things at the right layers (codex at config + `!provider`, agent-session sessionId in the switch guard, parser logging). Nothing here is blocking, but a few gaps in the same family as the issues commit 2 patched are worth a follow-up before merging.

---

### 🔴 Blocking
None.

---

### 🟡 Should-fix
1. **Codex defense-in-depth on persisted load** — `src/session/store/sqlite.ts:232`. `normalizeRunnerProvider("codex")` returns `"codex"` and round-trips it (the test on `sqlite.test.ts:69` asserts this), so a row with `provider: "codex"` (or `agent_sessions.provider = "codex"`) loads cleanly and the next spawn throws inside `runRunnerWithAgent`'s catch. The PR description claims rejection at every entry point; the persisted-load path is the third one. Demote unimplemented providers on load (or run `isImplementedRunnerProvider` symmetrically with config). Inline detail on `sqlite.ts:232`.
2. **`done` gated on `step_finish.part.tokens`** — `src/opencode/parser.ts:189–195`. Claude unconditionally emits `done` on every result event; OpenCode swallows it whenever tokens are absent. Tests only cover the happy path. Fix is one branch — emit `done` for every `step_finish`, with `usage` optional.
3. **`!provider` busy guard misses busy worker agents** — `src/session/manager.ts:421`. `handleAgentMessage` sets only `agentSession.status = "busy"`, never `session.status`, so `!provider` succeeds while a worker is mid-turn. The new `hasAgentSession.sessionId` check catches it after the first `init` lands but a first-turn-in-flight worker (sessionId still null) slips through. Add `agentSessions[*].status === "busy"` to the busy check.
4. **OpenCode `message` events are per-fragment, not per-turn** — `src/opencode/parser.ts:182–188`. Claude aggregates per-turn assistant text into one `message`; OpenCode emits one per text chunk. `responder.updateStatus` therefore fires N times per turn for OpenCode (Slack rate-limit risk), and the user only sees the trailing fragment in the live status line. Coalesce per-turn or document the asymmetry.
5. **`OPENCODE_PLAYWRIGHT_MCP_ENABLED` defaults to `true`** — `src/config.ts:125–128`. Every OpenCode spawn that has worktree-backed MCP wiring then runs `npx @playwright/mcp` cold per turn. Combined with the per-turn spawn model that's a real latency multiplier; doc itself argues "Junior should be explicit" about MCP. Suggest defaulting off, opt-in per thread.

---

### 🟢 Nits
- `src/config.ts:122` — `JUNIOR_OPENCODE_PERMISSION` breaks the `OPENCODE_*` prefix used by every other OpenCode env var.
- `src/lifecycle/dev-server.ts:52` & `dev-server.test.ts` — dev-server port-release plumbing and the fix/test fake-server commit are scope creep; belong in their own PR.
- `src/runners/index.ts:56` — the codex `throw` is now unreachable from any normal entry; one-line comment that it's defense-in-depth would prevent it being mistaken for partial codex wiring.

---

### What I liked
- **Boundary actually holds.** Slack formatting, manager state machine, timeout, and home tab consume `RunnerEvent` / `SpawnHandle` / `SpawnResult` — no Claude-native shape leaks across the seam. That's the hard part of this kind of refactor and it landed cleanly.
- **`buildRunnerRuntime` centralizes cwd/MCP/env policy** so the contract in `runner-providers.md` is enforced in one place. A future codex adapter can't accidentally drift on the `session.cwd` MCP carve-out or the env contract.
- **Parser logs unmapped event types at INFO** with the JSON snippet — that's a great pattern for the deferred tool-event work, you get a fixture-capture trail in production logs without having to manually add tracing later. The `delete env.OPENCODE_CONFIG` block also gets credit for the "why" comment explaining the merge-layer footgun; that comment will pay for itself the next time someone wonders what it's doing.

— Friday's review bot

src/session/store/sqlite.ts
anmolm-growthx: ✅ Fixed in d6a97b0. New `normalizePersistedRunnerProvider` (sqlite.ts:251) wraps `normalizeRunnerProvider` and demotes any non-implemented provider to `claude`, applied symmetrically on read and write. Test added: `demotes persisted unimplemented providers to claude` (sqlite.test.ts:88). Minor follow-up nit: silent demotion — a one-shot `log.warn` would still help operators notice pre-cutover rows getting rewritten, but the safety hole is closed.

src/opencode/parser.ts
anmolm-growthx: ✅ Fixed in d6a97b0. The token gate is gone — `step_finish` now always emits `done`, with `usage` attached only when present (parser.ts:191–201). Test added: `emits done even when step_finish has no usage` (parser.test.ts:103).

src/session/manager.ts
anmolm-growthx: ✅ Fixed in d6a97b0. New `hasBusyAgent = Object.values(session.agentSessions ?? {}).some(a => a.status === "busy")` joined into the `session.status !== "idle"` guard at manager.ts:421. Test added: `blocks provider changes while an agent session is busy before init` (manager.test.ts:632) — seeds a worker with `status="busy"` and `sessionId=null`, asserts `!provider opencode` is rejected and provider stays claude.

src/opencode/parser.ts
anmolm-growthx: ✅ Fixed in d6a97b0. The mapper now buffers in `pendingText` and emits exactly one coalesced `message` per `step_finish` (parser.ts:184–192). Test updated to assert two consecutive text fragments coalesce: `text:"Hi"` + `text:" there"` → single `message {text: "Hi there"}`. Slack now sees one updateStatus per turn, matching the Claude path.

src/config.ts
anmolm-growthx: ✅ Fixed in d6a97b0. Default flipped to `false` (config.ts:127), README + project-setup table updated, and runner-providers.md adds the rationale ("`npx @playwright/mcp` is cold per runner spawn"). Config test asserts the new default. Operators who want it can opt in via `OPENCODE_PLAYWRIGHT_MCP_ENABLED=true`.

src/runners/index.ts
anmolm-growthx: ✅ Fixed in d6a97b0. Defense-in-depth comment added at runners/index.ts:55–57.

anmolm-growthx:
## Re-review of d6a97b0

Walked the new commit against the 5 should-fix items from the prior review. All five close, with regression tests for each. Typecheck clean, 367/367 tests pass (was 364 — three new).

| # | Item | Status | Where |
|---|------|--------|-------|
| 1 | Codex on persisted-load | ✅ Fixed | `sqlite.ts:251` `normalizePersistedRunnerProvider`; test `sqlite.test.ts:88` |
| 2 | `done` gated on tokens | ✅ Fixed | `parser.ts:191–201`; test `parser.test.ts:103` |
| 3 | `!provider` busy-guard | ✅ Fixed | `manager.ts:421` adds `hasBusyAgent`; test `manager.test.ts:632` |
| 4 | Per-fragment OpenCode messages | ✅ Fixed | `parser.ts:184–192` `pendingText` buffer; test asserts `"Hi"` + `" there"` coalesce |
| 5 | Playwright MCP default | ✅ Fixed | `config.ts:127` flipped to `false`; README/project-setup/runner-providers updated; config test asserts new default |

Plus the type-safety bonus: `RunnerProvider` → `ImplementedRunnerProvider` propagated through `Config.runner.provider` and `parseRunnerProvider`'s return type, so the compiler now enforces what the runtime guards already did. That's the right shape.

### One follow-up nit (non-blocking)
`normalizePersistedRunnerProvider` silently demotes — a one-shot `log.warn` when a row gets rewritten would help operators notice when a `codex` row from older state is being rewritten on load. Not a ship blocker.

### Verdict: ✅ Ship.

— Friday's review bot

PR #31, opened 2026-05-15
Improve OpenCode runner prompt quality

PranavBakre:
## Summary
- compose OpenCode agent prompts from a Junior-owned provider baseline plus the dynamic Junior system prompt
- default OpenCode runs to the built-in `build` agent so native build-agent tools/permissions are preserved while the prompt is overridden
- document the prompt strategy and add prompt/spawner coverage

## Verification
- `bun test`
- `bun run typecheck && git diff --check && bun test`
- local `opencode debug agent` check on OpenCode 1.14.48 confirmed custom generated agents resolve `native: false`, while overriding `agent.build.prompt` keeps `native: true`

### DIFF
--- a/docs/features/runner-providers.md
+++ b/docs/features/runner-providers.md
@@ -20,7 +20,7 @@ specific needs because it has:
 - `--session <id>` for resume
 - `--dir <path>` for worktree cwd
 - `--file <path>` for attachments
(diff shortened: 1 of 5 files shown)


PranavBakre:
Superseded by other PRs

PR #38, opened 2026-05-18
Fix OpenCode MCP and support agent access

PranavBakre:
## Summary
- Include generated OpenCode MCP config for normal non-utility runs, including initial lead intake from Junior root.
- Mirror stateless support prompts into generated OpenCode subagents for `nr-research`, `sentry-fetch`, and `vercel-status`.
- Enable Playwright MCP by default for OpenCode parity and update related docs/tests.

## Verification
- `bun run typecheck` (twice)
- `bun test`
- `bun test src/opencode/config.test.ts src/opencode/spawner.test.ts src/config.test.ts` (twice)

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -138,7 +138,7 @@ export function loadConfig(): Config {
       slackMcpEnabled: parseBooleanEnv("OPENCODE_SLACK_MCP_ENABLED", true),
       playwrightMcpEnabled: parseBooleanEnv(
         "OPENCODE_PLAYWRIGHT_MCP_ENABLED",
-        false,
+        true,
       ),
     },
     repos: (JSON.parse(optional("REPOS", "[]")) as RepoConfig[]).map((r) => ({
--- a/src/opencode/config.ts
+++ b/src/opencode/config.ts
@@ -1,5 +1,6 @@
 export type OpenCodePermissionConfig = string | Record<string, unknown>;
 type OpenCodeAgentPermissionConfig = Record<string, unknown>;
+type OpenCodeAgentMode = "primary" | "subagent";
 
 export interface OpenCodeMcpEntry {
   type?: string;
@@ -18,28 +19,36 @@ export interface BuildOpenCodeConfigOptions {
   permission?: OpenCodePermissionConfig;
   description?: string;
   mcp?: OpenCodeMcpConfig | null;
+  subagents?: OpenCodeSubagentConfig[];
+}
+
+export interface OpenCodeSubagentConfig {
+  name: string;
+  prompt: string;
+  description?: string;
+  permission?: OpenCodePermissionConfig;
+}
+
+export interface OpenCodeGeneratedAgentConfig {
+  description: string;
+  mode: OpenCodeAgentMode;
+  permission: OpenCodeAgentPermissionConfig;
+  prompt: string;
 }
 
 export interface OpenCodeGeneratedConfig {
   $schema: string;
   model?: string;
   permission: OpenCodePermissionConfig;
-  agent: Record<
-    string,
-    {
-      description: string;
-      mode: "primary";
-      permission: OpenCodeAgentPermissionConfig;
-      prompt: string;
-    }
-  >;
+  agent: Record<string, OpenCodeGeneratedAgentConfig>;
   mcp?: OpenCodeMcpConfig;
 }
 
 export function buildOpenCodeConfig(
   options: BuildOpenCodeConfigOptions,
 ): OpenCodeGeneratedConfig {
-  if (!options.agentName.trim()) {
+  const primaryAgentName = options.agentName.trim();
+  if (!primaryAgentName) {
     throw new Error("OpenCode agentName is required");
   }
 
@@ -47,7 +56,7 @@ export function buildOpenCodeConfig(
     $schema: "https://opencode.ai/config.json",
     permission: options.permission ?? "allow",
     agent: {
-      [options.agentName]: {
+      [primaryAgentName]: {
         description: options.description ?? "Junior Slack runner",
         mode: "primary",
         permission: toAgentPermissionConfig(options.permission ?? "allow"),
@@ -55,6 +64,28 @@ export function buildOpenCodeConfig(
       },
     },
   };
+
+  for (const subagent of options.subagents ?? []) {
+    const name = subagent.name.trim();
+    if (!name) {
+      throw new Error("OpenCode subagent name is required");
+    }
+    if (name === primaryAgentName) {
+      throw new Error(`OpenCode subagent "${name}" conflicts with primary agent`);
+    }
+    if (config.agent[name]) {
+      throw new Error(`Duplicate OpenCode subagent "${name}"`);
+    }
+
+    config.agent[name] = {
+      description: subagent.description ?? `Junior support subagent: ${name}`,
+      mode: "subagent",
+      permission: toAgentPermissionConfig(
+        subagent.permission ?? options.permission ?? "allow",
+      ),
+      prompt: subagent.prompt,
+    };
+  }
 
   if (options.model) {
     config.model = options.model;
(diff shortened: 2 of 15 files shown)


src/config.ts
```diff
@@ -138,7 +138,7 @@ export function loadConfig(): Config {
       slackMcpEnabled: parseBooleanEnv("OPENCODE_SLACK_MCP_ENABLED", true),
       playwrightMcpEnabled: parseBooleanEnv(
         "OPENCODE_PLAYWRIGHT_MCP_ENABLED",
-        false,
+        true,
```
anmolm-growthx, line 141: `bin/opencode-with-mcp.sh` still documents `default: false` and `parse_bool ... false` here, but `src/config.ts:138` now defaults `OPENCODE_PLAYWRIGHT_MCP_ENABLED` to `true`. The script's own header (line 9) explicitly promises consistency with the spawned runtime — that promise is broken by this PR. Update the header comment and flip the parse_bool default to `true`, or operators who follow the script's documented behavior will get a different MCP set than Junior's spawn.

src/opencode/config.ts
```diff
@@ -56,6 +65,28 @@ export function buildOpenCodeConfig(
...
+      throw new Error(`Duplicate OpenCode subagent "${name}"`);
+    }
+
+    config.agent[name] = {
+      description: subagent.description ?? `Junior support subagent: ${name}`,
+      mode: "subagent",
+      permission: toAgentPermissionConfig(
+        subagent.permission ?? options.permission ?? "allow",
+      ),
+      prompt: subagent.prompt,
```
anmolm-growthx, line 86: `subagent.permission ?? options.permission ?? "allow"` means `nr-research`/`sentry-fetch`/`vercel-status` get the primary's full surface — `*: allow` (bash/edit/write) by default. These are documented as "stateless support fetchers" but they can shell out and mutate the worktree. Consider passing an explicit constrained permission set (read + relevant MCP) from `loadOpenCodeSupportSubagents()`, or have `buildOpenCodeConfig` apply a stricter default when a subagent omits `permission`. Lower blast radius for a prompt-injection on any of these fetchers.

src/opencode/support-agents.ts
```diff
@@ -0,0 +1,38 @@
...
+  const agents: OpenCodeSubagentConfig[] = [];
+
+  for (const name of STATELESS_SUPPORT_AGENTS) {
+    const promptPath = new URL(
+      `../../support/agents/${name}/prompt.md`,
+      import.meta.url,
+    );
+
+    if (!existsSync(promptPath)) continue;
+
```
anmolm-growthx: Silent skip via `existsSync` means a renamed or unbundled prompt file makes the support subagent vanish without an error — and the lead state machine relies on these three names. Either drop `existsSync` and let `readFileSync` throw, or `log.warn` the miss so a missing prompt is visible at startup.

src/opencode/spawner.ts
```diff
@@ -67,7 +70,13 @@ export function spawnOpenCode(
...
     model,
     permission: config.permission,
-    mcp: runtime.needsProjectMcp ? config.mcp : null,
+    // OpenCode receives MCP via generated config, not Claude's project
+    // .mcp.json. Keep the utility cwd carve-out (calendar/cloud utilities rely
+    // on their own integrations), but all normal Junior/lead/worker runs need
+    // Slack MCP even when cwd is Junior's project root so intake can call
+    // register_worktree before any worktree exists.
+    mcp: session.cwd ? null : config.mcp,
+    subagents: config.subagents ?? loadOpenCodeSupportSubagents(),
```
anmolm-growthx: `mcp` honors the `session.cwd` carve-out, but `subagents` does not — utility-cwd runs still get `nr-research`/`sentry-fetch`/`vercel-status` injected into their agent map. The docs argue utility runs should rely on their own integrations, so the same gating probably should apply: `subagents: session.cwd ? [] : (config.subagents ?? loadOpenCodeSupportSubagents())`. Minor — they're inert unless Task fans out — but the current asymmetry is confusing to read.

anmolm-growthx:
## Verdict: NON-BLOCKING

The three claims hold up against the diff:

- **MCP for non-utility root runs:** `src/opencode/spawner.ts:78` swaps the old `runtime.needsProjectMcp` (which excluded Junior root because `cwd === process.cwd()`) for `session.cwd ? null : config.mcp`. Root lead intake now gets `slack-bot` MCP and can call `register_worktree` before any worktree exists. ✓
- **Support subagents:** `src/opencode/support-agents.ts` loads the three stateless prompts from `support/agents/<name>/prompt.md` into generated `agent.<name>` entries with `mode: "subagent"`. `reproducer`/`thinker`/`review` are intentionally excluded. ✓
- **Playwright default-on:** `src/config.ts:138`, README, and `docs/features/project-setup.md` are aligned. ✓

## Findings

1. **`bin/opencode-with-mcp.sh` drifted from runtime** (real bug). Header comment says `OPENCODE_PLAYWRIGHT_MCP_ENABLED=true|false    default: false` and line 52 hardcodes `false` as the parse_bool default. The script's own preamble promises "behavior is consistent between manual and spawned runs" — that promise is now broken. Worth fixing in this PR or a fast follow.
2. **Permissions for support subagents inherit primary's surface.** `src/opencode/config.ts` sets `permission: toAgentPermissionConfig(subagent.permission ?? options.permission ?? "allow")` — `nr-research`/`sentry-fetch`/`vercel-status` get `*: allow` (bash/edit/etc.) by default. These are "stateless observability fetchers"; consider locking them to `read` + relevant MCP only.
3. **Subagents loaded even for utility-cwd runs.** `spawner.ts:79` calls `loadOpenCodeSupportSubagents()` unconditionally; only `mcp` honors the `session.cwd` carve-out. Probably harmless config bloat for cloud/calendar utility runs, but inconsistent with the reasoning that utility runs intentionally avoid Junior's local wiring.
4. **`existsSync` silent skip.** If any of the three hardcoded prompt paths goes missing (rename, bundling layout change), the subagent is silently omitted. Either let `readFileSync` throw or `log.warn` the miss — the three names are part of the lead state machine, silent degradation is hard to debug.
5. **MCP predicate now diverges from `needsProjectMcp`.** Claude still routes through `runtime.needsProjectMcp` (`runtime.ts:36`), OpenCode uses its own `session.cwd` check. They describe different things now (Claude needs `--mcp-config` only off-root because `.mcp.json` autoloads; OpenCode generates its config every spawn). A comment in `runtime.ts` calling out that `needsProjectMcp` is Claude-only would prevent the next refactor from accidentally unifying them.
6. **"Mirror" is the wrong verb.** There are no Claude-side `.claude/agents/<nr-research|sentry-fetch|vercel-status>.md` files — the prompts live only at `support/agents/<name>/prompt.md`. The PR description and `support-agents.ts` docstring say "mirror"; "expose existing standalone prompts as OpenCode subagents" is more accurate. Pure doc nit.

## Test results

- `bun run typecheck` → clean.
- `bun test src/opencode/config.test.ts src/opencode/spawner.test.ts src/config.test.ts` → **17/17 pass**.
- `bun test` → **514/514 pass**, no warnings worth flagging.

New tests cover: root MCP inclusion, utility-cwd carve-out, support-subagent exposure, subagent collision/duplicate rejection, primary-only with subagents disabled-via-empty.

## Per-commit notes

Single commit (`b024d12`); 12 files, +289/−24. Walking it didn't surface deferred fixes or skipped paths. Default flip for Playwright is the only operational change for existing deployments — call it out in any release notes so operators without `npx`/network connectivity know to set `OPENCODE_PLAYWRIGHT_MCP_ENABLED=false`.

Backwards compatibility: no schema bumps; no field renames. Persisted sessions from PR #30 will pick up the new defaults transparently.

pranav-growthx (commented):
Second pass done. I don’t have new line-specific findings beyond the existing inline comments. The main risks I’d fix before relying on this path are still: helper-script default drift for Playwright, broad default permissions for stateless support subagents, silent prompt-file skips, and the utility-cwd asymmetry for injected subagents.

Verification:
- `bun run typecheck` passed.
- `bun test src/opencode/config.test.ts src/opencode/spawner.test.ts src/config.test.ts` passed: 17/17.
- `bun test` currently fails 1 unrelated-looking runtime test in this shell: `runner runtime > omits JUNIOR_SLACK_ICON_EMOJI when identity has no iconEmoji` receives `:face_with_cowboy_hat:` from the inherited environment. The PR-targeted OpenCode tests are clean.

PR #39, opened 2026-05-18
Fix dormant Slack thread catch-up

PranavBakre:
## Summary
- Add dormant Slack thread catch-up on !listen/@mention for resumed sessions.
- Filter !aside messages from injected thread history, including punctuated !aside variants.
- Add tests covering catch-up prompts, session defaults, and aside filtering.
- Add Codex runner scope/CLI notes under docs/features.
- Update agents-org to be04031, adding the admin-account worker and tightening credential delivery guidance.

## Verification
- [x] bun run typecheck
- [x] bun test
- [x] bun run typecheck (second pass)
- [x] bun test (second pass)

### DIFF
--- a/src/session/manager.ts
+++ b/src/session/manager.ts
@@ -39,7 +39,6 @@ import { isDuplicateSlackToolResponse } from "../slack/formatting.ts";
 import { DEFAULT_CONTEXT_PROFILE, type AgentContextProfile } from "../agents/loader.ts";
 import { downloadSlackFiles } from "../slack/files.ts";
 import { log as _log } from "../logger.ts";
-import { isAsideText } from "../slack/commands.ts";
 
 export class SessionManager {
   private store: SessionStore;
@@ -122,7 +121,7 @@ export class SessionManager {
     // alternative (skip tracking) creates a corner where U-B posts only
     // asides, then a real non-mention message, and the gate fails to fire
     // because U-B was never registered as present.
-    if (event.command === "aside" || isAsideText(event.text)) {
+    if (event.command === "aside") {
       const isHuman = !event.isSelfBot && !event.botId;
       if (isHuman && event.user) {
         const session = await this.store.get(event.threadId);
--- a/src/session/manager.ts
+++ b/src/session/manager.ts
@@ -39,6 +39,7 @@ import { isDuplicateSlackToolResponse } from "../slack/formatting.ts";
 import { DEFAULT_CONTEXT_PROFILE, type AgentContextProfile } from "../agents/loader.ts";
 import { downloadSlackFiles } from "../slack/files.ts";
 import { log as _log } from "../logger.ts";
+import { isAsideText } from "../slack/commands.ts";
 
 export class SessionManager {
   private store: SessionStore;
@@ -121,7 +122,7 @@ export class SessionManager {
     // alternative (skip tracking) creates a corner where U-B posts only
     // asides, then a real non-mention message, and the gate fails to fire
     // because U-B was never registered as present.
-    if (event.command === "aside") {
+    if (event.command === "aside" || isAsideText(event.text)) {
       const isHuman = !event.isSelfBot && !event.botId;
       if (isHuman && event.user) {
         const session = await this.store.get(event.threadId);
@@ -144,6 +145,7 @@ export class SessionManager {
     if (event.command === "listen") {
       if (session && session.dormant) {
         session.dormant = false;
+        session.needsThreadCatchup = true;
         await this.store.set(event.threadId, session);
       }
       this.onReaction?.(event, "ear");
@@ -154,6 +156,7 @@ export class SessionManager {
     // processed normally.
     if (event.mentionsJunior && session?.dormant) {
       session.dormant = false;
+      session.needsThreadCatchup = true;
       await this.store.set(event.threadId, session);
     }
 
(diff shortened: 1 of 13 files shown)


src/session/manager.ts
```diff
@@ -121,7 +122,7 @@ export class SessionManager {
     // alternative (skip tracking) creates a corner where U-B posts only
     // asides, then a real non-mention message, and the gate fails to fire
     // because U-B was never registered as present.
-    if (event.command === "aside") {
+    if (event.command === "aside" || isAsideText(event.text)) {
```
anmolm-growthx: Minor: the second clause is redundant in practice. By the time `gateAttention` runs, `event.text` is `parsed.text` from `events.ts:156/203` — the `!aside` prefix has already been stripped — and `parseCommand` now maps `!aside.` / `!aside,` / etc. to `command: "aside"` via the new branch in `commands.ts:55`. So `isAsideText(event.text)` here only fires for the contrived case of a body that itself begins with `!aside` (e.g. user typed `!aside !aside foo`). Defensible as defense-in-depth, but worth a one-line comment so a future reader doesn't go hunting for the path that exercises it.

— Friday 🔔

src/session/manager.ts
```diff
@@ -889,8 +892,21 @@ export class SessionManager {
...
+          const preambleProfile: AgentContextProfile = needsThreadCatchup
+            ? {
+                ...contextProfile,
+                identity: false,
+                slack: true,
+                threadHistory: true,
+                threadHistoryLimit: Math.max(
+                  contextProfile.threadHistoryLimit,
+                  1000,
+                ),
```
anmolm-growthx, line 906: Catch-up payload sizing. With `Math.max(contextProfile.threadHistoryLimit, 1000)` (manager.ts:904), a dormant thread that ran long enough can fan out into a `conversations.replies` page of up to 1000 messages — and per the existing TODO at `thread-context.ts:263` each one then triggers an unbatched `users.info` lookup inside the `Promise.all`. On a chatty channel that's a real Slack-rate-limit cliff, and it's exactly the catch-up case where lots of activity is likely. Not blocking for this PR — but worth either (a) reusing the agent's declared limit and trusting the agent author to size up if needed, or (b) keeping 1000 as a ceiling but landing the batch-resolve TODO before this gets stressed.

— Friday 🔔

anmolm-growthx:
## Verdict
**NON-BLOCKING** — clean fix with a couple of follow-up notes.

## What this PR does
- Adds a `needsThreadCatchup` session flag set when `!listen` / @mention wakes a dormant thread; the next resumed runner turn re-injects thread history (with `threadHistoryLimit` floored at 1000) so the agent sees what arrived while it was staying out.
- Extends `!aside` recognition to punctuated variants (`!aside.`, `!aside,`, `!aside:`, `!aside!`, `!aside?`, `!aside-`) via a shared `isAsideText` helper, used both in `parseCommand` and to filter the injected thread history.
- Bundles aspirational Codex-runner design docs (~800 lines) and a submodule bump for `agents-org`.

## Strengths
- `isAsideText` is one regex shared across `commands.ts`, `session/manager.ts`, and `slack/thread-context.ts` — no drift risk between "is it an aside?" and "should the catch-up suppress it?". The regex is correctly anchored (`^!aside(?:$|\s|[.,:;!?-])`), so `!asider` doesn't false-positive.
- The new flag is defaulted in `createSession`, normalized in the session-load path (`manager.ts:1242`), and added to the test-helper sessions in `args.test.ts`, `tmux-driver.test.ts`, `runners/runtime.test.ts` — no stray `undefined` paths.
- Tests are integration-shaped where it counts: `manager.test.ts` exercises the full `!listen` → wake → resumed turn → assert prompt contains real msgs and excludes the aside; `thread-context.test.ts` covers the punctuated `!aside.` filter at the preamble layer. Cleared-flag assertion is included.
- `Math.max(contextProfile.threadHistoryLimit, 1000)` is the right shape — never *shrinks* an agent that asked for more than the catch-up default.

## Concerns
1. **Redundant clause** in `manager.ts:125` (`event.command === "aside" || isAsideText(event.text)`). By the time we're inside `gateAttention`, `event.text` is `parsed.text` and `parseCommand` already maps punctuated `!aside` variants to `command: "aside"`. The OR is harmless defense-in-depth but suggests a misread of the event shape — see inline.
2. **Catch-up payload size**. 1000-message floor × per-message unbatched `users.info` (existing TODO at `thread-context.ts:263`) is a Slack rate-limit cliff exactly in the case where catch-up matters most. Not blocking; see inline.
3. **Flag-clear is guarded by `slackApp && latestTs` and runs only after `buildPromptPreamble` resolves.** If preamble throws, the flag stays set and next turn re-fetches — benign (catch-up just re-runs), worth knowing.
4. **Codex-runner docs are aspirational** — `src/runners/index.ts:55` and `src/config.test.ts:100` still reject `RUNNER_PROVIDER=codex` with "not implemented". Fine to land design docs ahead of code; just flagging that future reviewers shouldn't treat these docs as load-bearing for this PR's behavior.
5. **`agents-org` submodule bump** is opaque from outside the private repo — trust but verify the diff doesn't shift credential handling.

## Nits
- The replacement regex `^!aside(?:[.,:;!?-])?\s*/i` inside `parseCommand` accepts trailing-punctuation variants the detector approves of, but it's a second regex of truth in the same function. Could factor to a single source by capturing-and-stripping in `isAsideText` itself.
- No negative test for `!asider` / `!asides` (regex is correct, but a single-line `expect(parseCommand("!asider thing").command).toBe(null)` would lock the boundary).

## Tests
Coverage is solid for the happy paths and the asymmetric "include real / exclude aside" assertion. Missing edge cases worth adding if you touch this again: (a) `!asider` negative, (b) `!ASIDE` casing, (c) thread of only-asides (`fetchThreadHistory` should return `null`), (d) concurrent second message arriving during catch-up — current behavior is fine since the second resumed turn already has --resume context from the first, but a test would pin it.

— Friday 🔔

PR #40, opened 2026-05-19
Scope Mixpanel MCP to feature metrics

PranavBakre:
## Summary
- Inject Mixpanel's hosted MCP only for `feature-metrics` OpenCode sessions behind `OPENCODE_MIXPANEL_MCP_ENABLED`.
- Add runner config tests and document the scoped MCP behavior.
- Add a manual OpenCode agent launcher that composes Junior agent prompts with MCP config.

## Verification
- `bun run typecheck`
- `bun test src/runners/index.test.ts src/config.test.ts src/session/manager.test.ts`

### DIFF
--- a/bin/opencode-agent-with-mcp.sh
+++ b/bin/opencode-agent-with-mcp.sh
@@ -1,0 +1,256 @@
+#!/bin/bash
+# Launch an OpenCode session with Junior agent prompts and MCP servers configured.
+#
+# This is the manual-dev companion to Junior's generated OpenCode runtime config:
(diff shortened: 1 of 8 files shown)


PranavBakre (commented):
No blocking findings. The Mixpanel MCP gate is scoped to feature-metrics for runtime sessions and the manual launcher matches the same flag behavior. Verified locally: `bun run typecheck` and `bun test src/runners/index.test.ts src/config.test.ts src/session/manager.test.ts`.

PR #41, opened 2026-05-19
fix: sanitize Slack error responses

PranavBakre:
## Summary
- Sanitize runner error messages before posting them to Slack.
- Withhold prompt/context-like stderr and overlong errors while keeping raw details in server logs.
- Add regression coverage for the leaked prompt failure mode.

## Verification
- bun test src/slack/formatting.test.ts
- bun run typecheck
- bun test

### DIFF
--- a/src/slack/formatting.ts
+++ b/src/slack/formatting.ts
@@ -54,6 +54,46 @@ export function extractRunnerMessageText(event: RunnerEvent): string | null {
 
 export const NO_SLACK_MESSAGE = "NO_SLACK_MESSAGE";
 
+const MAX_SLACK_ERROR_LENGTH = 500;
+const PROMPT_LEAK_MARKERS = [
+  /<\/?(?:identity|slack-context|thread-context|system|developer|user|assistant)\b/i,
+  /#\s*(?:IDENTITY|SOUL)\.md\b/i,
+  /Do NOT use Slack search/i,
+  /CRITICAL\s+[—-]\s+no double-posting/i,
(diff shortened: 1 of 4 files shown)


src/slack/formatting.ts
```diff
@@ -54,6 +54,46 @@ export function extractRunnerMessageText(event: RunnerEvent): string | null {
 
 export const NO_SLACK_MESSAGE = "NO_SLACK_MESSAGE";
 
+const MAX_SLACK_ERROR_LENGTH = 500;
+const PROMPT_LEAK_MARKERS = [
+  /<\/?(?:identity|slack-context|thread-context|system|developer|user|assistant)\b/i,
```
PranavBakre: This marker list is too narrow for a sanitizer whose job is to suppress prompt/context-shaped stderr. It catches `<identity>` / `<slack-context>` and long leaks, but short errors containing other injected blocks like `<workspace>`, `<persistent-agent-state>`, `<dispatch-allow>`, `<junior-core>`, or `<opencode-provider-baseline>` would still be posted if they are under 500 chars and do not come from the exact `File not found: <` path. Since this is a leak guard, I’d make this deny any XML-ish prompt tag we inject (or maintain a complete shared list of prompt wrapper tags) rather than only a few current examples.

PranavBakre (commented):
Found one security-hardening gap: the sanitizer still lets some short prompt/context-shaped errors through. Targeted tests and typecheck pass; full bun test is blocked in this checkout by missing proper-lockfile.

PR #43, opened 2026-05-19
fix: keep OpenCode prompt before file args

PranavBakre:
## Summary
- Move OpenCode file attachment flags after the positional prompt so `--file` does not consume injected prompt text as a filename.
- Add a regression test for the prompt-as-file parsing failure and update the runner provider code index.

## Verification
- bun test src/opencode/args.test.ts
- bun run typecheck
- bun test


PranavBakre (commented):
No blocking findings. I can't approve because GitHub marks this as your own PR from this auth context, but review is clean.\n\nThe arg order change matches OpenCode's `run [message..]` parser shape and the regression coverage catches the prompt-before-file invariant.\n\nVerified locally:\n- `bun test src/opencode/args.test.ts`\n- `bun run typecheck`\n- `bun test`

PR #44, opened 2026-05-19
docs: align documentation with multi-provider architecture and add tool prerequisites

PranavBakre:
This PR synchronizes the project documentation with the current implementation state. It acknowledges OpenCode as the default runner, preserves CLAUDE.md for manual use, documents external tool requirements (Vercel, Sentry, NewRelic), and fixes technical drift in package.json.

### DIFF
--- a/docs/code_index/agent-definitions.md
+++ b/docs/code_index/agent-definitions.md
@@ -1,0 +1,21 @@
+# Code Index: Agent Definitions
+
+This module handles loading and parsing of agent definitions from markdown files with YAML frontmatter.
+
+## Directory Structure
+
+- [.claude/agents/](file:///Users/psbakre/Projects/junior/.claude/agents/): Public agent definitions.
+- [agents-org/](file:///Users/psbakre/Projects/junior/agents-org/): Private/overlay agent definitions (git submodule).
+- [support/agents/](file:///Users/psbakre/Projects/junior/support/agents/): Stateless support sub-agent prompts.
+
+## Key Files
+
+- [loader.ts](file:///Users/psbakre/Projects/junior/src/agents/loader.ts): Handles reading `.md` files and parsing frontmatter identities.
+- [router.ts](file:///Users/psbakre/Projects/junior/src/agents/router.ts): High-level routing logic to pick an agent definition based on thread state.
+
+## Data Flow
+
+1. `AgentDispatcher` (in `support/router.ts`) identifies the requested agent.
+2. `AgentRouter` calls the `loader` to fetch the definition from the target repo's `.claude/agents/` or Junior's own registries.
+3. The frontmatter (username, emoji) is used for Slack attribution.
+4. The markdown body is passed as the system prompt to the runner provider.
--- a/docs/code_index/agent-definitions.md
+++ b/docs/code_index/agent-definitions.md
@@ -1,21 +1,31 @@
 # Code Index: Agent Definitions
 
-This module handles loading and parsing of agent definitions from markdown files with YAML frontmatter.
+This module handles loading, parsing, and routing of agent definitions from markdown files with YAML frontmatter.
 
-## Directory Structure
+## Definition Locations
 
-- [.claude/agents/](file:///Users/psbakre/Projects/junior/.claude/agents/): Public agent definitions.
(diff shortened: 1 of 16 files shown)


docs/code_index/agent-definitions.md
```diff
@@ -0,0 +1,21 @@
+# Code Index: Agent Definitions
+
+This module handles loading and parsing of agent definitions from markdown files with YAML frontmatter.
+
+## Directory Structure
+
+- [.claude/agents/](file:///Users/psbakre/Projects/junior/.claude/agents/): Public agent definitions.
```
PranavBakre: These `file:///Users/psbakre/...` links make the docs non-portable and bake one local checkout path into the repo. The existing code indexes use repo-relative file paths, so these new indexes should do the same. Same issue appears in the other new code-index files too.

docs/security.md
```diff
@@ -0,0 +1,25 @@
+# Security Posture
+
+Junior is designed as a developer tool running on a local workstation or a trusted private server.
+
+## 1. Loopback-only Dashboard
+The [HTTP Dashboard](./features/http-dashboard.md) binds only to `127.0.0.1` by default.
+- **No Remote Access**: It is not reachable from the internet without an explicit SSH tunnel.
+- **No Auth**: Because it is loopback-only, it does not include an authentication layer.
+- **CSRF Protection**: The API is restricted to the same origin; external sites cannot trigger Junior actions via the dashboard.
```
PranavBakre: This overstates what the code enforces. `startHttpServer` binds to `127.0.0.1` and intentionally omits CORS headers, but it does not validate `Origin`/`Sec-Fetch-Site` or otherwise restrict requests to same-origin callers. For the current read-only dashboard, a safer claim is that arbitrary websites cannot read API responses via CORS, not that same-origin is enforced or that this is CSRF protection.

PranavBakre (commented):
Found a couple of doc accuracy/portability issues. Typecheck and build both pass locally.

package.json
```diff
@@ -6,7 +6,9 @@
     "dev": "bun --watch src/index.ts",
     "start": "bun src/index.ts",
     "typecheck": "tsc --noEmit",
-    "test": "bun test"
+    "test": "bun test",
+    "build": "bun build ./src/index.ts --outdir ./dist --target bun",
+    "cleanup": "bun run src/lifecycle/cleanup.ts"
```
PranavBakre, line 11: This script is a no-op right now. `src/lifecycle/cleanup.ts` only exports `cleanupStaleSessions`; it has no CLI entrypoint, config/store creation, or invocation when executed directly. I verified `bun run cleanup` only prints `$ bun run src/lifecycle/cleanup.ts` and exits without deleting anything. Either add a real cleanup CLI wrapper or don't document/add this script yet.

docs/security.md
```diff
@@ -0,0 +1,27 @@
...
+- **Data Plane Isolation**: Junior creates isolated git worktrees for each Slack thread.
+- **No Shared State**: Edits in one thread cannot affect the main branch or other threads until explicitly pushed by the runner.
+
+## 3. Provider-level Permissions
+Junior uses the native permission models of its runners:
+- **OpenCode**: Junior generates config per spawn through `OPENCODE_CONFIG_CONTENT`, including permissions, MCP entries, and constrained support subagents.
+- **Claude Code**: Uses `--permission-mode bypassPermissions` for trusted developer use, or can be configured for restricted modes.
+
+## 4. MCP Server Security
+The in-process Slack MCP server only exposes tools to the locally spawned runner processes. It requires a shared `MCP_PORT` that is not exposed externally.
```
PranavBakre: This still overstates MCP exposure. `startMcpServer` calls `httpServer.listen(MCP_PORT)` with no hostname, which binds to the unspecified address on Node/Bun rather than explicitly to `127.0.0.1`. Unless the code is changed to bind loopback, the doc should not say the MCP server is not exposed externally.

PranavBakre (commented):
Re-reviewed. The prior portability + dashboard CSRF wording issues are fixed. I found two remaining accuracy issues. `bun run typecheck`, `bun run build`, and `bun test` pass; `bun run cleanup` currently executes but does no cleanup.

PR #45, opened 2026-05-21
feat: add dynamic workflow runtime

PranavBakre:
## Summary
- add dynamic workflow definitions loaded from fixed workflows/ and agents-org/workflows roots with fs.watch hot reload
- add workflow runtime state/run storage, scheduler, Slack controller commands, and worklog collection
- document schema, runtime behavior, submodule contract, and v2 hot reload backlog

## Verification
- bun run typecheck
- bun test
- bun run build


src/workflows/registry.ts
```diff
@@ -0,0 +1,202 @@
...
+      next.set(definition.name, definition);
+    }
+    for (const definition of loadedByRoot.get("overlay")?.values() ?? []) {
+      next.set(definition.name, definition);
+    }
+
+    // Last-known-good: invalid changed files do not evict previous active
+    // definitions. If a valid file is deleted, however, it is intentionally
+    // removed because the operator deleted the source.
+    for (const error of errors) {
```
PranavBakre, line 115: This only preserves an invalid overlay after that overlay was previously loaded successfully. On a cold boot with `agents-org/workflows/worklog.workflow.md` present but invalid, `previousDefinition` is empty, so `next` keeps the public workflow loaded at lines 104-107. That violates the documented fail-closed behavior (“does not silently fall back to the public workflow while an overlay with the same name is invalid”) and can run/post the public workflow instead of the intended private override after restart. Use the overlay filename/name as a tombstone even when validation fails, or otherwise block the public definition whenever an invalid overlay file for the same stem exists.

PranavBakre (commented):
Reviewed. One blocker around overlay fail-closed semantics.

src/workflows/scheduler.ts
```diff
@@ -0,0 +1,257 @@
...
+    }
+
+    const next = nextScheduledRun(definition, this.now());
+    if (!next) {
+      await this.store.setState({ ...state, nextRunAt: null });
+      return;
+    }
+
+    const delay = Math.max(0, next.getTime() - this.now().getTime());
+    const timer = setTimeout(() => {
```
anmolm-growthx, line 141: 🔴 setTimeout delay is not clamped to the 32-bit timer limit (2147483647 ms ≈ 24.85 days). `delay` here is `next.getTime() - now` with no upper bound, so any schedule whose next fire is more than ~24.8 days out (a monthly `0 0 1 * *` or yearly cron) overflows: Node/Bun clamp an over-range delay to 1 ms, the timer fires almost immediately, `handleScheduledRun` runs the workflow far too early, then reschedules to the same far-future target and fires immediately again — a tight early-run loop. WHY: monthly/quarterly/yearly workflows silently misfire, repeatedly. HOW: clamp delay to `Math.min(delay, 2_147_483_647)` and, when the real target is further out, re-arm a continuation timer instead of executing.

src/workflows/executor.ts
```diff
@@ -0,0 +1,309 @@
...
+    const provider: ImplementedRunnerProvider =
+      runner.provider === "default" ? this.config.runner.provider : runner.provider;
+    const session = createSession(
+      `workflow-${definition.name}`,
+      "workflow",
+      "quiet",
+      provider,
+      this.config.claude.defaultDriver,
+    );
+    session.cwd = process.cwd();
```
anmolm-growthx: 🔴 Security: `permissions.tools` is validated at load (definition.ts:230-243) but never enforced. This run path consumes `permissions.repos` (reposFor) yet nothing reads `permissions.tools`, and the runner is spawned with `session.cwd = process.cwd()` (the Junior repo root) plus an externally-authored `agentName`/`model`/`prompt`. Because overlay defs load from the `agents-org` git submodule (registry.ts roots), a crafted overlay `*.workflow.md` can run an arbitrary agent in the repo root with the agent default full tool access — the declared tool allowlist is cosmetic. WHY: the schema advertises `tools` as a security boundary that does not exist. HOW: translate `permissions.tools` into real runner restrictions (allowed/disallowed tool flags) before spawn, or remove it from the schema and document that workflow defs are fully trusted.

src/workflows/executor.ts
```diff
@@ -0,0 +1,309 @@
...
+  }
+
+  private async emitOutputs(
+    definition: WorkflowDefinition,
+    summary: string,
+  ): Promise<{ channel: string | null; threadTs: string | null }> {
+    let slackChannel: string | null = null;
+    let slackThreadTs: string | null = null;
+    for (const output of definition.outputs) {
+      if (output.type === "docs") continue;
```
anmolm-growthx, line 210: 🔴 The `docs` output `path` is dead. `emitOutputs` skips docs outputs with this `continue`, and the artifact is always written to `artifactPathFor(name, started, runId)` (executor.ts:96-102, 273-276) — the `path` declared in the workflow file is never used. So `outputs: [{type: docs, path: ...}]` has no effect on where artifacts land, and the load-time validation of that path (definition.ts:189-194) guards a value nothing reads. HOW: either honor the declared `path` when writing the artifact, or drop `path` from the docs output schema so it does not imply configurability.

src/workflows/definition.ts
```diff
@@ -0,0 +1,333 @@
...
+
+function parseOutputs(raw: unknown[], workflowName: string): WorkflowOutput[] {
+  if (raw.length === 0) throw new Error("outputs must contain at least one entry");
+  return raw.map((entry, index) => {
+    const output = objectRecord(entry, `outputs[${index}]`);
+    const type = stringField(output, "type");
+    if (type === "docs") {
+      const path = stringField(output, "path");
+      const prefix = `${WORKFLOW_ARTIFACT_ROOT}/${workflowName}`;
+      if (path !== prefix && !path.startsWith(`${prefix}/`)) {
```
anmolm-growthx: 🟡 This containment check permits `..` traversal. `path.startsWith(prefix + "/")` accepts e.g. `data/workflow-runs/worklog/../../../etc/foo`, which passes the test but resolves outside the artifact root. It is latent today because the docs `path` is never used for writing (executor.ts:210), but if that field is ever wired up this becomes a path-traversal write primitive. HOW: normalize with `path.resolve` and assert the resolved path is still inside the resolved artifact root, rather than a string-prefix test.

src/workflows/scheduler.ts
```diff
@@ -0,0 +1,257 @@
...
+      throw new Error(`Workflow ${definition.name} is disabled in its workflow file.`);
+    }
+    const state = await this.ensureState(definition);
+    if (state.status === "invalid") {
+      throw new Error(`Workflow ${definition.name} is invalid and cannot run.`);
+    }
+    if (state.status === "stopped" && options.reason !== "manual") {
+      throw new Error(`Workflow ${definition.name} is stopped. Use !workflow start ${definition.name} to resume triggers.`);
+    }
+    if (definition.concurrency === "skip" && this.running.has(definition.name)) {
```
anmolm-growthx: 🟡 TOCTOU on the `running` set for `concurrency: skip`. `ensureState` above (line 67) awaits, so two near-simultaneous `runNow` calls can both reach this check with `running` still empty, both pass, and both call `execute` — double-running a workflow that declared skip. The scheduled path (handleScheduledRun:149-154) has the same gap versus a concurrent manual run. HOW: add the name to `running` synchronously before any await in the admission path, or gate on a stored in-flight promise per workflow name.

src/workflows/controller.ts
```diff
@@ -0,0 +1,274 @@
...
+  return match?.[1] ?? null;
+}
+
+function slackEventMatches(
+  trigger: WorkflowSlackEventTrigger,
+  event: SlackMessageEvent,
+): boolean {
+  if (trigger.channel !== event.channel) return false;
+  if (!trigger.pattern) return true;
+  return new RegExp(trigger.pattern).test(event.text);
```
anmolm-growthx, line 269: 🟡 `new RegExp(trigger.pattern)` is compiled and executed against every message in the trigger channel on each event. The pattern is only validated for compilability at load (definition.ts:170-174), not for catastrophic backtracking — so an owner-authored pattern can ReDoS the bot on arbitrary channel text, and even benign patterns pay a recompile per message. HOW: precompile the RegExp once at load and store it on the trigger; consider a length/complexity guard or a non-backtracking matcher.

src/workflows/worklog.ts
```diff
@@ -0,0 +1,315 @@
...
+  const result = await runCommand("gh", ["api", "user", "--jq", ".login"], cwd);
+  if (result.exitCode !== 0 || !result.stdout.trim()) {
+    throw new Error(result.stderr || `gh api user exited ${result.exitCode}`);
+  }
+  return result.stdout.trim();
+}
+
+export function parseGithubSlug(remoteUrl: string): string | null {
+  const match =
+    remoteUrl.match(/^git@github\.com:([^/]+\/[^/.]+)(?:\.git)?$/) ??
```
anmolm-growthx: 🟡 `parseGithubSlug` uses `[^/.]+` for the repo segment, so it stops at the first dot — a repo whose name contains a dot (e.g. `org/my.repo` or `org/site.io`) is truncated to the wrong slug, and the subsequent `gh pr list --repo (wrong-slug)` then yields no PRs (or errors) with no signal. HOW: capture the repo name up to an optional trailing `.git` boundary (strip a trailing `.git`, then split on `/`) instead of excluding dots from the name character class.

src/workflows/definition.ts
```diff
@@ -0,0 +1,333 @@
...
+function parseConcurrency(raw: unknown): WorkflowConcurrency {
+  if (raw == null) return "skip";
+  const value = stringValue(raw, "concurrency");
+  if (value === "skip" || value === "parallel") return value;
+  throw new Error(`Invalid concurrency: ${value}`);
+}
+
+function validateCron(cron: string, timezone: string): void {
+  try {
+    CronExpressionParser.parse(cron, { tz: timezone });
```
anmolm-growthx, line 287: Note — root-cause of the ~7.4s `accepts a valid workflow schema` test (over the 5s limit): it is the first timezone-aware cron parse. `CronExpressionParser.parse(cron, { tz })` pulls in cron-parser then luxon then the ICU/IANA timezone database, and the first zone resolution in a fresh process pays the entire cold-init cost. That test is the only validation test with a `schedule` trigger (the others use command triggers), so it absorbs the one-time cost. The validation logic itself is pure and linear — no regex backtracking, no sync fs, no blocking loop — so this is a test artifact, not a per-validation cost; production pays it once at startup. To make the test fast/honest, warm the parser once in a `beforeAll`.

src/index.ts
```diff
@@ -214,6 +249,16 @@ setInterval(() => {
     log.error("reconcile", `tmux reconcile threw: ${err instanceof Error ? err.message : String(err)}`);
   }
 
+  try {
+    await workflowRegistry.startWatching();
```
anmolm-growthx, line 257: 🟡 No symmetric teardown. `startWatching()` opens `fs.watch` handles and the scheduler arms timers, but `setupGracefulShutdown` (src/lifecycle/shutdown.ts) is never handed the workflow components — so on SIGTERM/SIGINT `workflowRegistry.stopWatching()`, `workflowScheduler.stop()`, and `workflowStore.close()` never run. The SQLite store therefore never closes cleanly (no WAL checkpoint on shutdown), and the implemented stop/close methods are effectively dead in production. HOW: pass the registry/scheduler/store into the shutdown handler and call their stop/close in the teardown sequence.

anmolm-growthx:
Reviewed the PR. Here's what I found 🤠

Overall: Clean, well-typed dynamic workflow runtime (loader + registry + scheduler + executor + store + Slack controller) with good last-known-good reload semantics; a few real correctness/security gaps to close before it carries scheduled production traffic.

───
🔴 Issues to fix
1. scheduler.ts:131 — setTimeout delay is unclamped. Any schedule whose next fire is more than ~24.8 days out (monthly/quarterly/yearly cron) exceeds the 32-bit timer max (2147483647 ms), gets clamped to ~1 ms, fires immediately, then reschedules to the same far target and fires again — a repeated early-run loop. WHY: those workflows silently misfire. HOW: Math.min(delay, 2_147_483_647) plus a continuation timer when the real target is further out.
2. executor.ts:169 (+ definition.ts:230-243) — permissions.tools is validated but never enforced. The runner is spawned in the Junior repo root (session.cwd = process.cwd()) with externally-authored agentName/model/prompt, and overlay defs come from the agents-org submodule. The declared tool allowlist is cosmetic, so a crafted overlay workflow can run an arbitrary agent with full tool access. WHY: the schema advertises a security boundary that does not exist. HOW: translate tools into real runner restrictions before spawn, or drop it and document that defs are fully trusted.
3. executor.ts:210 (+ definition.ts:189-194) — the docs output path is dead. emitOutputs skips docs outputs; the artifact is always written to artifactPathFor(...). The declared path never takes effect and its load-time validation guards a value nothing reads. HOW: honor the path, or remove it from the schema.

───
🟡 Worth considering
1. definition.ts:190 — the docs-path containment check is a string prefix test that permits ".." traversal; latent only because the path is currently unused (see 🔴 #3). Normalize with path.resolve and assert containment.
2. scheduler.ts:74 (and handleScheduledRun:149-154) — TOCTOU on the running set: ensureState awaits before the skip check, so two concurrent runNow calls can both admit and double-run a concurrency:skip workflow. Mark running synchronously before the first await.
3. controller.ts:269 — new RegExp(trigger.pattern) is recompiled and run against every channel message; patterns are validated for compilability but not backtracking, so an owner pattern can ReDoS the bot. Precompile at load and bound complexity.
4. worklog.ts:246 — parseGithubSlug uses [^/.]+ for the repo segment, truncating repos whose names contain a dot to the wrong slug (PR collection then silently returns nothing). Strip a trailing .git instead of excluding dots.
5. index.ts:253 / lifecycle/shutdown.ts — graceful shutdown never calls workflowScheduler.stop(), workflowRegistry.stopWatching(), or workflowStore.close(); the SQLite store never closes cleanly and those teardown methods are dead in production. Wire them into setupGracefulShutdown.
6. Test coverage — definition/registry/scheduler have tests, but executor.ts, controller.ts, store.ts, and worklog.ts have none. The runner-summary/fallback path, output emission, and the git/gh worklog parsing are untested.
7. definition.ts:277 — root-cause of the ~7.4s validation test: first timezone-aware cron parse triggers cron-parser→luxon→ICU timezone cold-init. It is a one-time process cost (and a test artifact), NOT regex backtracking, sync fs, or a blocking loop. Warm the parser in a beforeAll if you want the test honest.

───
✅ What is good
- Strong typing throughout — no any in src/workflows.
- Reload is robust: invalid/partially-written files are caught per-file and last-known-good definitions are retained (registry.ts:112-125); fs.watch is debounced (350ms) and watchers/timers have close/clear methods.
- workflow_runs has a (workflow_name, started_at DESC) index and listRuns is LIMIT-bounded — no unbounded scans.
- enabled is a real file-level kill switch; effective state = file enabled AND status active is implemented consistently.
- Worklog has a deterministic fallback and an injectable runCommand, so the core path is testable without spawning git/gh.

Biggest priority: enforce permissions.tools on the spawned runner (🔴 #2) — combined with overlay defs from the submodule and cwd = repo root, an unrestricted runner is the one issue with real blast radius.

Note: bun test has one unrelated failure in src/mcp/slack-server.test.ts ("finds private overlay agent definitions") — environmental (agents-org submodule not initialized in this checkout), not a defect in this PR. typecheck is green.

PranavBakre:
Addressed the review comments in `8895d84`:

- Clamped long scheduler timers to the 32-bit `setTimeout` max and added continuation scheduling for far-future cron fires.
- Fixed `concurrency: skip` admission by claiming the run before the first executor await.
- Docs outputs now honor the configured workflow artifact path, and validation normalizes/rejects absolute or traversal paths.
- Clarified `permissions.tools` as trusted workflow capability metadata rather than a runner sandbox, and moved runner summaries out of the repo root to `/tmp/junior-utility`.
- Fixed dotted GitHub repo remote parsing and added worklog parser tests.
- Wired workflow scheduler/registry/store teardown into graceful shutdown.

Verification:
- `bun run typecheck`
- `bun test src/workflows`
- `bun run build`
- `bun test`

PranavBakre (commented):
Re-reviewed 8895d84. The previous blocker is fixed: invalid private overlays now fail closed on cold boot instead of falling back to public definitions.\n\nNo new blocking findings.\n\nVerification run locally:\n- bun run typecheck\n- bun test src/workflows\n- bun run build\n- bun test (still has the existing private overlay fixture/environment failure in src/mcp/slack-server.test.ts: expected agents-org/db-executioner.md, received [])

PR #46, opened 2026-05-21
Add direct worklog cron

PranavBakre:
## Summary
- Adds an env-gated daily worklog cron that collects configured repo commits and GitHub PRs.
- Writes a markdown worklog snapshot and posts a grouped Slack summary.
- Adds deterministic fallback summary formatting plus optional runner-based summarization.
- Documents worklog env vars and submodule layout guidance.

## Verification
- `bun run typecheck`
- `bun test src/cron src/worklog src/config.test.ts`
- `bun run build`
- `bun test`

## Notes
- Rebased onto `personal/main` after PR #45 was merged.
- The older local `docs/features/dynamic-workflows.md` draft was resolved in favor of the newer merged PR #45 doc.

### DIFF
--- a/src/worklog/activity.ts
+++ b/src/worklog/activity.ts
@@ -1,0 +1,288 @@
+import { mkdir, writeFile } from "node:fs/promises";
+import { join } from "node:path";
+import type { RepoConfig } from "../config.ts";
+
+export interface WorklogCommit {
+  repo: string;
+  hash: string;
+  shortHash: string;
+  date: string;
+  subject: string;
+}
+
+export interface WorklogPr {
+  repo: string;
+  number: number;
+  title: string;
+  state: string;
+  url: string;
+  updatedAt: string;
+  createdAt?: string;
+  mergedAt?: string | null;
+  headRefName?: string;
+  baseRefName?: string;
+}
+
+export interface WorklogActivity {
+  generatedAt: string;
+  since: string;
+  until: string;
+  repos: string[];
+  commits: WorklogCommit[];
+  prs: WorklogPr[];
+  errors: string[];
+}
+
+export interface CommandResult {
+  exitCode: number;
+  stdout: string;
+  stderr: string;
+}
+
+export type RunCommand = (
+  command: string,
+  args: string[],
+  cwd: string,
+) => Promise<CommandResult>;
+
+export interface CollectWorklogOptions {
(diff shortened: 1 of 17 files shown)


PranavBakre (commented):
Reviewed. Found two worklog issues worth fixing before merge.

src/worklog/activity.ts
```diff
@@ -0,0 +1,288 @@
...
+  for (const key of ["user.email", "user.name"]) {
+    const result = await runCommand("git", ["config", "--get", key], cwd);
+    if (result.exitCode === 0 && result.stdout.trim()) {
+      return result.stdout.trim();
+    }
+  }
+  return null;
+}
+
+export function parseGithubSlug(remoteUrl: string): string | null {
```
PranavBakre, line 260: This parser drops valid GitHub repositories whose repo name contains a dot. For example `https://github.com/org/site.io.git` normalizes to no match because `[^/.]+` rejects `site.io`; `collectPullRequests` then returns `[]` with no error, so the daily worklog silently omits PRs for those repos. PR #45 already fixed the same edge case in the dynamic workflow worklog path by stripping a trailing `.git` first and then matching `owner/repo` segments; this direct cron path should use the same parser/test coverage.

src/worklog/summary.ts
```diff
@@ -0,0 +1,161 @@
...
+): Promise<string | null> {
+  const channel = config.worklog.channel ?? "worklog-cron";
+  const session = createSession(
+    `cron-worklog-${activity.until.slice(0, 10)}`,
+    channel,
+    "quiet",
+    config.runner.provider,
+    config.claude.defaultDriver,
+  );
+  session.cwd = process.cwd();
```
PranavBakre, line 59: The summarizer runner is launched from `process.cwd()`, which means this utility summarization runs with Junior’s project workspace/MCP context even though it only needs the JSON activity payload. That is the same footgun we just fixed in PR #45 by using a utility cwd (`/tmp/junior-utility`) so project MCP wiring is skipped for non-repo workflows. Please move this to a utility cwd; otherwise a Slack-summary prompt can accidentally inherit repo tools/context and broaden what should be a read-only compression step.

PranavBakre:
Closing this direct cron PR. Worklog should continue through the workflow definition path instead.

PR #48, opened 2026-05-21
fix: surface workflow runs and parse opencode output

PranavBakre:
## Summary
- Surface workflow states and recent runs in Slack Home and the local web UI.
- Expose workflow observability through `/api/workflows` and persist provider session IDs on workflow runs.
- Parse current OpenCode part-stream events so workflow output, tool events, and usage are captured correctly.

## Test plan
- [x] `bun run typecheck`
- [x] focused OpenCode/workflow tests
- [x] `bun test` passing with 558 pass / 0 fail


PranavBakre:
review: no blockers found.\n\nVerified locally:\n- bun run typecheck\n- bun test\n\nCouldn’t submit an approval because GitHub rejects approving your own PR from the current auth.

PR #50, opened 2026-05-23
Clarify memory consolidation plan

PranavBakre:
## Summary
- define raw memory classes for Junior without tying them to Markdown files
- make V2 consolidation/dreaming responsibilities explicit
- document that memory consolidation should run through dynamic workflows rather than a bespoke scheduler
- document usage boundaries so raw capture stays cheap and expensive extraction runs offline
- add stale-fact metadata, routing-memory validity, and ingestion/consolidation learning examples
- add recall, routing, stale-fact, cost, and consolidation benchmark gates

## Verification
- git diff --check


anmolm-growthx:
Pushed the design-review doc updates as #51, which targets this branch — merge it and the commit folds straight into this PR. (Couldn't push directly onto the branch: `anmolm-growthx` has no write access to the repo.)

What changed, from the discussion with Anmol:

1. **ingestion-rule-learning** — "dreaming writes the rules" is now the primary path; Popper/Metagol demoted to a fallback spike. Collapses ILP-V3 into V2 (no labeled-data gate, no second runtime, cost flips positive). Safety pipeline unchanged — only the rule author changes. Confabulation + model-monoculture risks called out.
2. **agent-selection** — added "the DSL is the contract between extraction and execution, not between model and host": the LLM emits validated structured facts (tool-call args = bounded vocabulary), the host compiles to DSL terms and runs fixed goals. `tau-prolog` (in-process) only when recursion/proof-trees are needed; TS evaluator otherwise.
3. **associative-memory** — new Identity & Deduplication section (dedup vs supersession; identity-key merge bumps `use_count`), and an explicit "v1 won't feel associative until edges exist" expectation.

src/memory/sqlite.ts
```diff
@@ -0,0 +1,756 @@
...
+      );
+  }
+
+  async upsertEvent(event: MemoryEventInput): Promise<void> {
+    const importance = event.importance ?? 0.5;
+    const txn = this.db.transaction(() => {
+      this.upsertNode(event.id, "event", event.createdAt);
+      this.db
+        .query(
+          `INSERT INTO memory_event (
```
anmolm-growthx, line 87: **blocker — no identity-key dedup; `use_count` / `frequency` are dead signals.**

The just-merged Identity & Deduplication section says derived nodes must dedup against a normalized identity key: "If a live node already matches the key, do not create a new node — instead bump `use_count`, refresh `last_used_at`, and append the new `source_record_id` to its provenance."

This insert only dedups on the *caller-supplied* `id` (`ON CONFLICT(id)`). Two ingests of the same fact with different ids produce two nodes. And `use_count` / `last_used_at` are **never written anywhere in this file** (they are only read in scoring at line 671 and the archive filter at line 364). So `frequency` in the ranking model is permanently 0 — it measures "duplicate rows we failed to merge," not recurrence, which is exactly the failure the spec warns about.

Suggest for v1: compute a normalized identity key (normalized claim text, or `(entity, predicate, normalized_value)` for facts); on a live-node match bump `use_count` / `last_used_at` + append provenance instead of inserting. At minimum increment `use_count` / `last_used_at` on the conflict path so the signal is non-zero.

src/memory/sqlite.ts
```diff
@@ -0,0 +1,756 @@
...
+          ...row,
+          reasons: ["Matched requested tag/entity"],
+          activation: 2,
+        });
+      }
+      seedIds.add(row.id);
+    }
+
+    if (depth > 0 && seedIds.size > 0) {
+      for (const row of this.edgeRows([...seedIds], depth, limit * 6)) {
```
anmolm-growthx, line 549: **blocker — bounded-traversal controls are incomplete vs the Performance Expectations checklist.**

The spec lists seven traversal controls (depth cap, seed cap 10–30, per-node/type fanout cap, indexes, stop-low-score-paths, avoid generic tags, allow-list before traversing tag/entity nodes). This recall implements depth (`Math.min(Math.max(depth,0),3)`, default 2) and the indexes — good — but:

- **No seed cap.** `[...seedIds]` is every FTS hit (`limit*4`) plus *every* tag/entity match. `tagEntityRows` is uncapped, so a popular tag floods the seed set. Spec wants top 10–30 seeds.
- **No per-node fanout cap and no low-score cutoff** in the recursive step. The only bound is the final `LIMIT ?3` (`limit*6`) on `ranked`; the CTE still expands every edge from every seed down to max depth before that limit applies. On a dense graph this is the runaway case the spec calls out.

For v1 these caps are the whole point of "bounded" traversal. Suggest: sort seeds by activation and take the top N before calling edgeRows, and add a score-floor guard in the recursive step so weak paths stop expanding.

src/memory/sqlite.ts
```diff
@@ -0,0 +1,756 @@
...
+      .query<CandidateRow, [string, number, number]>(
+        `WITH RECURSIVE seed(id) AS (
+           SELECT value FROM json_each(?1)
+         ), related(id, depth, score, path) AS (
+           SELECT e.dst_id, 1, e.weight, e.src_id || '>' || e.dst_id
+           FROM edge e JOIN seed s ON s.id = e.src_id
+           UNION ALL
+           SELECT e.dst_id, r.depth + 1, r.score * e.weight * 0.7, r.path || '>' || e.dst_id
+           FROM edge e JOIN related r ON e.src_id = r.id
+           WHERE r.depth < ?2 AND instr(r.path, e.dst_id) = 0
```
anmolm-growthx: **warning — substring-based cycle guard can both over- and under-match.**

`instr(r.path, e.dst_id) = 0` does a raw substring test on the delimiter-joined id path. Node ids are not prefix-free here (`event-1` vs `event-12`, `tag:memory` vs `tag:memory_architecture`, `entity:system:junior` vs `entity:system:junior_bot`). `instr` reports a false "already visited" when a candidate id is a substring of an id already on the path, silently pruning a legitimate edge; a delimiter collision can also mask a real cycle the other way.

Suggest anchoring each id with delimiters on both sides before the membership test (wrap every path segment, then test for the delimiter-bracketed id) so matches are whole-id only.

src/memory/sqlite.ts
```diff
@@ -0,0 +1,756 @@
...
+    const repeatedThreshold = options.repeatedCorrectionThreshold ?? 2;
+    const decisions: ConsolidationDecisionRecord[] = [];
+    const promotedMemoryIds: string[] = [];
+    const archivedEventIds: string[] = [];
+    const proposedRuleIds: string[] = [];
+
+    const archiveRows = this.db
+      .query<{ id: string; source_record_id: string }, [number, number]>(
+        `SELECT id, source_record_id FROM memory_event
+         WHERE active = 1 AND importance <= ? AND use_count = 0 AND created_at < ?`,
```
anmolm-growthx, line 586: **warning — archive predicate is effectively `use_count = 0` for everything.**

Because `use_count` is never incremented anywhere in the store, `use_count = 0` is always true. Combined with the dedup gap, this filter degenerates to "archive every active event older than the window with importance at or below the threshold," regardless of how often it was actually recalled or how many sources back it. Once dedup/use-count tracking lands this becomes correct; until then the "unused" half of the predicate is a no-op and the consolidation test only passes because the fixture event genuinely has `use_count = 0`. Worth a comment or a follow-up so it is not mistaken for working recency-aware archiving.

src/memory/sqlite.ts
```diff
@@ -0,0 +1,756 @@
...
+          event.outcome ?? null,
+          importance,
+          event.createdAt,
+          event.sourceTs ?? null,
+          event.sourceUrl ?? null,
+        );
+      this.replaceProvenance(event.id, [event.sourceRecordId]);
+      this.replaceTags(event.id, "event", event.tags ?? []);
+      this.replaceEntities(event.id, "event", event.entities ?? []);
+      this.upsertSearchDoc({
```
anmolm-growthx, line 116: Nice — FTS sync is done the right way. `upsertSearchDoc` writes `memory_search_doc` (the authoritative projection) and the `memory_fts` delete+insert inside the **same** transaction as the node/event write, so the derived index can never drift from the source of truth, and `rebuildSearchIndex()` gives the rebuild path the spec asks for. This satisfies both FTS-sync options in the spec. (Same pattern correctly repeated for lessons and facts.)

src/memory/sqlite.test.ts
```diff
@@ -0,0 +1,296 @@
...
+    db.run("DELETE FROM memory_fts");
+    expect(await store.search("PR review")).toEqual([]);
+
+    await store.rebuildSearchIndex();
+
+    const results = await store.search("PR review");
+    expect(results.map((result) => result.id)).toEqual(["event-1"]);
+  });
+
+  it("recalls memories by tag, entity, and bounded edge traversal", async () => {
```
anmolm-growthx, line 131: **warning — tests cover happy-path CRUD + supersession but not the hard parts.**

Good coverage on source/event separation, FTS sync, FTS rebuild, supersession suppression, and consolidation. Missing the cases most likely to regress:

- **Dedup / use_count:** no test ingests the same fact twice and asserts a single node with `use_count` incremented (it would currently fail — see the dedup comment).
- **Traversal bounds:** this test exercises a 2-node graph at `depth: 2`, which never stresses depth/seed/fanout caps. A fixture with a cycle and a high-fanout tag would catch the unbounded-seed and substring-cycle-guard issues.
- **Substring-id cycle guard:** add ids like `event-1` and `event-12` on a path to lock the `instr` behavior.

Not blocking on its own, but these are the paths the spec treats as load-bearing, so they are the ones worth fixtures.

anmolm-growthx:
Reviewed the PR. Here's what I found 🤠

Overall: V1 SQLite memory store + V2 consolidation scaffold, reviewed against the just-merged design docs as the spec. Schema, FTS discipline, source/derived split, and supersession are solid; the dedup/use-count path and bounded-traversal caps are the gaps.

───
🔴 Issues to fix
1. **`src/memory/sqlite.ts:84` (and `:133`, `:183`) — no identity-key dedup; `use_count` / `frequency` are dead signals.** The spec's Identity & Deduplication section requires derived nodes to dedup on a normalized identity key and bump `use_count` / `last_used_at` + append provenance on a match. The inserts only dedup on the caller-supplied `id`, so the same fact ingested with a new id duplicates. And `use_count` / `last_used_at` are never written anywhere in the file — they are only read in scoring (`:671`) and the archive filter (`:364`). Result: `frequency` is permanently 0 and measures "rows we failed to merge," not recurrence — the exact failure the spec calls out. Fix: identity-key lookup on the upsert path; bump `use_count` / `last_used_at` + append provenance instead of inserting (or at minimum increment on the conflict path).

2. **`src/memory/sqlite.ts:329` / `:645` — bounded traversal is only partly bounded.** Depth cap (default 2, max 3) and the edge indexes are correctly implemented. Missing from the spec control list: no seed cap (passes every FTS hit plus every uncapped tag/entity match as a seed — spec wants top 10–30), no per-node fanout cap, and no low-score path cutoff in the recursive CTE (only a final LIMIT on ranked results, applied after full expansion). On a dense graph this is the runaway case. Fix: slice seeds to top N by activation before traversal, and add a score-floor stop condition in the recursive step.

───
🟡 Worth considering
1. **`src/memory/sqlite.ts:645` — substring cycle guard.** `instr(r.path, e.dst_id) = 0` substring-matches on the joined id path; ids are not prefix-free (`event-1` vs `event-12`, `tag:memory` vs `tag:memory_architecture`), so it can falsely prune a legit edge or mask a real cycle. Anchor each id with delimiters on both sides before the membership test.
2. **`src/memory/sqlite.ts:364` — archive predicate.** `use_count = 0` is always true today (use_count never incremented), so the "unused" half of the filter is a no-op until dedup/use-count lands. The consolidation test only passes because the fixture genuinely has use_count 0. Worth a comment/follow-up so it is not mistaken for working recency-aware archiving.
3. **`src/memory/sqlite.test.ts:131` — test coverage.** Good on CRUD, source/event split, FTS sync+rebuild, supersession, consolidation. Missing the load-bearing paths: re-ingest dedup/use_count (would fail today), traversal under cycles + high-fanout tags, and substring-id cycle behavior.
4. **Tag/entity traversal is a no-op in the current edge model.** `tagged_as` / `mentions` edges are always src=memory, dst=tag/entity, and traversal only follows src_id = seed. So tags/entities are reached as leaves and traversal never hops memory → tag → other-memory. Co-occurrence via shared tags is not actually discovered by spreading activation — only directed memory-to-memory edges (`lesson_from`, `supersedes`, etc.) propagate. Fine for v1 if intentional, but it means the spec "avoid generic high-fanout tag" worry does not bite yet — and neither does the intended tag-based association.

───
✅ What is good
- **FTS discipline is exactly right.** `memory_search_doc` is the authoritative projection; `memory_fts` is rewritten in the same transaction as the node/event/lesson/fact write, and `rebuildSearchIndex()` provides the rebuild path. Satisfies both spec options; no drift.
- **Source-vs-derived split is enforced.** `memory_source_record` is append-only (plain INSERT, no upsert), `memory_event.source_record_id` is NOT NULL with a FK back to the source, and provenance is tracked separately.
- **Hot path is LLM-free and cheap.** Capture/recall is pure SQLite + TypeScript scoring; all LLM/extraction work is deferred to the operator-triggered consolidation workflow. Honors the "never call an LLM just to decide whether to store raw memory" rule.
- **Temporal supersession works.** `supersedes` edges set `invalid_at` / `superseded_by` on the node, and recall suppresses invalid rows by default while `includeInvalid` exposes history.
- **SQL is fully parameterized** (only interpolation is a fixed table-name allow-list in `rowsForJoin` and the `candidateSelect` score expression), multi-row writes are wrapped in transactions, and the required indexes (`edge.src_id`, `edge.dst_id`, `(type, src_id)`) are present.

Biggest priority: the dedup / `use_count` path (🔴 #1) — without it `frequency` ranking and the consolidation archive predicate both measure nothing, and recall accumulates duplicate nodes. Land that, then the seed/fanout caps (🔴 #2).

src/memory/sqlite.ts
```diff
@@ -0,0 +1,756 @@
...
+          candidates.set(row.id, {
+            ...row,
+            reasons: ["Related by edge traversal"],
+            activation: row.fts_rank ?? 0,
+          });
+        }
+      }
+    }
+
+    return [...candidates.values()]
```
PranavBakre: `recall()` reads `use_count`/`last_used_at` for ranking and `consolidate()` archives `use_count = 0` events, but nothing increments those fields when memories are actually returned. That means a low-importance memory can be recalled repeatedly and still look unused, then get archived on the next consolidation pass. Please update the selected result ids transactionally before returning, or keep usage-based scoring/archival out until the write path exists.

src/memory/sqlite.ts
```diff
@@ -0,0 +1,756 @@
...
+         ON CONFLICT(src_id, dst_id, type) DO UPDATE SET
+           weight = excluded.weight,
+           directed = excluded.directed`,
+      )
+      .run(
+        edge.srcId,
+        edge.dstId,
+        edge.type,
+        edge.weight ?? 1,
+        edge.directed === false ? 0 : 1,
```
PranavBakre, line 867: `directed: false` is persisted but never honored. `edgeRows()` only traverses `src_id -> dst_id`, so an undirected associative edge only works from whichever side happened to be inserted as `srcId`. Either insert the reverse row for `directed === false` or teach the recursive query to traverse reverse edges where `directed = 0`; otherwise recall misses half of the graph.

PranavBakre (commented):
Reviewed. Two correctness issues below; tests/typecheck are green, but these make the new memory signals lie in normal use.

PranavBakre (commented):
Good work — clean implementation with solid test coverage (17 tests, all passing). Two bugs to fix:

**1. Duplicate method declaration (`src/memory/store.ts:35-36`)**
```ts
consolidate(options?: ConsolidationOptions): Promise<ConsolidationResult>;
consolidate(options?: ConsolidationOptions): Promise<ConsolidationResult>;
```
Delete the duplicate.

**2. Operator precedence bug (`src/support/router.ts:508`)**
```ts
const h = (memory.title ?? "" + " " + memory.body).toLowerCase();
```
`+` binds tighter than `??`, so this evaluates as `memory.title ?? (" " + memory.body)`. When title is set, body is never consulted. Fix:
```ts
const h = ((memory.title ?? "") + " " + memory.body).toLowerCase();
```

Otherwise the code is well-structured — types → store interface → SQLite impl → ingestion → CLI → wiring. The dual-path consolidation (native deterministic pass → LLM inspection pass) is a clean design. The 60s rule cache in ingestion is smart for the hot path. Ship after the two fixes.

PR #69, opened 2026-05-28
Fix Codex app-server Claude model aliases

PranavBakre:
## What
- Ignore Claude-only model aliases (`sonnet`, `opus`, `haiku`, `claude-*`) when spawning Codex app-server sessions.
- Fall back to `CODEX_MODEL` or Codex's account default instead of passing an unsupported model through.
- Add coverage for model resolution behavior.

## Why
Oogway/worker agent sessions can carry historical Claude model frontmatter. With Codex app-server + ChatGPT auth, passing `sonnet` causes: `The 'sonnet' model is not supported when using Codex with a ChatGPT account.`

## Test
- `bun test src/codex-app-server/spawner.test.ts src/codex-app-server/config.test.ts`
- `bun run typecheck`


pranav-growthx (approved):
Reviewed diff and verified in a combined temp worktree with PRs 70, 71, 72, 74, and 75. Typecheck and full test suite passed.

PR #77, opened 2026-05-29
Clean up spawned process trees on shutdown

PranavBakre:
## Summary
- add process-group cleanup helpers and use them for runner/dev-server process trees
- preserve provider session IDs early and reset busy sessions/agents to idle on shutdown for resumption
- stop workflow runners on scheduler shutdown and document the cleanup contract

## Verification
- bun run typecheck
- bun test
- pgrep -xlaf bun (clean)

### DIFF
--- a/src/session/manager.ts
+++ b/src/session/manager.ts
@@ -341,17 +341,24 @@ export class SessionManager {
    */
   async terminateActiveRuns(reason = "shutdown"): Promise<void> {
     const handles = [...this.handles.entries()];
+    const unsettledHandles = new Set(handles.map(([, handle]) => handle));
     for (const [key, handle] of handles) {
       this.handles.delete(key);
       handle.kill("SIGINT");
     }
 
     await Promise.race([
-      Promise.allSettled(handles.map(([, handle]) => handle.result)),
+      Promise.allSettled(
+        handles.map(([, handle]) =>
+          handle.result.finally(() => {
+            unsettledHandles.delete(handle);
+          })
+        ),
+      ),
       new Promise((resolve) => setTimeout(resolve, 10_000)),
     ]);
 
-    for (const [, handle] of handles) {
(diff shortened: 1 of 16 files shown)


src/session/manager.ts
```diff
@@ -326,6 +334,52 @@ export class SessionManager {
...
+      handle.kill("SIGINT");
+    }
+
+    await Promise.race([
+      Promise.allSettled(handles.map(([, handle]) => handle.result)),
+      new Promise((resolve) => setTimeout(resolve, 10_000)),
+    ]);
+
+    for (const [, handle] of handles) {
+      handle.kill("SIGKILL");
```
PranavBakre, line 362: warning: this sends SIGKILL to every handle even when `Promise.allSettled` won the race and the process already exited cleanly after SIGINT. With the new process-tree helper, `handle.kill("SIGKILL")` first targets `-pid` and then falls back to the positive pid if that group is gone; after a clean exit that makes a second kill capable of hitting a reused pid. `WorkflowExecutor.terminateActiveRuns()` avoids this by checking whether the handle is still active before escalating. Please do the same here: only SIGKILL handles whose result did not settle before the 10s grace window, or track/remove settled handles before the escalation loop.

PranavBakre (commented):
Found one shutdown-safety issue to fix before merge.

pranav-growthx:
Addressed the shutdown-safety review. `SessionManager.terminateActiveRuns()` now tracks unsettled handles after SIGINT and only sends SIGKILL to handles whose `result` did not settle within the grace window.

Verification:
- `bun test src/session/manager.test.ts`
- `bun run typecheck`

pranav-growthx (commented):
Re-review clean. The shutdown escalation now tracks unsettled handles and the tests assert the settled path only receives SIGINT. No new issues found.

PR #78, opened 2026-05-29
Add internal agent dispatch MCP tool

PranavBakre:
## Summary
- add `agent_dispatch` to the Slack MCP server for direct persistent-agent dispatch without Slack directive posts
- pass `SessionManager` into the MCP server so the tool can call `handleAgentMessage`
- enforce existing dispatch allow-list before spawning the target agent

## Verification
- `bun test src/mcp/slack-server.test.ts`
- `bun run typecheck`

### DIFF
--- a/src/claude/spawner.ts
+++ b/src/claude/spawner.ts
@@ -129,7 +129,7 @@ export function spawnClaude(
   };
 }
 
-function writeClaudeMcpConfig(session: ThreadSession): string {
+export function writeClaudeMcpConfig(session: ThreadSession): string {
   mkdirSync(MCP_CONFIG_DIR, { recursive: true });
   const agent = session.activeAgentName ?? "default";
   const path = join(MCP_CONFIG_DIR, `${session.threadId}-${agent}.json`);
--- a/src/claude/spawner.ts
+++ b/src/claude/spawner.ts
@@ -1,4 +1,5 @@
-import { resolve } from "node:path";
+import { mkdirSync, writeFileSync } from "node:fs";
+import { join, resolve } from "node:path";
 import type { Config } from "../config.ts";
 import type { AgentIdentity, ThreadSession } from "../session/types.ts";
 import type { ContentBlockToolUse, StreamEvent } from "./types.ts";
@@ -6,9 +7,9 @@ import type { RunnerEvent, SpawnHandle, SpawnResult } from "../runners/types.ts"
 import { buildRunnerRuntime } from "../runners/runtime.ts";
 import { buildClaudeArgs } from "./args.ts";
 import { createStreamParser } from "./parser.ts";
+import { buildSlackMcpUrl } from "../mcp/context.ts";
 
-// Junior's .mcp.json — pass to spawned processes so they find the slack-bot server
-const PROJECT_MCP_CONFIG = resolve(import.meta.dirname ?? ".", "../../.mcp.json");
+const MCP_CONFIG_DIR = resolve(import.meta.dirname ?? ".", "../../data/mcp-configs");
 
 export function spawnClaude(
   session: ThreadSession,
@@ -24,7 +25,7 @@ export function spawnClaude(
     botToken,
     agentIdentity,
   });
-  const mcpConfigPath = runtime.needsProjectMcp ? PROJECT_MCP_CONFIG : undefined;
+  const mcpConfigPath = writeClaudeMcpConfig(session);
   const args = buildClaudeArgs(session, prompt, config, mcpConfigPath);
 
   const proc = Bun.spawn(["claude", ...args], {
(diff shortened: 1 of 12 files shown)


src/mcp/slack-server.ts
```diff
@@ -319,6 +322,78 @@ function registerTools(server: McpServer) {
...
+          content: [
+            {
+              type: "text" as const,
+              text: `Error: unknown agent "${targetAgent}". Use agent_search to find registered agents.`,
+            },
+          ],
+        };
+      }
+
+      const caller = source_agent?.trim() || "default";
```
PranavBakre: blocker: this makes the allow-list effectively caller-controlled. Any persistent agent with MCP access can omit `source_agent` and get `default`'s orchestrator permissions, or set `source_agent` to `lead`, then dispatch agents it is not allowed to dispatch. That bypasses the existing router guard, where caller identity comes from the self-bot username rather than model-provided input. Please derive the caller from trusted per-run context, or fail closed when the caller cannot be authenticated; don't default to an orchestrator identity.

PranavBakre:
fixed the review blocker in 03eb855. `agent_dispatch` now derives the caller from trusted session state: exactly one active busy top-level/persistent agent must be present, otherwise it fails closed. Removed caller-provided `source_agent`.

src/mcp/slack-server.ts
```diff
@@ -479,6 +563,40 @@ async function withMemoryStore<T>(fn: (memory: ReturnType<typeof createMemorySto
...
+
+  const candidates: string[] = [];
+  if (session.status === "busy") {
+    candidates.push(session.activeAgentName ?? session.topLevelTmuxAgent ?? topLevelAgentForSession(session));
+  }
+  for (const agentSession of Object.values(session.agentSessions ?? {})) {
+    if (agentSession.status === "busy") candidates.push(agentSession.agentName);
+  }
+
+  const unique = [...new Set(candidates.filter(Boolean))];
```
pranav-growthx: blocker: resolving the caller by requiring exactly one busy agent in the thread breaks the persistent-agent model. Threads are designed to have multiple agents busy concurrently; after lead/default dispatches one worker, any second `agent_dispatch` call from the same still-running caller will see both the caller and the newly spawned worker as busy and fail. Same for an allowed worker dispatch while another worker is active. The caller needs to come from trusted per-request/run context, not from a thread-wide "only busy agent" heuristic.

pranav-growthx:
fixed the concurrency blocker in 4b95a62. `agent_dispatch` now gets caller identity from the per-run MCP URL context, not thread-wide busy-agent state. The tool also checks the requested channel/thread against that authenticated context before dispatching. Updated OpenCode, Codex app-server, and Claude MCP config generation to pass contextual Slack MCP URLs.

src/claude/spawner.ts
```diff
@@ -128,6 +129,34 @@ export function spawnClaude(
   };
 }
 
+function writeClaudeMcpConfig(session: ThreadSession): string {
```
pranav-growthx: blocker: this fixes the headless Claude path, but the tmux Claude driver still starts interactive `claude` with `buildClaudeArgs(input.session, "", input.config)` and no generated MCP config. In tmux mode the Slack MCP URL remains the project/root config without `agent/channel/thread`, so `agent_dispatch` always fails with `MCP run context missing`. Since persistent agents can run through the tmux driver (`!driver tmux` / `DEFAULT_CLAUDE_DRIVER=tmux`), the new tool needs the same per-run MCP config there too, or tmux sessions lose the feature entirely.

pranav-growthx:
fixed the Claude tmux gap in 7d307b8. The tmux driver now starts interactive Claude with the same generated per-run MCP config as headless Claude, so the Slack MCP URL includes `agent/channel/thread`. Added a tmux-driver test that reads the generated config from the `--mcp-config` arg and verifies the contextual URL.

src/claude/spawner.ts
```diff
@@ -24,7 +25,7 @@ export function spawnClaude(
     botToken,
     agentIdentity,
   });
-  const mcpConfigPath = runtime.needsProjectMcp ? PROJECT_MCP_CONFIG : undefined;
+  const mcpConfigPath = writeClaudeMcpConfig(session);
```
pranav-growthx: blocker: this drops the existing `runtime.needsProjectMcp` guard, so Claude headless now always gets Junior's generated MCP config even for `session.cwd` utility runs. Those runs deliberately use an explicit utility cwd to avoid Junior's project MCP wiring and rely on their own integrations (for example the `!adhoc`/`!bugs` calendar path and workflow utility runs). Passing `--mcp-config` here can hide those tools and exposes Slack/playwright/mongodb where the runtime policy says MCP should be skipped. Please preserve the carve-out, e.g. only write/pass this config when `runtime.needsProjectMcp` (and make the tmux equivalent respect the same policy).

pranav-growthx:
fixed the Claude MCP utility carve-out in 1427a1a. Headless Claude now only writes/passes the generated MCP config when `runtime.needsProjectMcp` is true, and tmux uses the same `buildRunnerRuntime` decision before constructing interactive Claude args. Added a tmux regression test for `session.cwd` utility runs to ensure `--mcp-config` is omitted. Verified with two clean passes of `bun run typecheck` and `bun test` (654 passing).

pranav-growthx (approved):
approved: previous blocker is fixed. The generated MCP config now preserves the utility-cwd carve-out and tmux gets the same contextual Slack MCP URL as headless runs.\n\nVerified with targeted tests and typecheck.

PR #79, opened 2026-06-01
Wire onboarding agent dispatch context

PranavBakre:
## Summary
- require agents to use memory recall for substantive work, especially routing/setup/env/onboarding questions
- add `slack_send_dm`, a Slack MCP helper that opens a DM by user ID before posting
- document the DM tool in runtime/MCP docs
- update the private agents submodule to include onboarding and credential-agent DM prompt fixes

## Verification
- bun test src/support/router.test.ts src/support/agents.test.ts src/agents/loader.test.ts src/agents/router.test.ts
- bun test src/mcp/slack-server.test.ts src/agents/loader.test.ts src/support/agents.test.ts src/agents/router.test.ts
- bun run typecheck
- loaded onboarding/admin-account/onboard-member through Junior agent loader
- Codex app-server smoke: plain onboarding question -> default Junior emitted !onboarding -> dispatcher targeted onboarding

## Related
- Includes GrowthX-Club/junior-private-agents#4
- Depends on GrowthX-Club/junior-private-agents#5 for the submodule pointer update

### DIFF
--- a/.claude/agents/common/core.md
+++ b/.claude/agents/common/core.md
@@ -10,6 +10,17 @@ Before responding, classify the user's message:
 
 If action is required, do the action. Do not only describe how to do it.
 
+## Memory recall first
+
+For every substantive task, call `memory_recall` before answering or acting when the tool is available. Use the user's request as the query, add relevant repo/product tags when obvious, and read the returned snippets as prior operating context. If no relevant memory is returned, proceed normally and do not mention the empty recall unless it matters.
+
+Use memory especially before:
+
+- routing to another agent
+- answering repo setup, env, credential, workflow, or onboarding questions
+- touching production-connected scripts or data
+- repeating a workflow that Junior may have learned from earlier corrections
+
 Common implicit actions:
 
 | User shape | Required action |
--- a/.claude/agents/common/runtime-environment.md
+++ b/.claude/agents/common/runtime-environment.md
@@ -35,7 +35,7 @@ The active dev config and any production-like fallback paths (e.g. `/etc/hosts` 
 
 Pre-loaded — do NOT `ToolSearch` for them:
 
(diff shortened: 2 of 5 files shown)


agents-org
```diff
@@ -1 +1 @@
-Subproject commit 8016b39e7b4c419c36ee569eaa70811209eabe0c
+Subproject commit d25a1f62d971bcde52f84bb4bb9d45215e3db7ff
```
pranav-growthx: warning: The submodule's new `onboarding.md` env flow tells the worker to DM Pranav/requesters, but Junior's Slack MCP only exposes `slack_send_message(channel_id, ...)`; it has no `conversations.open`/DM-channel helper. For a user ID, `chat.postMessage` targets App Home semantics rather than a normal DM, and the worker has no reliable way to discover/open the requester DM before sending env contents. Since env sharing is the sensitive path this worker owns, either add a DM/open tool and update the prompt to use it, or change the instructions to a flow the available Slack tools can actually execute safely.

pranav-growthx (changes requested):
Found one wiring issue in the new onboarding agent's env approval path.

PranavBakre:
Addressed the Slack MCP DM capability review.

Changes added here:
- new `slack_send_dm(user_id, text, username?, icon_emoji?)` MCP tool
- implementation opens the DM via `conversations.open` before `chat.postMessage`
- unit test covers open-before-post behavior
- runtime/MCP docs now list `slack_send_dm` separately from channel/thread posting
- agents-org prompts now use `slack_send_dm` for env/password delivery instead of treating user IDs as channels

Also opened the follow-up private-agents PR for the prompt updates: GrowthX-Club/junior-private-agents#5.

Verification: MCP/agent tests passed and `bun run typecheck` passed.

pranav-growthx (approved):
Re-reviewed after the DM helper update. The previous env-sharing concern is resolved: `slack_send_dm` opens a DM via `conversations.open`, the onboarding/admin/member prompts now use it for secret delivery, and focused tests + typecheck pass.

PR #81, opened 2026-06-01
Internalize persistent agent directive dispatch

PranavBakre:
## Summary
- internalize pure `!<agent>` runner responses into direct persistent-agent dispatch instead of posting them to Slack
- enforce the same persistent-agent dispatch allow-list on the internal path
- expose/prefer `agent_dispatch` in orchestrator prompts and shared runtime docs

## Tests
- `bun test src/session/manager.test.ts`
- `bun test src/support/router.test.ts src/mcp/slack-server.test.ts`
- `bun run typecheck`

## Notes
- Full `bun test` still has unrelated existing failures in dev-server integration readiness and an OpenCode fallback prompt sync test.


src/session/manager.ts
```diff
@@ -1628,6 +1651,45 @@ export class SessionManager {
...
+    threadId: string,
+    sourceAgentName: string,
+    directives: AgentDirective[],
+  ): Promise<void> {
+    const session = await this.store.get(threadId);
+    if (!session) return;
+
+    const sourceIdentity = identityForAgent(sourceAgentName);
+    const now = Date.now();
+    await Promise.all(
```
PranavBakre: warning: Dispatching every parsed directive concurrently reintroduces the same per-agent race the Slack router avoids. If a response contains two lines for the same worker (for example two `!review ...` directives), both `handleAgentMessage` calls can read the session before either marks `agentSessions.review.status = "busy"`, spawn two review runs, and the second handle overwrites the first in `this.handles`; the first completion is then discarded. Group by `agentName` and await each group serially like `AgentDispatcher` does, or otherwise serialize same-agent dispatches.

PranavBakre:
Addressed review comment https://github.com/PranavBakre/Junior/pull/81#discussion_r3332660826 in `1d6bff3`.

Internal directive dispatch now groups by agent and serializes each same-agent group, matching `AgentDispatcher` behavior, while still allowing different agents to fan out concurrently. Added a regression test for duplicate `!review` directives ensuring the second prompt is buffered instead of spawning a second review handle.

Checks:
- `bun test src/session/manager.test.ts`
- `bun test src/support/router.test.ts src/mcp/slack-server.test.ts`
- `bun run typecheck`

src/session/manager.ts
```diff
@@ -1628,6 +1651,54 @@ export class SessionManager {
...
+    const sourceIdentity = identityForAgent(sourceAgentName);
+    const now = Date.now();
+    const byAgent = new Map<string, Array<{ directive: AgentDirective; index: number }>>();
+    directives.forEach((directive, index) => {
+      const entries = byAgent.get(directive.agentName) ?? [];
+      entries.push({ directive, index });
+      byAgent.set(directive.agentName, entries);
+    });
+
+    await Promise.all(
```
PranavBakre: warning: This still runs different target agents in parallel, and `handleAgentMessage` is not safe to call concurrently for the same thread. With the SQLite store, each call reads a full `ThreadSession`, mutates one `agentSessions` entry, then `set()` rewrites/syncs the whole agent-session set. A response like `!review ...` + `!reproducer ...` can therefore lose whichever worker writes first when the second stale snapshot is persisted. The existing same-agent serialization fixes one case; the internal path should serialize all directives for the thread, or the store update needs to become per-agent/atomic before parallel fan-out is safe.

PranavBakre:
Addressed the latest change request from https://github.com/PranavBakre/Junior/pull/81#discussion_r3332705796 in commit 80343f1.\n\nChanged internal directive dispatch to process all directives sequentially, including cross-agent handoffs, so each handleAgentMessage call persists its session update before the next agent starts. Added a regression test covering !review + !reproducer from the same lead response.\n\nChecks run:\n- bun test src/session/manager.test.ts\n- bun test src/support/router.test.ts src/mcp/slack-server.test.ts\n- bun run typecheck

PR #84, opened 2026-06-03
Improve memory consolidation quality

PranavBakre:
## Summary
- stop memory consolidation from generating tag-count lessons and leave learned tag rules in draft
- rank actionable derived memories above raw events, dedupe identical recall results, and add archive-memory support
- update consolidation workflow guidance and sync the OpenCode fallback core prompt

## Verification
- bun run src/memory/cli.ts consolidate --json
  - succeeded; promoted summary_1779784687_549839_1780462733437
- bun run typecheck
- bun test
  - 672 pass, 0 fail

## Notes
- Ignored local memory DB/learnings updates and unrelated untracked files are not included in this PR.


src/memory/sqlite.ts
```diff
@@ -664,9 +675,6 @@ export class SqliteMemoryStore implements MemoryStore {
       });
       decisions.push(decision);
```
pranav-growthx, line 675: blocker: after removing auto-accept, re-proposing the same `rule_tag_*` ID now resets an already accepted rule back to draft because `proposeRule` upserts `status = excluded.status`. A human can accept `rule_tag_frontend`, but the next consolidation over the same corrections calls `proposeRule({ status: "draft" })`, `getAcceptedRules()` returns nothing, and live ingestion stops applying the approved rule. I reproduced this by consolidating, accepting `rule_tag_frontend`, then consolidating again. Please preserve existing accepted/rejected status when re-proposing, or only insert draft when the rule is new.

pranav-growthx (changes requested):
changes-requested: accepted draft rules are reset back to draft on the next consolidation run.

src/memory/sqlite.ts
pranav-growthx: Fixed in the latest push. `proposeRule` no longer updates `candidate_rule.status` on conflict, so accepted/rejected rules keep their reviewed state when consolidation re-proposes the same rule ID. Added a regression test that consolidates, accepts `rule_tag_frontend`, consolidates again, and verifies it remains accepted.

Verification:
- `bun test src/memory/sqlite.test.ts src/memory/cli.test.ts` -> 26 pass
- `bun run typecheck` -> pass
- `bun test` -> 673 pass, 0 fail

PranavBakre (commented):
approved: prior blocker is fixed; accepted rules stay accepted after re-consolidation.\n\nVerification: bun run typecheck; bun test src/memory/sqlite.test.ts src/memory/cli.test.ts src/memory/ingestion.test.ts; manual accepted-rule regression repro.

PR #85, opened 2026-06-03
Add worktree prune workflow

PranavBakre:
## Summary
- add a scheduled/manual `worktree-prune` workflow
- prunes only worktrees whose HEAD is already merged into the default branch
- treats `.DS_Store`, `.claude/**`, and `.codex/**` as ignorable local artifacts
- migrates useful `learnings.md` content into the primary checkout before pruning, leaving migrated learnings uncommitted for operator review

## Verification
- `bun test src/workflows`
- registry load check: 4 workflows, 0 errors

### DIFF
--- a/workflows/worktree-prune.workflow.md
+++ b/workflows/worktree-prune.workflow.md
@@ -32,8 +32,8 @@ Use the runtime context as the source of truth for repositories and their absolu
 
 1. Establish the protected primary checkout and default branch.
    - Use the repo path from runtime context as the primary checkout. Never delete or move this directory.
-   - Fetch the configured default base from the remote with pruning. Prefer `repo.defaultBase`, then `main`, then `master`.
-   - Resolve the base ref to the freshest available ref, preferring `origin/<base>`, then the local branch.
+   - Normalize `repo.defaultBase` before fetching or resolving it. If it starts with `origin/`, treat that full value as the preferred base ref and fetch only its branch part from the remote, e.g. `origin/main` means fetch `main`. Otherwise prefer `origin/<base>`, then the local `<base>` branch.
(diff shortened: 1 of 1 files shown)


workflows/worktree-prune.workflow.md
```diff
@@ -0,0 +1,73 @@
...
+---
+
+Prune stale local git worktrees across the configured Junior repos.
+
+Use the runtime context as the source of truth for repositories and their absolute paths. For each repo with a usable local git checkout:
+
+1. Establish the protected primary checkout and default branch.
+   - Use the repo path from runtime context as the primary checkout. Never delete or move this directory.
+   - Fetch the configured default base from the remote with pruning. Prefer `repo.defaultBase`, then `main`, then `master`.
+   - Resolve the base ref to the freshest available ref, preferring `origin/<base>`, then the local branch.
```
PranavBakre: warning: `repo.defaultBase` is not consistently a bare branch name in this repo. Several code paths/tests use values like `origin/main`, and the workflow runtime passes that string through as-is. With the current wording, a runner can reasonably try to fetch `origin/main` from the remote or resolve `origin/origin/main`, then skip/fail cleanup even though the repo is configured correctly. I’d make the prompt normalize `repo.defaultBase` first: if it starts with `origin/`, use it directly as the base ref and fetch its branch part (`main`); otherwise resolve `origin/<base>` then `<base>`.

PranavBakre (commented):
Reviewed the workflow definition. One warning inline; tests pass locally.

PranavBakre (commented):
approved from review pass: the base-ref warning is fixed; workflow schema loads cleanly and `bun test src/workflows` passes. GitHub will not let this account submit an APPROVE review on its own PR.

PR #86, opened 2026-06-03
Route prose through default agent

PranavBakre:
## Summary
- remove memory-based worker auto-routing for ordinary prose
- keep explicit !agent directives and the hard PR-link review route intact
- add regressions for onboarding-shaped prose, ordinary prose, and mention-only pings falling through to default

## Verification
- bun test src/support/router.test.ts
- bun run typecheck

### DIFF
--- a/src/support/router.ts
+++ b/src/support/router.ts
@@ -4,7 +4,6 @@ import type { SessionManager } from "../session/manager.ts";
 import type { SessionStore } from "../session/store/interface.ts";
 import type { DevServerQueue } from "../lifecycle/dev-server-queue.ts";
 import type { RepoConfig } from "../config.ts";
-import type { MemoryStore } from "../memory/store.ts";
 import {
   agentForUsername,
   isOrchestratorAgent,
@@ -115,7 +114,6 @@ export class AgentDispatcher {
       sessionStore?: SessionStore;
       slackClient?: WebClient;
       repos?: RepoConfig[];
-      memoryStore?: MemoryStore;
     } = {},
   ) {
     this.manager = manager;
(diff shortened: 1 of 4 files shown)


PranavBakre (commented):
warning: docs stale after removing memory-based routing

src/support/router.ts
```diff
@@ -195,15 +193,6 @@ export class AgentDispatcher {
         return;
       }
 
-      const memoryAgent = await this.memorySuggestedAgent(event);
-      if (memoryAgent && targetAgent !== "lead") {
```
PranavBakre, line 199: warning: Removing the memory-routing dispatch path leaves `docs/code_index/associative-memory.md` stale: it still lists `AgentDispatcher.memorySuggestedAgent` and a `memory-suggested routing` data flow. This repo is docs-first for code navigation, so future agents will read that index and believe ordinary prose still consults routing memory. Please update the associative-memory code index (and any feature doc if needed) with this behavior change.

src/support/router.ts
```diff
@@ -106,7 +106,6 @@ export class AgentDispatcher {
   private sessionStore: SessionStore | null;
   private slackClient: WebClient | null;
   private repos: RepoConfig[];
-  private memoryStore: MemoryStore | null;
```
PranavBakre, line 108: nit: `memoryStore` is now accepted by `AgentDispatcher` but ignored. I’d remove this option and the `src/index.ts` wiring in the same PR so tests don’t need `as never` and the constructor doesn’t advertise a routing dependency that no longer exists.

PranavBakre:
Addressed review in 32013b8:
- removed the stale AgentDispatcher memoryStore option and src/index.ts wiring
- updated docs/code_index/associative-memory.md to reflect that memory records routing decisions but no longer selects worker agents
- cleaned tests to assert prose falls through to default without fake memoryStore plumbing

Verification:
- bun test src/support/router.test.ts
- bun run typecheck

pranav-growthx (approved):
approved: prior warning/nit are fixed; no new issues found. Verification: bun test src/support/router.test.ts; bun run typecheck

PR #87, opened 2026-06-03
Internalize persistent-agent directives from Slack MCP posts

PranavBakre:
## Summary
- intercept pure persistent-agent directives passed through slack_send_message and dispatch them internally instead of posting directive text to Slack
- internalize allowed directive responses before duplicate Slack MCP post suppression
- add coverage for MCP Slack directive interception and duplicated Slack MCP directive posts

## Verification
- bun run typecheck
- bun test
- bun run typecheck
- bun test


pranav-growthx (approved):
approved — reviewed directive interception path and targeted tests; no blockers found. Verified typecheck and targeted tests.

PR #90, opened 2026-06-03
chore: reword dormant-thread side-conversation notice

PranavBakre:
## What

Rewords the notice Junior posts when it detects a side conversation (two humans talking) and steps out of a thread.

Before:
> Looks like a side conversation. I'll stay out — `@` me or `!listen` to bring me back.

After:
> Two people are interacting here, so I'll stop replying. @ me or use !listen to bring me back.

## Why

Clearer, less ambiguous phrasing about why the bot went quiet.

## Changes

- `src/session/manager.ts` — the posted message string
- `src/session/manager.test.ts` — assertion updated to exact-match the new copy
- `docs/features/thread-commands.md` — doc example updated

Typecheck green; all 75 session-manager tests pass.


PranavBakre:
It's already present

PR #93, opened 2026-06-07
fix: label self-mentions distinctly in resolved Slack text

PranavBakre:
## What

`resolveSlackMentions` now takes an optional `selfUserId`. When a Slack message tags the bot alongside other users, the bot's own mention renders as `Junior (you <@id>)` and other users render as `User(name <@id>)` — so the agent can tell a mention addressed to *it* apart from mentions of other people in the same message.

The prompt preamble also spells out explicitly that a `<@botId>` tag is addressed to Junior even when the message tags other people too.

## Why

Previously every mention resolved to the same `@name (<@id>)` shape, so when a message tagged Junior plus another person, the agent had no signal distinguishing "you were asked" from "someone else was mentioned."

## Changes

- `src/slack/thread-context.ts` — `resolveSlackMentions` gains `selfUserId`; distinct self vs. other rendering; preamble line clarifying self-addressed mentions
- `src/session/manager.ts` — threads `botUserId` through to `resolveSlackMentions`
- `src/slack/thread-context.test.ts` — coverage for self-vs-other labeling

## Verification

- `bun test src/slack/thread-context.test.ts` → 10 pass
- `bun run typecheck` → clean

### DIFF
--- a/src/session/manager.ts
+++ b/src/session/manager.ts
@@ -1054,7 +1054,11 @@ export class SessionManager {
       if (this.slackApp && latestTs) {
         const isFirstTurn = !runSession.sessionId;
         const needsThreadCatchup = !!session.needsThreadCatchup;
-        const readablePrompt = await resolveSlackMentions(this.slackApp, prompt);
+        const readablePrompt = await resolveSlackMentions(
+          this.slackApp,
+          prompt,
+          this.botUserId,
+        );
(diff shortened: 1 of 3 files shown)


PranavBakre:
Duplicate of #92 — same commit (aa5cf86e). Closing in favor of the original.

PR #96, opened 2026-06-09
feat(slack): add agent action buttons

PranavBakre:
## Summary
- add a structured <junior-actions> contract for agents to attach Slack buttons to final responses
- persist action tokens in SQLite and disable stale buttons on click or same-agent follow-up
- wire dispatch_agent and cleanup_worktree actions, including safe worktree cleanup and auto-cleanup after review approval
- update the private review agent submodule to emit Re-review, Make fix, and Cleanup worktree actions

## Testing
- bun run typecheck
- bun test

## Notes
- agents-org branch pushed: GrowthX-Club/junior-private-agents:add-review-action-buttons
- parent submodule pointer references agents-org commit 7f6bdaa

### DIFF
--- a/src/slack/responder.ts
+++ b/src/slack/responder.ts
@@ -3,6 +3,17 @@ import type { AgentIdentity } from "../session/types.ts";
 import { splitResponse } from "./formatting.ts";
 import { log } from "../logger.ts";
 
+export interface SlackMessageButton {
+  token: string;
+  label: string;
+  style?: "primary" | "danger";
+}
+
+export interface PostedSlackMessage {
+  ts: string;
+  text: string;
+}
+
 interface StatusEntry {
   messageTs: string;
   lastUpdateTime: number;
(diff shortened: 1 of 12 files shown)


src/slack/responder.ts
```diff
@@ -13,6 +24,29 @@ function preview(text: string, n = 80): string {
   return flat.length > n ? `${flat.slice(0, n)}…` : flat;
 }
 
+function buildActionBlocks(text: string, buttons: SlackMessageButton[]): Array<Record<string, unknown>> {
+  return [
```
pranav-growthx, line 30: blocker: Slack section block text is capped at 3,000 chars, but `splitResponse()` still allows the first chunk to be up to 4,000 chars. Any final agent response between 3,001-4,000 chars with buttons will send `blocks` here and Slack will reject the whole message with `invalid_blocks`, so the verdict and the action tokens never land. Please split/cap the block-bearing chunk at the Block Kit text limit, or fall back to plain text/no buttons for oversized first chunks.

pranav-growthx (changes requested):
Changes requested: found a blocker in the action-button render path. Long final responses with buttons can exceed Slack section block limits and fail to post entirely. See inline comment.

pranav-growthx (commented):
review: changes-requested — action buttons can fail to post for long verdicts because Block Kit section text has a tighter limit than `splitResponse()`.

PranavBakre:
Addressed the cleanup safety blocker in `a6dd028`.

Fixes:
- cleanup now checks committed-but-unpushed work before removing a worktree
- uses `@{u}..HEAD` when an upstream exists
- falls back to `repo.defaultBase..HEAD` when there is no upstream, which covers Junior-created `slack/<thread>` branches
- refuses cleanup before `git worktree remove --force` / `git branch -D` when local-only commits exist
- added worktree integration coverage for fresh clean worktrees and committed local-only work

Verification:
- `bun run typecheck`
- `bun test` (694 pass, 0 fail)

pranav-growthx (approved):
re-review: approved. The previous Block Kit limit blocker is fixed: button-bearing chunks are now split at 3,000 chars, with coverage in `responder.test.ts`. Focused tests and typecheck pass.

PR #97, opened 2026-06-10
fix: prevent real runner launches in tests

PranavBakre:
## Summary
- add a test-runtime fail-closed guard so SessionManager refuses to use the real runner unless explicitly allowed
- route SessionManager tests through a helper that always injects the mocked spawn runner
- update admin/clear/tmux/auto-trigger test setup to avoid accidental real Claude/OpenCode launches

## Verification
- bun test src/session/manager.test.ts
- bun run typecheck
- bun test
- git diff --check -- src/session/manager.ts src/session/manager.test.ts

### DIFF
--- a/src/session/manager.test.ts
+++ b/src/session/manager.test.ts
@@ -84,6 +84,19 @@ import { InMemorySessionStore } from "./store/memory.ts";
 import type { DriverMap } from "../claude/factory.ts";
 import type { ClaudeDriver } from "../claude/driver.ts";
 
+function createTestManager(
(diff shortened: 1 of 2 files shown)


PranavBakre:
review: approved from code/test inspection, but GitHub would not accept an approval because this gh session is authenticated as the PR owner.\n\nReviewed head SHA: 7373cfdc9bdcfd394d33b1f2551bd8203b83ea2f\n\nNo blockers or warnings found.\n\nVerified:\n- bun test src/session/manager.test.ts\n- bun run typecheck\n- bun test

PR #98, opened 2026-06-11
fix(slack): render action buttons from MCP posts

PranavBakre:
## Summary
- parse `<junior-actions>` in `slack_send_message` MCP posts
- render and store Slack action button tokens from MCP-posted messages
- split long action-bearing Slack posts so button blocks stay under the 3,000-character section limit
- use unique Slack `action_id`s per button while routing them to one handler
- route `review:make-fix` to `default` outside support channels and `thinker` only in support/bug channels
- update `agents-org` so approved reviews show `Merge via gxt-admin` + cleanup, and make-fix defaults to Junior outside bug channels

## Verification
- `bun run typecheck`
- `bun test src/slack/action-buttons.test.ts src/slack/responder.test.ts src/slack/action-store.test.ts src/slack/formatting.test.ts src/mcp/slack-server.test.ts`

### DIFF
--- a/src/mcp/slack-server.ts
+++ b/src/mcp/slack-server.ts
@@ -29,6 +29,10 @@ import {
 import { parsePureAgentDirectiveResponse } from "../support/directives.ts";
 import { parseSlackMcpRunContext, type SlackMcpRunContext } from "./context.ts";
 import { handleMongoMcpRequest } from "./mongodb-proxy.ts";
+import { prepareSlackResponseWithActions } from "../slack/formatting.ts";
+import { buildActionBlocks, splitActionMessageText } from "../slack/responder.ts";
+import type { SlackActionStore } from "../slack/action-store.ts";
+import { disableAgentActionMessages } from "../slack/action-buttons.ts";
 
 const MCP_PORT = Number(process.env.MCP_PORT ?? "3456");
 const FALLBACK_AGENTS_DIR = ".claude/agents";
@@ -39,6 +43,7 @@ let slack: WebClient;
 let sessionStore: SessionStore | undefined;
 let worktreeManager: WorktreeManager | undefined;
 let sessionManager: SessionManager | undefined;
+let slackActionStore: SlackActionStore | undefined;
 
 function registerTools(server: McpServer, runContext: SlackMcpRunContext | null = null) {
   server.registerTool(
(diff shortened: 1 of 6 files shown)


src/mcp/slack-server.ts
```diff
@@ -69,18 +74,63 @@ function registerTools(server: McpServer, runContext: SlackMcpRunContext | null
...
+      }
+
       const resolvedIconUrl = icon_url ?? image_url;
       const identity = {
         ...(username ? { username } : {}),
         ...(icon_emoji ? { icon_emoji } : {}),
         ...(resolvedIconUrl && !icon_emoji ? { icon_url: resolvedIconUrl } : {}),
       };
+      const sourceAgent = runContext?.agent ?? "mcp";
+      if (slackActionStore && thread_ts && prepared.actions.length > 0) {
```
pranav-growthx, line 91: warning: This only disables the prior buttons when the new MCP message also has actions. The action-button contract says buttons expire when the same source agent posts a newer response; with this guard, a later `slack_send_message` from the same agent without a `<junior-actions>` block leaves the old buttons active/clickable. This is especially easy to hit if an agent sends a final fallback/status via MCP. Please disable same-agent thread actions for every visible MCP post in the thread, not only posts that include replacement actions.

src/mcp/slack-server.ts
```diff
@@ -69,18 +74,63 @@ function registerTools(server: McpServer, runContext: SlackMcpRunContext | null
...
+        );
+      }
+      const buttonTokens = slackActionStore
+        ? prepared.actions.map((action) => ({
+            action,
+            token: crypto.randomUUID(),
+          }))
+        : [];
+      const blocks = buttonTokens.length > 0
+        ? buildActionBlocks(
```
pranav-growthx, line 109: warning: `buildActionBlocks` puts the full message into a Slack section block, whose `mrkdwn` text limit is 3000 chars. `SlackResponder.postResponse` handles this by splitting to a 3000-char first chunk when buttons are present, but the MCP path bypasses that logic, so a longer MCP post with actions will be rejected by Slack and no message/buttons will appear. Reuse the responder splitting behavior here or cap/split the block-bearing text before calling `chat.postMessage`.

pranav-growthx (changes requested):
Found two edge cases in the MCP action-button path that can leave stale buttons active or make longer action messages fail to post.

src/mcp/slack-server.ts
pranav-growthx: Still unresolved on head `88bfe3a`: this guard still only disables prior buttons when the new MCP post also includes replacement actions. A visible same-agent MCP post without `<junior-actions>` will still leave the previous buttons active.

src/mcp/slack-server.ts
pranav-growthx: Still unresolved on head `88bfe3a`: the MCP path still passes the full `prepared.text` into `buildActionBlocks`, so action-bearing posts over Slack section limits can still fail instead of splitting/capping like `SlackResponder.postResponse` does.

pranav-growthx (changes requested):
Re-review on 88bfe3a: the two prior warnings are still unresolved. I replied on both existing threads; no new issues found in the added action_id pattern change.

pranav-growthx (approved):
Approved. Reviewed the single commit and final diff; no blockers or warnings found. Verified with bun test src/slack/action-buttons.test.ts src/slack/responder.test.ts src/slack/formatting.test.ts and bun run typecheck.

PR #101, opened 2026-06-17
feat(claude): bring claude -p runner to parity with codex-app-server

PranavBakre:
## Why

Junior is switching its default runner back to Claude from `codex-app-server`. The Claude `claude -p` adapter was deliberately minimal — it ignored per-agent permission intent, lacked the MCP/model/isolation features the codex adapter had grown, and had no human-gated approval path. This brings the **mechanism** to parity.

## What

- **Permissions** — `src/claude/policy.ts` maps `session.agentPermissions.intent` to Claude flags: read-only/no-tools → `--permission-mode plan` + `--disallowedTools`; human-gated → `plan` (or approval, below); utility → config; **normal/unset → `bypassPermissions`**.
- **Model resolution** — `src/claude/model.ts`: `model.claude` override > Claude alias > GPT map (`gpt-5.5 → opus`) > config default. Mirror of `codexCompatibleModel`.
- **Slack approval round-trip** — human-gated agents get `--permission-prompt-tool mcp__slack-bot__request_permission`; the tool posts Allow/Deny buttons (reusing `SlackActionStore`) and blocks on an in-process registry (`src/mcp/approval.ts`), default-denying after `CLAUDE_APPROVAL_TIMEOUT_MS` (240s). Only wired when the slack-bot MCP is actually configured for the run.
- **Isolation** — `--strict-mcp-config` + `--setting-sources project` keep developer-global `~/.claude` config/MCP out.
- **MCP / prompt / cost** — mixpanel MCP parity; Junior baseline `--append-system-prompt`; `total_cost_usd`/`usage` on the done event + `--max-budget-usd`.

All knobs config-gated, default-on except the budget cap.

## ⚠️ Scope / honest limitations (read before flipping the runner)

- **Claude has no OS-level sandbox.** Unlike codex's Seatbelt (`workspace-write`, network off), Claude can only confine via tool allow/deny + the working directory. So **writer agents run broad** on the Claude path — that's inherent, not a gap this PR can close.
- **The lockdown only activates for agents that declare `permissions.intent`.** Today only `agents-org/onboarding.md` does; the main agents (review/thinker/pm/architect/build/…) are intent-unset and therefore hit `bypassPermissions`. **Declaring per-agent intents is a deliberate follow-up**, not part of this PR. Until then, null-intent agents are *more* permissive on Claude than on codex — factor this into the runner-flip decision.
- **Unverified assumption (#2):** the approval path relies on headless `--permission-mode default` routing un-allowed tools to `--permission-prompt-tool`. This needs **one live human-gated turn** (tool call → Slack buttons → Allow → tool runs) confirmed before the default runner is flipped to Claude.

## Verification

- `bun run typecheck` clean; `bun test` → **755 pass / 0 fail** (two consecutive clean passes).
- Reviewed by the structured reviewer agent; should-fix items addressed in `f520af2` (approval gated on MCP presence; doc nits). #4 (tmux permission-mode) was a false positive — tmux already inherited `--permission-mode` pre-PR.

## Deferred (from the audit)

Graceful mid-turn interrupt (needs Agent SDK), `--json-schema` structured output, `--include-partial-messages` streaming, and the per-agent intent declarations above.

### DIFF
--- a/src/agents/loader.test.ts
+++ b/src/agents/loader.test.ts
@@ -278,6 +278,50 @@ body`;
     }
   });
 
+  describe("modelClaude frontmatter field", () => {
+    it("parses model.claude frontmatter into modelClaude", async () => {
+      const tmpPath = path.join(import.meta.dir, "__test_model_claude.md");
+      const content = `---
+name: claude-override
+description: Test
+model: gpt-5.5
+model.claude: sonnet
+---
+
+body`;
+      await Bun.write(tmpPath, content);
+
+      try {
+        const def = await loadAgentDefinition(tmpPath);
+        expect(def).not.toBeNull();
+        expect(def!.model).toBe("gpt-5.5");
+        expect(def!.modelClaude).toBe("sonnet");
+      } finally {
+        await fs.unlink(tmpPath).catch(() => {});
+      }
+    });
+
+    it("modelClaude is null when model.claude is not declared", async () => {
+      const tmpPath = path.join(import.meta.dir, "__test_model_claude_absent.md");
+      const content = `---
+name: no-claude-override
+description: Test
+model: gpt-5.5
+---
+
+body`;
+      await Bun.write(tmpPath, content);
+
+      try {
+        const def = await loadAgentDefinition(tmpPath);
+        expect(def).not.toBeNull();
+        expect(def!.modelClaude).toBeNull();
+      } finally {
+        await fs.unlink(tmpPath).catch(() => {});
+      }
+    });
+  });
+
   it("handles incomplete frontmatter (missing closing ---)", async () => {
     const tmpPath = path.join(import.meta.dir, "__test_incomplete.md");
     const content = `---
--- a/src/agents/loader.ts
+++ b/src/agents/loader.ts
@@ -53,6 +53,11 @@ export interface AgentDefinition {
   description: string;
   tools: string | null;
   model: string | null;
+  /**
+   * Explicit Claude-runner model override, from `model.claude` frontmatter.
+   * Wins over the GPT→Claude map.
+   */
+  modelClaude: string | null;
   common: string[];
   prompt: string;
   context: AgentContextProfile;
@@ -84,6 +89,7 @@ export async function loadAgentDefinition(
     description: frontmatter["description"] ?? "",
     tools: frontmatter["tools"] ?? null,
     model: frontmatter["model"] ?? null,
+    modelClaude: frontmatter["model.claude"] ?? null,
     common: readCommonProfile(frontmatter),
     prompt: body.trim(),
     context: readContextProfile(frontmatter),
--- a/src/claude/args.test.ts
+++ b/src/claude/args.test.ts
@@ -196,9 +196,9 @@ describe("buildClaudeArgs", () => {
     expect(args).toContain("project");
   });
 
-  it("wires the Slack approval round-trip for human-gated when approval enabled", () => {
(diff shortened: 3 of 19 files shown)


PranavBakre:
**Structured review complete — changes requested, now addressed.**

Resolved in `f520af2`:
- **#3** (real, latent): `--permission-prompt-tool` was added whenever approval was active, but `shouldUseClaudeMcpConfig` drops MCP for `session.cwd` runs → the approval tool could be advertised without the slack-bot MCP. Now `approvalActive` requires `mcpConfigPath`, falling back to safe `plan` mode otherwise.
- **#5 / #7** (nits): documented that `plan` mode (not the deny list) is the real read-only gate; documented the approval-timeout < idle-timeout invariant.

Acknowledged, not changed (by decision):
- **#1**: null-intent agents run `bypassPermissions`. Claude has no OS sandbox so writer agents run broad regardless; declaring per-agent `read-only` intents is a follow-up. PR description corrected to drop the overclaim.
- **#4**: false positive — tmux already inherited `--permission-mode` before this PR.

Open pre-flip gate:
- **#2**: headless `--permission-mode default` + `--permission-prompt-tool` routing must be validated with one live human-gated turn before the default runner is switched to Claude.

pranav-growthx (approved):
Second independent review round — two consecutive clean passes, approving.

**Verification of prior findings (all confirmed):**
- **#3 (approval tool advertised without MCP):** genuinely fixed in `f520af2`, not papered over. Traced every human-gated path — `approvalActive` now requires `mcpConfigPath != null`, so a `session.cwd` human-gated run falls back to the safe `plan` base and never advertises `--permission-prompt-tool`. No path leaves the approver half-wired, none falls through to `bypassPermissions`, and the fix changed permission-mode for no other run.
- **#1:** PR body + code comments now honest about null-intent → `bypassPermissions`; no false confinement claim anywhere.
- **#2:** headless `--permission-mode default` routing is correctly flagged as an open pre-flip gate, not silently relied upon.
- **#4:** confirmed false-positive (tmux `--permission-mode` inheritance is pre-PR; `--permission-prompt-tool` is now explicitly stripped on the tmux path).
- **#7 invariant:** verified — idle timeout default 300000 > approval default 240000; idle timer resets on every runner event.

**Independent pass — new checks, all clear:**
- All Claude CLI flags verified to exist against `claude --help` (`--max-budget-usd`, `--setting-sources`, `--strict-mcp-config`, `--permission-prompt-tool`, etc.) — no invented flags.
- `modelClaude` persists correctly (sqlite JSON-blob serialization; `normalizeSession` is a passthrough); dotted `model.claude` frontmatter key parsed and tested.
- Approval round-trip wiring sound: in-process HTTP MCP server + in-memory pending Map bridge; per-request `runContext` via URL; double-click race fully closed (atomic claim + idempotent resolve + button disabling); timer leak bounded and unref'd.
- Config defaults default-on are safe to ship — live default runner is `opencode`, so this path is inert until the gated flip to `claude`.

**Nits only (non-blocking):**
- `src/mcp/approval.ts:38` — "INVARIANT" comment overstates certainty about which stream event resets the idle timer (the permission-prompt-tool's stream visibility is undocumented per Claude Code docs); safety margin holds regardless via the original tool_use reset + 240<300 default-deny.
- `src/mcp/slack-server.ts:170` — on allow, echoing `...rest` as `updatedInput` could carry extraneous fields if Claude sends loose top-level input; #2's live turn will surface this.
- `src/slack/action-buttons.ts:40` — disable/remove run outside the try/catch (degraded-but-safe default-deny on Slack API throw; pre-existing across all action types).

Fixes are sound and only nits remain. Approved.

PR #118, opened 2026-07-09
Rewrite agent definitions around three working modes + continuous memory

PranavBakre:
Rewrites all agent definitions around Pranav's actual working styles (mined from ~30 interactive junior sessions, 22 GX product-work sessions, and Junior's memory DB), makes memory usage continuous, **merges default+lead+thinker into one orchestrator**, and deduplicates the composed prompt tiers.

## 1. Working-style rewrite (three modes)
- **Build features** (build, frontend, architect, pm): interrogate raw-note specs, HTML-mockup gate, approval as a hard gate, two consecutive clean passes, baseline-matched verification, PR-body claims verified against the diff, no `as any`, explicit-path staging.
- **Resolve bugs**: evidence before hypothesis, root cause before patch, prod read-only, quantified investigation, honest outcomes.
- **Grunt work**: scope-test then autonomy, keep=local deliverable, artifact inspection before delivery.

## 2. Continuous memory
Every agent gets `mcp__slack-bot__memory_recall` in tools with recall checkpoints woven into its workflow (task start with entity_refs, before dispatch, before merges/destructive steps, on unfamiliar entities, when surprised); default and review also get `memory_add` for atomic claims on correction. core.md carries the habit (single home) within its 2500-char budget; the OpenCode fallback prompt is synced.

## 3. 3-way orchestrator merge
One Junior orchestrator (`default`) routes, runs the bug pipeline, and hypothesizes/scopes itself; **fix implementation is dispatched to build/frontend via Task** — the orchestrator never edits product code. `lead`/`thinker` are router aliases of `default`; the pipeline playbook lives in `common/bug-pipeline.md`, appended only for lead/thinker sessions together with merge-workflow + runtime-environment. Survives unchanged: the two-turn human gate (Message 1 → stop → human reply → Phase 2), reproducer/review contracts, round caps, silence allow-list, byte-exact terminal merge message. `!thinker` retires; the pipeline guard accepts a Phase-1 Message 1 as a valid advance (new tests). Prod `CHANNEL_DEFAULTS` keeps working — `"lead"` stays the support-channel marker; no config migration.

## 4. Prompt dedup (single-home policy)
Each rule is stated once per composed prompt: memory habit in core.md, git guardrails in generic merge-workflow, builder verification in building-philosophy; org overlays keep only org-specific facts. Composed prompts (re-sent every runner turn): default 22.1K→18.9K chars, review 42.7K→38.9K, pipeline session 62.2K→56.1K; on review, memory guidance 8×→2×, force-push rule 3×→1×, baseRefName 5×→2×. Remaining bulk is single-homed contract content (pipeline playbook, worker catalogue, review methodology), not repetition. Corrected merge-flow wording throughout: primary PR targets main, reviewed against main; secondary dev PR raised post-review, merged first; main human-gated.

## Verification
Two consecutive clean passes at each stage: `bun run typecheck` green; `bun test` 837 pass / 0 fail (includes the prompt-lint suite and new router-alias + pipeline-guard tests). Independent second-pass review agents ran after both the rewrite and the merge; all their blockers/should-fixes are fixed in-branch (incl. a real regression where aliased pipeline sessions lost merge-workflow/runtime-environment).

## Companion PRs (merge those first, then bump submodule pointers separately)
- Identity: https://github.com/PranavBakre/junior-identity/pull/1
- Overlay: https://github.com/GrowthX-Club/junior-private-agents/pull/14

## Roadmap (not in this branch)
- Peer-to-peer agent messaging with respond-or-decline (swarm), stuck-alerts over hop-kills.

https://claude.ai/code/session_01DAVCPHF6NUooDJPAVchwTs


src/agents/router.ts
```diff
@@ -60,8 +78,24 @@ export class AgentRouter {
...
+    // Support-channel sessions (marker "lead", plus resumed pre-merge "thinker"
+    // sessions) run the bug pipeline. Append the pipeline preambles AFTER the
+    // agent's declared profile so they participate like any other common name:
+    // readProfileMarkdownFiles resolves them through the target-repo → public
+    // tiers, and the org overlay tier can supplement or override them by the
+    // same additive mechanics. merge-workflow and runtime-environment ride along
+    // because the pipeline's merge step and dev-server/bug-folder contracts live
+    // there — the lean default profile deliberately omits them for casual
+    // threads. Casual "default" sessions get none of this.
+    if (rawName === "lead" || rawName === "thinker") {
```
gxt-admin, line 94: 🟢 **Nit (non-blocking) — vestigial `thinker` alias composes a self-contradictory prompt.**

This branch appends `bug-pipeline.md` for a `thinker` session, and the playbook instructs the session to run Phase 2 and emit `!reproducer validate` / `!review`. But `thinker` is not in `ORCHESTRATOR_AGENTS`, so `dispatchableAgentsFor("thinker")` returns `[]` and `buildDispatchAllowBlock("thinker")` injects *"You may NOT emit `!<agent>` directives"* into the same composed prompt. The two blocks directly contradict each other.

In practice this is near-unreachable: no live path emits `!thinker` (the orchestrator's allow-list excludes it), and pre-merge `thinker` agent sessions have their pending messages dropped as stale on restart — so the branch essentially never fires. That's why it's a nit, not a blocker.

Suggest picking one of: (a) drop the `"thinker"` alias + this `rawName === "thinker"` arm entirely, since resumed thinker sessions can't reach it post-restart; or (b) if you genuinely want resumed thinker sessions to run the pipeline, add `"thinker"` to `ORCHESTRATOR_AGENTS` so the playbook it's handed is actually executable. Either keeps the code honest with the router comment and the `persistent-agents.md` claim.

gxt-admin (approved):
Approved — two consecutive clean passes, zero blockers, zero should-fixes.

Verified locally on the branch:
- `bun run typecheck` green.
- `bun test` 835 pass / 0 fail (incl. prompt-lint, new router-alias tests, new pipeline-guard tests).
- core.md is 2486 chars (< 2500) and byte-identical to `FALLBACK_JUNIOR_CORE_PROMPT`.
- Terminal merge message in `bug-pipeline.md` is byte-exact against the old `lead.md` string.
- Two-turn human gate (Message 1 → stop → human → Phase 2) preserved in `bug-pipeline.md`.
- Every agent's tools CSV carries `memory_recall`; `memory_add` only on `default` + `review`.
- Prod `CHANNEL_DEFAULTS` agentType `lead` still resolves (alias → `default.md` + pipeline preambles + orchestrator dispatch powers).
- `action-buttons` review:make-fix now routes to `lead` — the dispatch guard already whitelists `lead`, no throw.
- No dead references to the deleted `lead.md` / `thinker.md` files anywhere in `src`.

One 🟢 nit inline (non-blocking).

PR #121, opened 2026-07-11
feat(whatsapp): Hermes buildathon task tracker (Baileys ingestion + LLM extraction + Notion sync)

PranavBakre:
## Summary

- **src/whatsapp ingestion**: Baileys WebSocket socket for live message events from Hermes buildathon groups, with SQLite message store (append-only, group allowlist filter)
- **src/whatsapp/extraction sweep**: Periodic LLM extraction pipeline (claude -p runner) that infers tasks from new messages, maintains open tasklist state, detects completion via replies, with subprocess security lockdown (untrusted WhatsApp text → no-tools, neutral cwd, no MCP)
- **src/notion sync**: REST API (no MCP dependency) to sync extracted tasks to a Notion database on the Hermes page, with automatic mentor-links.csv → sub-page script

**Phases 0–2 implemented** (read-only, no WhatsApp send path; reply flow is a gated later phase).

## Prerequisites to go live

- `NOTION_TOKEN` in `.env` (internal integration, Hermes page shared)
- `WHATSAPP_ENABLED=true` to activate socket on startup
- One-time QR pairing: `bun run src/whatsapp/pair.ts` (scan from mobile app)

## Review

4 codex (gpt-5.6-sol) review rounds, 13 findings fixed and re-verified. 958 tests pass + typecheck clean twice.

## Follow-up

Apply the same subprocess lockdown audit to the memory-consolidation runner (processes Slack-derived content) in a separate branch.

https://claude.ai/code/session_01DAVCPHF6NUooDJPAVchwTs

### DIFF
--- a/scripts/mentor-links-to-notion.ts
+++ b/scripts/mentor-links-to-notion.ts
@@ -1,0 +1,281 @@
+/**
+ * One-shot: read hermes-scoring mentor-links.csv and create a "Mentor Links"
+ * child page (with a table) under the Hermes Notion page.
+ *
+ * Usage: bun run scripts/mentor-links-to-notion.ts
+ * Env: NOTION_TOKEN (required), HERMES_NOTION_PAGE_ID (optional)
+ */
+
+const NOTION_API_BASE = "https://api.notion.com/v1";
+const NOTION_VERSION = "2022-06-28";
+const CSV_PATH = "/Users/psbakre/Projects/hermes-scoring/mentor-links.csv";
+const DEFAULT_PAGE_ID = "39a3578dc3f08015b386cc6638029bed";
+const CHILD_PAGE_TITLE = "Mentor Links";
+
+interface MentorLinkRow {
+  city: string;
+  name: string;
+  role: string;
+  token: string;
+  link: string;
+}
+
+interface NotionRichText {
+  type: "text";
+  text: {
+    content: string;
+    link?: { url: string };
+  };
+}
+
+type NotionTableCell = NotionRichText[];
+
+interface NotionTableRowBlock {
+  type: "table_row";
+  table_row: {
+    cells: NotionTableCell[];
+  };
+}
+
+interface NotionTableBlock {
+  type: "table";
+  table: {
+    table_width: number;
+    has_column_header: boolean;
+    has_row_header?: boolean;
+    children: NotionTableRowBlock[];
+  };
+}
+
+interface NotionChildPageBlock {
+  object: "block";
+  id: string;
+  type: "child_page";
+  child_page: {
+    title: string;
+  };
+}
+
+interface NotionBlockListResponse {
+  object: "list";
+  results: Array<{
+    object: "block";
+    id: string;
+    type: string;
+    child_page?: { title: string };
+  }>;
+  next_cursor: string | null;
+  has_more: boolean;
+}
+
+interface NotionPageResponse {
+  object: "page";
+  id: string;
+  url?: string;
+}
+
+function plainCell(content: string): NotionTableCell {
+  return [{ type: "text", text: { content } }];
+}
+
+function linkCell(url: string): NotionTableCell {
+  return [{ type: "text", text: { content: url, link: { url } } }];
+}
+
+function tableRow(cells: NotionTableCell[]): NotionTableRowBlock {
+  return { type: "table_row", table_row: { cells } };
+}
+
+function headerRow(): NotionTableRowBlock {
+  return tableRow([
+    plainCell("City"),
+    plainCell("Name"),
+    plainCell("Role"),
+    plainCell("Token"),
+    plainCell("Link"),
+  ]);
+}
+
+function rowToTableRow(row: MentorLinkRow): NotionTableRowBlock {
+  return tableRow([
+    plainCell(row.city),
+    plainCell(row.name),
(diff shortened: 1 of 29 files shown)


src/whatsapp/extraction/runner.ts
```diff
@@ -0,0 +1,166 @@
...
+): ExtractionRunner {
+  const timeoutMs = opts.timeoutMs ?? DEFAULT_EXTRACTION_TIMEOUT_MS;
+  const model = sanitizeClaudeModel(opts.model ?? DEFAULT_EXTRACTION_MODEL);
+  const sandboxDir = opts.sandboxDir ?? DEFAULT_EXTRACTION_SANDBOX_DIR;
+
+  return async (prompt: string): Promise<string> => {
+    // Neutral, empty cwd (created on demand) so the subprocess doesn't inherit
+    // junior's CLAUDE.md / `.claude/` settings / `.mcp.json` project context —
+    // the run embeds untrusted group messages (see `buildExtractionArgs`).
+    mkdirSync(sandboxDir, { recursive: true });
```
PranavBakre, line 99: **warning** — the "neutral cwd → no project CLAUDE.md / `.claude/` / `.mcp.json` inherited" claim (doc comment above + line 97-98) isn't actually achieved. `index.ts:402` sets `sandboxDir` to `resolve(dbPath, '..', 'whatsapp-extraction-sandbox')` → `<repo>/data/whatsapp-extraction-sandbox`, which is **inside the junior repo**. Claude Code discovers `CLAUDE.md` and project `.claude/settings.json` by walking up from cwd to the project root, so junior's CLAUDE.md and project settings still load into the untrusted-content run. `--setting-sources` isn't passed either, so user-global `~/.claude` settings load too.

Contained **today** only because `--tools ""` and `--strict-mcp-config` both hold (untrusted content has nothing to act with). But the whole point of this file is untrusted-content hardening, and the primary stated control is a no-op — if junior ever adds a tool-enabling or auto-approve setting to `.claude/settings.json`, this run silently inherits it. Fix: put the sandbox **outside** the repo tree (e.g. `os.tmpdir()`), and/or pass a scoped `--setting-sources`. Also correct the comment to match what actually isolates the run.

src/whatsapp/extraction/runner.ts
```diff
@@ -0,0 +1,166 @@
...
+      const exitCode = await proc.exited;
+      if (timedOut) {
+        throw new Error(
+          `extraction runner: claude timed out after ${timeoutMs}ms`,
+        );
+      }
+      if (exitCode !== 0) {
+        let stderr = "";
+        try {
+          stderr = (await new Response(proc.stderr).text()).trim();
```
PranavBakre, line 131: **nit** — stderr is only drained *after* stdout fully reads and the process exits. If the child ever fills its ~64KB stderr pipe buffer it would block on the stderr write, stall stdout, and hang until the 5-min timeout SIGINT. Low likelihood here (json goes to stdout), but draining stdout+stderr concurrently (`Promise.all`) before `await proc.exited` removes the risk entirely.

src/notion/sync.ts
```diff
@@ -0,0 +1,371 @@
...
+
+function buildProperties(task: TaskForSync): PageProperties {
+  return {
+    Task: { title: [{ type: "text", text: { content: task.task } }] },
+    // Stable dedupe key — written on both create and update so an existing page
+    // can always be matched back to its SQLite task on a later sweep.
+    [TASK_ID_PROPERTY]: {
+      rich_text: [{ type: "text", text: { content: task.id } }],
+    },
+    Owner: { select: task.owner ? { name: task.owner } : null },
```
PranavBakre, line 310: **nit** — `Owner` is a Notion `select` set from the LLM-derived `task.owner`. Notion rejects `select` option names containing a comma, so an owner like `"John, Jane"` (or any name the model emits with a comma) would 429/400 and keep that row perpetually `notion_dirty=1`, retried and failing every sweep. Either strip commas when mapping owner, or make `Owner` a `rich_text` property. Priority/Status are safe (fixed enums).

scripts/mentor-links-to-notion.ts
```diff
@@ -0,0 +1,281 @@
...
+  }
+
+  const rows: MentorLinkRow[] = [];
+  for (let i = 1; i < lines.length; i++) {
+    const parts = lines[i]!.split(",");
+    if (parts.length < 5) {
+      throw new Error(`CSV line ${i + 1} has fewer than 5 columns: ${lines[i]!}`);
+    }
+    // Link may contain no commas; join remainder for safety if any.
+    const [city, name, role, token, ...linkParts] = parts;
```
PranavBakre, line 138: **nit** — naive `split(",")` means a comma in `Name`/`Role` shifts every column (only the trailing `link` is comma-tolerant via `linkParts.join`). Fine for the known 37-row one-shot, but worth a header-validated caveat. Also the CSV path is hardcoded to an absolute machine path (line 11) — acceptable for a run-once script, just noting it isn't portable.

PranavBakre (commented):
Review of #121 (WhatsApp Hermes tracker, Phases 0-2, read-only). Solid, thoroughly-tested, well-commented feature. Typecheck clean, 121/121 whatsapp/notion/config tests pass locally. No blockers. One warning on the extraction sandbox isolation + a few nits below.

PR #125, opened 2026-07-11
feat(memory): automatic pre-recall hook before runner turns

PranavBakre:
## Summary
- Adds a pre-recall hook that runs before every runner turn: a cheap LLM (haiku/deepseek/gpt-4o-mini) extracts 0-3 recall queries from the incoming Slack message, then calls `recallMemory()` and injects matching claims as a `<pre-recall>` block into the prompt
- Supports all three CLI runners (claude, opencode, codex) via `PRE_RECALL_RUNNER` env var; defaults to claude/haiku
- Solves the chicken-and-egg problem where agents skip `memory_recall` and waste tokens re-discovering knowledge that's already stored

## Changes
- **`src/memory/pre-recall.ts`** (new) — factory function, per-runner subprocess spawning, query extraction prompt, recall + dedup + formatting
- **`src/config.ts`** — `preRecall` section in memory config (`PRE_RECALL_ENABLED`, `PRE_RECALL_RUNNER`, `PRE_RECALL_MODEL`, `PRE_RECALL_TIMEOUT_MS`)
- **`src/session/manager.ts`** — captures raw message early, runs pre-recall, prepends results before runner spawn

## Config
| Env var | Default | Description |
|---------|---------|-------------|
| `PRE_RECALL_ENABLED` | `true` | Enable/disable the hook |
| `PRE_RECALL_RUNNER` | `claude` | Which CLI to use (claude/opencode/codex) |
| `PRE_RECALL_MODEL` | per-runner cheapest | Override model |
| `PRE_RECALL_TIMEOUT_MS` | `15000` | Timeout for the LLM extraction call |

## Test plan
- [ ] Verify typecheck passes (done locally)
- [ ] Test with `PRE_RECALL_ENABLED=true` — send a message mentioning a known memory topic, confirm `<pre-recall>` block appears in the prompt
- [ ] Test with `PRE_RECALL_ENABLED=false` — confirm no pre-recall block
- [ ] Test timeout handling — set `PRE_RECALL_TIMEOUT_MS=1` and confirm graceful fallback (no block, warning logged)
- [ ] Test each runner (claude, opencode, codex) extracts queries correctly

### DIFF
--- a/src/memory/pre-recall.ts
+++ b/src/memory/pre-recall.ts
@@ -1,0 +1,368 @@
+// Pre-recall hook: runs BEFORE the runner spawns to inject operational memory
+// into the prompt. A cheap one-shot LLM extracts 0-3 recall queries from the
+// raw Slack message, then each query hits recallMemory() from the MCP server.
+// Results are formatted as a <pre-recall> XML block prepended to the prompt.
+//
+// The LLM call is a CLI subprocess (CLAUDE.md rule 1), not an SDK call. Same
+// timeout + process-tree SIGINT pattern as the consolidation runner. The module
+// exports a factory function returning a closure (CLAUDE.md rule 14).
+
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { readFile, rm } from "node:fs/promises";
+
+import type { Config } from "../config.ts";
+import type { MemoryToolDeps, RecallMemoryResult } from "../mcp/slack-server.ts";
+import { recallMemory } from "../mcp/slack-server.ts";
+import { createMemoryStore } from "./factory.ts";
+import { createProfileStore } from "./profiles/index.ts";
+import { signalProcessTree } from "../lifecycle/process-tree.ts";
+import { sanitizeClaudeModel } from "./consolidation/runner.ts";
+import { createOpenCodeStreamParser, createOpenCodeEventMapper } from "../opencode/parser.ts";
+import { log as _log } from "../logger.ts";
+
+// ── Runner type (same as ConsolidationRunner) ────────────────────────────────
+export type PreRecallRunner = "claude" | "opencode" | "codex";
+
+// ── Pinned cheapest models per runner ────────────────────────────────────────
+const DEFAULT_CLAUDE_MODEL = "claude-haiku-4-5-20251001";
+const DEFAULT_OPENCODE_MODEL = "opencode-go/deepseek-v4-pro";
+const DEFAULT_CODEX_MODEL = "gpt-4o-mini";
+
+// ── Extraction system prompt ─────────────────────────────────────────────────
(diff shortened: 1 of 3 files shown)


src/memory/pre-recall.ts
```diff
@@ -0,0 +1,368 @@
...
+ * Extract the final assistant text from OpenCode's `--format json` NDJSON
+ * stdout using the production stream parser.
+ */
+function extractOpenCodeAssistantText(stdout: string): string {
+  const parser = createOpenCodeStreamParser();
+  const mapper = createOpenCodeEventMapper();
+  for (const event of parser.feed(stdout)) mapper.map(event);
+  for (const event of parser.flush()) mapper.map(event);
+  return mapper.response || stdout.trim();
+}
```
PranavBakre, line 303: Should-fix (Code Reuse): This function is byte-for-byte identical to `extractOpenCodeAssistantText` in `src/memory/consolidation/runner.ts:283`, which is already exported. This file already imports `sanitizeClaudeModel` from that module, so the dependency path exists.

Import it instead of duplicating:
```ts
import { sanitizeClaudeModel, extractOpenCodeAssistantText } from "./consolidation/runner.ts";
```

Similarly, `extractClaudeAssistantText` (line 232) is functionally identical to the consolidation runner's `extractAssistantText` (line 251) -- that one isn't exported today, but could be shared if it were.

src/memory/pre-recall.ts
```diff
@@ -0,0 +1,368 @@
...
+
+    const store = createMemoryStore(config.memory.sqlitePath);
+    const { createEmbeddingProvider } = await import("./embedding/factory.ts");
+    const provider = createEmbeddingProvider(
+      config.memory.embedProvider ?? "local",
+    );
+    const profileStore = createProfileStore();
+    deps = { store, provider, profileStore };
+    return deps;
+  }
```
PranavBakre, line 81: Should-fix (Resource waste): `getDeps()` lazily creates its own `MemoryStore`, `EmbeddingProvider`, and `ProfileStore`. The MCP server (`src/mcp/slack-server.ts:65`) documents the cost: the embedding provider lazy-loads a ~500MB ONNX model on its first embed (~20s).

When pre-recall runs its first `recallMemory()` call, it loads the ONNX model into a new `LocalEmbeddingProvider`. If the agent then calls the `memory_recall` MCP tool during the same turn, the MCP server loads a second copy. Two instances in the same process: ~1GB for identical data.

Fix: Accept `MemoryToolDeps` (or its components) as a parameter to `createPreRecall`, and have the caller provide shared instances. This also eliminates the TOCTOU race in this lazy init (two concurrent threads can both see `deps === null` before the `await import()` resolves, creating duplicate instances).

```ts
// Suggested signature:
export function createPreRecall(config: Config, deps: MemoryToolDeps): PreRecallFn
```

PranavBakre (commented):
Two should-fix items found. See inline comments for details.

PR #133, opened 2026-07-21
Isolate pipeline agents in managed worktrees

PranavBakre:
## Summary
- resolve every durable pipeline repo ref before dispatch
- provision and persist one Junior-managed worktree per repo/thread
- select the invocation cwd by review and durable workstream affinity
- coalesce concurrent setup and fail closed with a durable, recoverable escalation
- keep recovery continuations in the same managed worktree
- allow repo-less planner/orchestrator assignments while keeping human recovery turns usable
- synchronize public agent prompts and worktree documentation

## Why
Pipeline agents must not run in the developer main checkout or inherit a stale utility cwd. Multi-repo runs also need every referenced repo provisioned before fan-out.

## Review fixes
- prevent repo-less and unknown-repo failures from wedging the thread
- contain pipeline-store errors during pre-spawn settlement
- keep incidental PR URLs from rerouting non-review agents
- diagnose ambiguous workstream affinity and dedupe unresolved refs
- clear stale utility cwd for repo-less orchestration

## Validation
Two consecutive clean passes on the final head:
- `git diff --check`
- `bun run typecheck`
- `bun test` — 1,290 passed, 2 expected embedding skips, 0 failed, 3,938 assertions

## Runtime note
The currently running Junior process was not restarted.

### DIFF
--- a/src/session/manager.ts
+++ b/src/session/manager.ts
@@ -1,5 +1,5 @@
 import type { App } from "@slack/bolt";
-import type { Config } from "../config.ts";
+import type { Config, RepoConfig } from "../config.ts";
 import type { RunnerEvent, RunnerProvider, SpawnHandle, SpawnResult, SpawnRunnerFn } from "../runners/types.ts";
 import type {
   SlackMessageEvent,
@@ -14,6 +14,7 @@ import type {
   ThreadSession,
 } from "./types.ts";
 import type { AgentRouter } from "../agents/router.ts";
+import { resolveAgentManifest } from "../agents/registry.ts";
 import type { WorktreeManager } from "../worktree/manager.ts";
 import type { PipelineStore } from "../pipelines/store/interface.ts";
 import {
@@ -1339,28 +1340,24 @@ export class SessionManager {
           )
         : undefined;
 
-      if (activePipelineRun) {
+      if (activePipelineRun && pipelineInvocation) {
         // Durable repo refs, not a developer checkout or stale session cwd,
         // define the pipeline workspace. Provision every referenced repo so
-        // fan-out agents can collaborate through the injected path map.
-        if (!this.worktreeManager) {
-          throw new Error(
-            `Pipeline ${activePipelineRun.id.slice(0, 8)} cannot run: worktree manager is unavailable.`,
-          );
-        }
-
+        // fan-out agents can collaborate through the injected path map. Plain
+        // human turns on a needs-human run do not own an assignment, so they
+        // must remain usable to diagnose or reset an unrecoverable run.
         const resolution = resolvePipelineRepos(
           this.config.repos,
           activePipelineRun.repoRefs,
         );
         if (resolution.unresolvedRefs.length > 0) {
           throw new Error(
-            `Pipeline ${activePipelineRun.id.slice(0, 8)} cannot run: repository refs are not uniquely configured: ${resolution.unresolvedRefs.join(", ")}.`,
+            `Pipeline ${activePipelineRun.id.slice(0, 8)} cannot run assignment ${pipelineInvocation.assignmentId.slice(0, 8)}: repository refs are not uniquely configured: ${resolution.unresolvedRefs.join(", ")}. Update the run repository refs/configuration or use !reset after preserving any work.`,
           );
         }
 
         const pipelineRepos = [...resolution.repos];
-        const addRepo = (repo: Config["repos"][number] | undefined) => {
+        const addRepo = (repo: RepoConfig | undefined) => {
           if (repo && !pipelineRepos.some((item) => item.name === repo.name)) {
             pipelineRepos.push(repo);
           }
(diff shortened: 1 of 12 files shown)


src/session/manager.ts
```diff
@@ -1303,37 +1309,127 @@ export class SessionManager {
...
+                  (repo) => repo.name === session.targetRepo,
+                )
+              : undefined,
+          );
+        }
+        if (pipelineRepos.length === 0) {
+          throw new Error(
+            `Pipeline ${activePipelineRun.id.slice(0, 8)} cannot run without a configured repository and isolated worktree.`,
+          );
+        }
```
PranavBakre: **blocker** — a pipeline run with empty (or unresolvable) `repoRefs` now hard-fails *every* turn in the thread, with no in-thread recovery.

Reachable today:
- `tryCreateProductRun` creates the run with `repoRefs: session.targetRepo ? [session.targetRepo] : []` (manager.ts:243), so `!pm` / `!build` in a thread that never ran `!repo` yields `repoRefs: []`.
- `pipeline_start`'s `repo_refs` is optional (`src/pipelines/tools.ts:90,183`) and `createProductRun`/`createBugRun` do not validate it (`product/controller.ts:177`, `bug/controller.ts:157`).

In that case the fallback above is a no-op (it only fires when `session.targetRepo` is set — the exact case that would have produced non-empty `repoRefs`), so we land here and throw. Consequences:

1. A `pm`/discovery assignment that legitimately has no repo yet (writing a spec) can never run.
2. The throw settles the invocation via the new `!runnerStarted` path → `escalate` → run `needs-human`. `needs-human` is **not** `terminal` (`src/pipelines/types.ts:11-15`), and nothing clears `session.activePipelineRunId`, so `activePipelineRun` stays defined and every subsequent message in the thread — including a plain human question to `lead`, which reaches this block via `pipelineRunId = ... ?? session.activePipelineRunId` (line 1315-1316) — throws again before spawn. The orchestrator cannot even explain or repair it; only `!reset` gets the thread back.

Suggested fix: scope the hard requirement to agents that actually write code (or to turns carrying a `pipelineInvocation`), and let repo-less orchestration/PM turns fall back to Junior's workspace root as before. At minimum, don't let a non-terminal wedge apply to top-level/human turns.

No test covers the empty-`repoRefs` path — `manager.test.ts` only exercises runs with a resolvable ref.

src/session/manager.ts
```diff
@@ -1303,37 +1309,127 @@ export class SessionManager {
...
+
+        const resolution = resolvePipelineRepos(
           this.config.repos,
+          activePipelineRun.repoRefs,
+        );
+        if (resolution.unresolvedRefs.length > 0) {
+          throw new Error(
+            `Pipeline ${activePipelineRun.id.slice(0, 8)} cannot run: repository refs are not uniquely configured: ${resolution.unresolvedRefs.join(", ")}.`,
+          );
+        }
```
PranavBakre, line 1357: **warning** — failing closed on unknown refs is the right instinct, but the failure is permanent from the thread's point of view. A run whose `repoRefs` name a repo that isn't in this instance's config (config changed after the run was created, or an agent passed an org repo Junior doesn't have) throws here on every turn forever, for the same reason described on the empty-`repoRefs` comment below: `needs-human` is not `terminal` and `activePipelineRunId` is never cleared.

Please pair the fail-closed throw with a recovery path — e.g. terminalize/abandon the run on an unrecoverable workspace error, or surface an actionable message (`!repo <name>` / `!reset`) instead of a bare setup error. Also worth a test: this branch is currently uncovered.

src/session/manager.ts
```diff
@@ -1987,6 +2084,37 @@ export class SessionManager {
...
+              reason: "provider_error",
+              retryable: false,
+            },
+          },
+          expectedHandle,
+          session.worktreePath ?? process.cwd(),
+        );
+        if (!setupResult) return;
+        fatalMessage = setupResult.error ?? fatalMessage;
+      }
```
PranavBakre, line 2150: **warning** — this settlement runs *inside* the `catch` block, and `resolvePipelineInvocation` can itself throw: `getRun`/`getAssignment`/`listOutcomes`/`recordOutcomeTransaction` are store calls, and line ~2255 explicitly rethrows non-ownership errors from `mutateSession`. If it throws here, the exception escapes `runRunnerWithAgent` entirely — skipping the `mutateSession` that marks the session idle/failed and the `onError` post below — and because the entry points at lines 502 and 569 invoke `runRunnerWithAgent(...)` as a floating promise with no `.catch()`, it surfaces as an unhandled rejection while the session stays `busy` with the handle still registered.

Wrap this block in its own `try/catch` and fall through to the existing setup-error path on failure.

src/worktree/pipeline-routing.ts
```diff
@@ -0,0 +1,93 @@
...
+  targetAgent: string;
+  assignmentContextRefs?: string[];
+}): RepoConfig | undefined {
+  const explicit = inferReviewRepo(input.configuredRepos, input.prompt);
+  if (
+    explicit &&
+    input.pipelineRepos.some((repo) => repo.name === explicit.name)
+  ) {
+    return explicit;
+  }
```
PranavBakre, line 64: **warning** — the explicit-URL signal outranks assignment workstream affinity for *every* agent, not just `review`. `inferReviewRepo` scans the whole prompt, and pipeline dispatch prompts embed `<bug-context>` / `<product-context>` blocks that routinely contain PR and issue URLs. A `build` assignment carrying `workstream:backend` whose bug context quotes a frontend PR link will silently start in the frontend worktree as its primary cwd.

Suggest ranking the assignment's declared workstream first for non-`review` agents (or only honouring the URL signal when `targetAgent === "review"`), since `contextRefs` is durable, authored intent while the URL is incidental prose. The existing test only covers the `review` case.

src/worktree/pipeline-routing.ts
```diff
@@ -0,0 +1,93 @@
...
+    ?.slice("workstream:".length)
+    .trim()
+    .toLowerCase();
+  const affinity = workstream || agentWorkstream(input.targetAgent);
+  if (affinity) {
+    const matches = input.pipelineRepos.filter((repo) =>
+      repoMatchesWorkstream(repo.name, affinity)
+    );
+    if (matches.length === 1) return matches[0];
+  }
```
PranavBakre, line 97: **nit** — when the affinity matches zero or ≥2 repos this silently falls through to `pipelineRepos[0]`, which for a `frontend` agent in a run whose first ref is the backend repo means starting in the backend worktree. Behaviourally survivable (the full path map is still injected and the next turn recomputes), but it's invisible. A `log.warn` on the ambiguous/no-match case would make misroutes diagnosable instead of showing up as "the agent edited the wrong repo."

src/worktree/pipeline-routing.ts
```diff
@@ -0,0 +1,93 @@
...
+    input.pipelineRepos.some((repo) => repo.name === explicit.name)
+  ) {
+    return explicit;
+  }
+
+  const workstream = input.assignmentContextRefs
+    ?.find((ref) => ref.startsWith("workstream:"))
+    ?.slice("workstream:".length)
+    .trim()
+    .toLowerCase();
```
PranavBakre: **nit** — `.find()` silently takes the first `workstream:` ref. If an assignment ever carries two (a full-stack coordinator handing down both), the second is dropped without a signal. Worth either asserting single-workstream here or logging when more than one is present.

src/session/manager.ts
```diff
@@ -1303,37 +1309,127 @@ export class SessionManager {
...
+          activePipelineRun.repoRefs,
+        );
+        if (resolution.unresolvedRefs.length > 0) {
+          throw new Error(
+            `Pipeline ${activePipelineRun.id.slice(0, 8)} cannot run: repository refs are not uniquely configured: ${resolution.unresolvedRefs.join(", ")}.`,
+          );
+        }
+
+        const pipelineRepos = [...resolution.repos];
+        const addRepo = (repo: Config["repos"][number] | undefined) => {
```
PranavBakre: **nit** — `Config["repos"][number]` here while the rest of this path (and `pipeline-routing.ts`) uses the named `RepoConfig` type. Prefer `RepoConfig | undefined` for consistency.

src/worktree/pipeline-routing.ts
```diff
@@ -0,0 +1,93 @@
...
+): PipelineRepoResolution {
+  const resolved: RepoConfig[] = [];
+  const unresolvedRefs: string[] = [];
+  const seen = new Set<string>();
+
+  for (const ref of repoRefs) {
+    const matches = repos.filter((repo) => repoMatchesRef(repo.name, ref));
+    if (matches.length !== 1) {
+      unresolvedRefs.push(ref);
+      continue;
```
PranavBakre, line 30: **nit** — resolved repos are deduped via `seen`, but `unresolvedRefs` is not: a run listing the same unknown ref twice produces a duplicated entry in the error message the human sees (`... are not uniquely configured: X, X`). Cheap to dedupe alongside `seen`.

PranavBakre (commented):
**review: changes-requested** — the isolation model is right, but the fail-closed guard wedges repo-less pipeline threads with no in-thread recovery.

Reviewed per-commit against base `edd4598`:
- `222923584` — Add deterministic pipeline repo routing (`src/worktree/pipeline-routing.ts` + test, `repoMatchesRef` exported from `review-routing.ts`)
- `387fe7e1` — Isolate pipeline agents in managed worktrees (`src/session/manager.ts`, `src/session/manager.test.ts`)

**What's good:** replacing the review-only repo inference with durable-`repoRefs`-driven provisioning is the correct direction; `ensurePipelineWorktree` correctly single-flights concurrent fan-out setup (and the test gates on it); clearing `fresh.cwd` so cold starts/recovery continuations stay in the managed worktree closes a real escape hatch; settling the pipeline invocation on pre-spawn setup failure (`runnerStarted`) stops a worktree failure from silently hanging an assignment.

**Findings: 1 blocker, 4 warnings, 3 nits.** Details inline.

Warning not attached to a line — **docs are not updated.** `CLAUDE.md`'s documentation workflow asks for a `docs/code_index/<module>.md` update after a module change; there is no code-index entry for `src/worktree/pipeline-routing.ts`, and `docs/features/bug-pipeline-worktrees.md` still describes the superseded model ("Lead calls `register_worktree` on intake, once per routed repo" — line 110/118) which this PR replaces with manager-owned provisioning. Please sync at least the bug-pipeline-worktrees feature doc and add the code index.

Verification: I reviewed read-only in the routed worktree `junior-wt-pipeline-isolation` and diffed each changed file against the `main` checkout at the PR base; no commands were run and nothing was edited.

gxt-admin (commented):
**review: approved** — every finding from the first pass is fixed and verified at `6b35ba0`. Two consecutive clean passes.

(Note: my second-pass review at `ddc1127` could not be posted — the GitHub write surface rejected it with an expired signed run context — so this review also records that pass's verification.)

### Commits reviewed (base `edd4598`)
- `222923584` Add deterministic pipeline repo routing
- `387fe7e15` Isolate pipeline agents in managed worktrees
- `306183bd3` Keep pipeline workspace failures recoverable
- `ddc1127a8` Document pipeline-owned worktree routing
- `6b35ba051` Clear stale cwd for repo-less pipelines

### Final finding — resolved
**nit (prior pass): repo-less orchestration left a stale utility `cwd`.** Fixed at `manager.ts:1388-1394` — the repo-less branch now runs its own `mutateSession` that clears `fresh.cwd`, with `assertRunOwnership()` inside the mutator matching the provisioned branch at `:1433`. Both branches of the pipeline block now uphold the stated invariant ("not a developer checkout or stale session cwd"), and `docs/code_index/pipeline-routing.md:19` ("clears stale utility cwd") is now accurate for both.

**Regression coverage present:** `manager.test.ts:709` seeds `existing.cwd = "/tmp/stale-utility-cwd"` before dispatching the repo-less `pm` assignment and asserts both the spawned `runSession.cwd` **and** the persisted `store.get("thread-1").cwd` are `null` — so the override is cleared durably, not just for the in-flight turn. That covers provider cold starts and recovery continuations too.

### Earlier findings — all verified fixed (carried from the unposted `ddc1127` pass)
- **blocker** repo-less/unresolvable run wedged every turn → routing gated on `activePipelineRun && pipelineInvocation` (`:1343`); `activePipelineInvocation` is reset from the inbound event on every message (`:485`, `:553`), so human turns can't inherit a failed assignment's gate; repo-less runs only hard-fail for roles that need a checkout, decided from the trusted catalog with unknown roles failing closed (`pipelineAgentRequiresWorktree`, `:3263`). Tests `manager.test.ts:709`, `:754`.
- **warning** unresolved refs had no recovery path → actionable error naming the assignment and `!reset` (`:1353-1357`); test `manager.test.ts:806`.
- **warning** settlement inside `catch` could escape as an unhandled rejection → own `try/catch` at `:2110-2150`, falls through to the normal setup-error path; test stubs `recordOutcomeTransaction` to throw and asserts the combined message plus `agentSessions.build.status === "failed"`.
- **warning** incidental PR URL outranked workstream affinity → URL inference scoped to `targetAgent === "review"` (`pipeline-routing.ts:56`); regression test asserts a backend build assignment ignores a frontend PR link.
- **warning** docs not updated → `docs/code_index/pipeline-routing.md` added; `docs/features/bug-pipeline-worktrees.md` marks the `register_worktree`-on-intake convention historical and records this PR as pipeline-owned provisioning.
- **nits** silent affinity fallback → `onDiagnostic` + `pipeline.workspace.affinity` warn log; multi-`workstream:` refs deduped, diagnosed, and fall through; `Config["repos"][number]` → `RepoConfig`; `unresolvedRefs` deduped via `seenUnresolved`.

### Findings this pass
**0 blockers, 0 warnings, 0 nits.**

### Verification and scope limits
Read-only inspection in the routed worktree at `6b35ba0`, diffed against the `main` checkout at base `edd4598`; re-traced the routing block, the role gate, the settlement `catch`, `ensurePipelineWorktree` (unchanged), and `pipeline-routing.ts` on both passes. The `onDiagnostic` closure capturing the `let activePipelineRun` is safe under TS 5.8 post-last-assignment closure narrowing, consistent with the reported clean `tsc --noEmit`.

Two limits stated plainly rather than assumed:
1. I ran no commands this invocation. The validation evidence — two identical clean local passes, `tsc --noEmit` clean, 1,290 passed / 2 expected embedding skips / 0 failed / 3,937 assertions — is the implementation's report, corroborated by reading the new/changed tests but not independently executed by me.
2. My read-only GitHub surface returns head SHA, reviews and inline comments only — it does not expose the PR description, so I verified the **diff, commit set, and tests** rather than the PR body prose. If the body makes claims beyond "pipeline-owned worktree routing + recoverable workspace failures + docs", those specific sentences remain unchecked by me.

Nothing was edited, committed, pushed, or merged.

PR #136, opened 2026-07-22
fix: enforce safe MongoDB MCP access

PranavBakre:
## Summary
- Keep MongoDB credentials out of child runners and block/redact direct database access so reads stay behind the read-only MCP boundary.
- Centralize the MCP tool registration Zod compatibility cast without changing dependencies; package.json and bun.lock are unchanged, and no Zod downgrade is included.

## Test plan
- [x] bun run typecheck
- [x] bun test (1316 pass, 2 skipped, 0 fail)
- [x] bun run build


src/runners/runtime.ts
```diff
@@ -125,6 +125,15 @@ export function buildRunnerEnv(
   // MCP_CONTEXT_SECRET authenticates the agent identity embedded in MCP URLs.
   // A child that receives it could forge a different agent's capability token.
   delete env.MCP_CONTEXT_SECRET;
+  // Database credentials belong to Junior's MCP backend, never to spawned
+  // coding agents. Exposing them lets an unrestricted Bash session bypass the
+  // read-only MCP surface with mongosh or a driver script.
+  delete env.MDB_MCP_CONNECTION_STRING;
```
pranav-growthx: 🔴 **Blocker — Authorization:** Deleting the inherited key does not keep this credential out of child runners. Junior starts normal/default runners in a cwd that contains the real `.env`, and Bun automatically reloads that file for child commands. I reproduced this from the configured repo cwd with `MDB_MCP_CONNECTION_STRING` explicitly removed from the environment: a fresh `bun -e` still reported the key present. Because the default OpenCode policy allows shell/read access (and the Claude deny list can be bypassed by a command that merely launches Bun without mentioning `.env` or the variable), an agent can recover the URI and connect with a driver, bypassing the read-only MCP boundary. Move the Mongo credential to storage outside every runner-readable cwd and inject it only into the MCP backend process, or enforce an OS-level read boundary that also prevents dotenv rehydration. Add a subprocess test from the actual runner cwd; asserting only on the `buildRunnerEnv` object misses this path.

pranav-growthx (commented):
Blocking security finding; formal REQUEST_CHANGES is unavailable because the authenticated reviewer owns this PR. See the inline comment.

src/runners/runtime.ts
pranav-growthx: 🔴 **Blocker — Authorization:** The headless/Bun rehydration path is fixed, but the newly added `new-session -e KEY=` protection only applies when a tmux session is created. On Junior restart, `reconcileTmuxSessions` deliberately calls `adoptExistingSession` for each surviving tmux session, so the already-running pre-fix Claude process never reaches this line and keeps the original credential environment. I verified the state transition with real tmux: a legacy session retained `MDB_MCP_CONNECTION_STRING=legacy-present` while a newly created sentinel session had an empty value. Updating tmux metadata after adoption would not repair the existing Claude process environment; the safe fix is to discard/cold-resume pre-fix sessions (or persist/version a “credential-sanitized” generation and adopt only that generation). Add a reconciliation regression proving a legacy persisted session is recreated rather than adopted. Until then, deploying this PR leaves exactly the long-lived tmux agents it is meant to protect able to bypass the MCP boundary.

pranav-growthx: ✅ Verified fixed at `d874a54739954d2ed213853961ccb35ce027f46c`. Reconciliation now accepts only sessions carrying every explicit empty credential sentinel; missing/non-empty legacy environments fail closed, the live tmux process is discarded, and the persisted Claude session ID is retained for a sanitized cold resume. Real tmux behavior and the new top-level reconciliation regression both confirm the transition.

pranav-growthx (approved):
Reviewed at d874a54739954d2ed213853961ccb35ce027f46c: two consecutive clean passes, no blockers or should-fix items found.

PR #138, opened 2026-07-23
feat: short follow-up interruption + durable speaker identity

PranavBakre:
## Summary

- **Short follow-up interruption**: adds a narrow interrupt-and-consolidate path for short, same-author conversational follow-ups. When an eligible short message arrives while Junior is busy on a same-author turn (and no tool side-effects have started), the active turn is interrupted and a consolidated prompt replaces it. Feature is disabled by default behind `SESSION_SHORT_FOLLOWUP_INTERRUPT_ENABLED`.

- **Durable speaker identity**: preserves the original human Slack user ID (`sourceSlackUserId`) through the full assignment → outbox → dispatch → prompt path. Author and mentions now use visibly different syntax (`User(Name <@ID>)` for author vs `@Name` for inline mentions), preventing the identity confusion where a mentioned person was mistaken for the message author.

- **Pump deduplication**: when a new child assignment's follow-up text is identical to its objective, the pump no longer duplicates it with a `[task-follow-up]` wrapper.

## Key files

- `src/session/followup-policy.ts` — pure eligibility function with `BusyFollowupDecision`
- `src/session/manager.ts` — interrupt flow, response suppression, consolidated replay, inline mention resolution
- `src/pipelines/types.ts` — `sourceSlackUserId` on assignments
- `src/pipelines/pump.ts` — deduplication for new-child objective/follow-up
- `src/slack/thread-context.ts` — inline mention mode (`@Name` format)
- `src/config.ts` — feature flag + thresholds

## Test plan

- [x] Pure policy function tests (all eligibility conditions)
- [x] Session types expected-keys test updated
- [x] Forged block delimiter test updated for inline mention format
- [x] Two consecutive clean typecheck + test suite passes
- [ ] Integration: exercise short follow-up with feature flag enabled
- [ ] Verify Slack shows one coherent response for a short conversational burst


src/session/manager.ts
```diff
@@ -587,10 +637,46 @@ export class SessionManager {
...
+      this.activeTurnSideEffects.set(handleKey, false);
+      this.runRunnerWithAgent(
+        session,
+        drainPrompt,
+        event.ts,
+        event.files,
+        agentName,
+        event.user,
+        reservation,
+      );
```
PranavBakre: **warning** — Interrupt path produces a malformed consolidated prompt.

`buildDrainPrompt` (line 649) emits structural `<buffered-message from="<@ID>">` blocks. Passing `event.user` here triggers the attributable-sender path in prompt construction (line 1734), which runs `escapeBlockDelimiters(prompt)` — mangling the structural tags into `&lt;buffered-message from="@Name">`.

The regular drain path (line 2962) avoids this by passing `senderUserId = undefined`, so `escapeBlockDelimiters` never runs on the structural tags.

When enabled, the consolidated prompt will contain escaped tags that lose their structural meaning. Options:
1. Pass `senderUserId = undefined` here (match the regular drain) and rely on `buildDrainPrompt`'s `from` attribute for attribution
2. Build a dedicated consolidated prompt (the plan's `<conversation-burst>` envelope) that bypasses the per-message escaping path

src/session/manager.ts
```diff
@@ -2783,6 +2887,8 @@ export class SessionManager {
             s.pendingMessages = [];
             s.activePipelineInvocation = null;
             s.status = "idle";
+            s.activeTurnAuthor = null;
+            s.activeTurnWasInterrupted = false;
             settle.action = "muted-discard";
           } else if (pendingMessages.length > 0) {
```
PranavBakre, line 3015: **warning** — Drain path doesn't reset `activeTurnAuthor` / `activeTurnWasInterrupted`.

The muted-discard branch (2890-2891) and idle-settle branch (2907-2908) both clear these, but the drain branch does not. The drain turn inherits stale state:

- `activeTurnAuthor` = old turn's author → same-author check uses wrong baseline if a different user's message is being drained
- `activeTurnWasInterrupted = true` persists from a prior interrupt → blocks interrupts during drain even though the drain is a new burst

Suggestion: reset both fields here, same as the other branches.

src/session/followup-policy.ts
```diff
@@ -0,0 +1,53 @@
...
+  if (ctx.messageLength > ctx.maxMessageLength) {
+    return { action: "buffer", reason: "message-too-long" };
+  }
+  if (ctx.activeTurnHasSideEffects) {
+    return { action: "buffer", reason: "side-effects-started" };
+  }
+  if (ctx.activeTurnAlreadyInterrupted) {
+    return { action: "buffer", reason: "already-interrupted" };
+  }
+  return { action: "interrupt", reason: "eligible-short-followup" };
```
PranavBakre, line 109: **warning** — Policy omits several plan-specified eligibility conditions.

The plan doc (lines 78-88) specifies for the initial policy:
- 240 chars AND 40 words → implementation uses 280 chars only (no word limit)
- 20-second window from active message start → not checked
- Same top-level agent slot → not checked
- Neither message contains files, commands, pipeline/dispatch metadata → not checked
- Exclude persistent worker-agent sessions → not checked

The pure-function design makes these easy to add, and the feature is disabled. But the gap means when enabled, the interrupt path would fire on cases the plan explicitly excludes (e.g. messages with files, commands, or arriving 60s into a long turn).

Is this intentional simplification for v1?

src/pipelines/pump.ts
```diff
@@ -193,20 +193,34 @@ async function handleOutboxItem(
...
     }
 
+    // Build resume prompt, deduplicating when follow-up text matches objective.
+    let prompt: string | undefined;
+    if (item.eventType === "assignment.resume") {
+      if (typeof item.payload.prompt === "string") {
+        // When the follow-up text is identical to the objective (common for
+        // new child assignments whose objective IS the human message), avoid
+        // prepending it again with a redundant [task-follow-up] wrapper.
+        if (item.payload.prompt === assignment.objective) {
```
PranavBakre: **nit** — The plan says "avoid solving this with loose text deduplication" and wants the child-vs-resume distinction encoded in the outbox event/payload. This is text comparison (`item.payload.prompt === assignment.objective`), which works for the common case but would suppress the `[task-follow-up]` wrapper if a user happens to send identical text to an existing assignment.

A `payload.isNewChild: boolean` flag on the outbox event would make this structural.

src/session/manager.ts
```diff
@@ -86,6 +88,7 @@ export class SessionManager {
   private store: SessionStore;
   private config: Config;
   private handles = new Map<string, SpawnHandle>();
+  private activeTurnSideEffects = new Map<string, boolean>();
```
PranavBakre, line 91: **nit** — `activeTurnSideEffects` entries are set to false/true but never deleted. Over the bot's lifetime, one entry accumulates per unique thread+agent key. The values are small (boolean), so this is cosmetic — but a `.delete(handleKey)` in the settle path would keep the map bounded.

PranavBakre (commented):
Review of PR #138 — 6-pass structured review, per-commit.\n\nTypecheck: clean. Tests: 1329 pass / 1 pre-existing fail (agents-org submodule missing in worktree).\n\nPipeline identity fix is solid — sourceSlackUserId flows correctly through assignment → store → pump → dispatch → prompt attribution. SQLite migration is safe (nullable column, try/catch for idempotency). Inline mention mode is backward-compatible and preserves the security property (block delimiter escaping still prevents forged author attribution).\n\n3 warnings, 2 nits. See inline comments.

gxt-admin (commented):
Re-review after 713d671. All 3 prior warnings resolved:\n\n1. **Interrupt prompt fix** ✓ — `runRunnerWithAgent` now passes `undefined` for `latestTs`, `files`, and `senderUserId`, matching the regular drain path. Structural `<buffered-message>` tags are no longer mangled by `escapeBlockDelimiters`.\n2. **Drain state reset** ✓ — `activeTurnAuthor` and `activeTurnWasInterrupted` now cleared in the drain branch, matching muted-discard and idle-settle.\n3. **Side-effects cleanup** ✓ — `activeTurnSideEffects.delete(handleKey)` added in the non-drain settle path, keeping the map bounded.\n\nTypecheck clean. 1329 tests pass (1 pre-existing fail: agents-org submodule).\n\nTwo consecutive clean passes. Approved.

PR #140, opened 2026-07-23
feat(runbooks): add schema, loader, validator, registry, and MCP tools

PranavBakre:
## Summary
Iteration 1 of the common-task-agent-authoring plan. Builds the foundational runbook infrastructure:

- **Types** (`src/runbooks/types.ts`) — `RunbookDefinition`, `RunbookInput`, `RunbookRisk`, validation error types
- **Capability bundles** (`src/runbooks/capabilities.ts`) — 8 provider-neutral bundles (mongo.read, migration.execute, slack.read/post, credential.deliver, github.read/propose, migration.inspect) with subset validation against owner agent catalog capabilities
- **Loader** (`src/runbooks/loader.ts`) — YAML frontmatter parser for `.runbook.md` files with SHA-256 content digest
- **Validator** (`src/runbooks/validator.ts`) — name/kebab-case, schema version, owner agent resolution, input type allowlist, risk-to-gate enforcement, capability subset check, secret scanning
- **Registry** (`src/runbooks/registry.ts`) — fixed-root discovery, last-known-good reload, search/get/list/digest
- **MCP tools** — `runbook_search` and `runbook_get` registered alongside existing agent tools
- **Fixture** — `transfer-ai-roadmaps.runbook.md` example from the plan doc

## Test plan
- [x] 85/85 runbook tests pass (loader, validator, registry, capabilities — 198 assertions)
- [x] 138/138 existing agent tests pass (no regressions)
- [x] Fixture loads, validates, is searchable, has stable content digest
- [x] Capability widening beyond owner agent fails closed
- [x] High-risk runbooks without approval gates rejected
- [x] Secret patterns in content detected and rejected
- [x] Last-known-good registry behavior on invalid reload

### DIFF
--- a/src/runbooks/loader.ts
+++ b/src/runbooks/loader.ts
@@ -1,0 +1,310 @@
+import type { RunbookDefinition, RunbookLoadResult } from "./types.ts";
+import { validateRunbook } from "./validator.ts";
+
+export async function loadRunbookDefinition(
+  filePath: string,
+  options?: { origin?: "private" | "public"; filenameForValidation?: string },
+): Promise<RunbookLoadResult> {
+  const file = Bun.file(filePath);
+  if (!(await file.exists())) {
+    return {
+      ok: false,
+      errors: [{ field: "file", message: `file not found: ${filePath}` }],
+      filePath,
+    };
+  }
+
+  const raw = await file.text();
+  const contentDigest = computeDigest(raw);
+  const { frontmatter, body } = extractFrontmatter(raw);
+
+  if (!frontmatter) {
+    return {
+      ok: false,
+      errors: [{ field: "frontmatter", message: "missing YAML frontmatter" }],
+      filePath,
+    };
+  }
+
+  const parsed = parseYamlFrontmatter(frontmatter);
+  const origin = options?.origin ?? "private";
+
+  const definition: RunbookDefinition = {
+    schemaVersion: toNumber(parsed.schemaVersion) ?? 0,
+    name: toString(parsed.name) ?? "",
+    description: toString(parsed.description) ?? "",
+    ownerAgent: toString(parsed.ownerAgent) ?? "",
+    intent: {
+      examples: toStringArray(parsed.intent?.examples) ?? [],
+      excludes: toStringArray(parsed.intent?.excludes) ?? [],
+    },
+    inputs: toInputArray(parsed.inputs),
+    risk: toString(parsed.risk) as RunbookDefinition["risk"],
+    approval: {
+      required: toBool(parsed.approval?.required) ?? false,
+      afterSteps: toStringArray(parsed.approval?.afterSteps),
+    },
+    capabilities: toStringArray(parsed.capabilities) ?? [],
+    verification: {
+      required: toBool(parsed.verification?.required) ?? false,
+      assertions: toStringArray(parsed.verification?.assertions) ?? [],
+    },
+    tags: toStringArray(parsed.tags) ?? [],
+    prompt: body.trim(),
+    filePath,
+    origin,
+    contentDigest,
+  };
+
+  const filename = options?.filenameForValidation ?? filenameStem(filePath);
+  const errors = validateRunbook(definition, filename);
+
+  if (errors.length > 0) {
+    return { ok: false, errors, filePath };
+  }
+
+  return { ok: true, definition };
+}
+
+function computeDigest(content: string): string {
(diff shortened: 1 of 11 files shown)


src/runbooks/loader.ts
```diff
@@ -0,0 +1,310 @@
...
+
+    const indent = countIndent(line);
+    if (indent < baseIndent) break;
+
+    const trimmed = line.trim();
+    if (!trimmed.startsWith("- ")) break;
+
+    const value = trimmed.slice(2).trim();
+    const colonIdx = value.indexOf(":");
+    if (colonIdx !== -1 && value.slice(colonIdx + 1).trim() !== "") {
```
PranavBakre, line 194: **warning** — YAML parser silently mis-parses colon-containing array scalars.

`parseArray` checks whether the value after `- ` contains a colon with non-empty content after it (line 194), and if so, treats the item as an object. For plain string array items that happen to contain colons — e.g. `- connect via http://example.com` or `- source count: 0` in verification.assertions — this produces `{"connect via http": "//example.com"}` instead of the string `"connect via http://example.com"`. The downstream `toStringArray` then coerces that to `"[object Object]"`.

Current runbook fields (intent examples, tags, assertions) are unlikely to contain colons, so no current fixture breaks. But it's a latent parsing bug with no error signal — a future runbook author writing `- step: run the migration` in intent.examples would get silently wrong data.

Quoted strings in arrays (`- "step: run the migration"`) would also hit this since quote-stripping happens inside `parseScalar`, which runs after the object-vs-scalar decision.

Mitigation options: check if the item has continuation keys at deeper indent before deciding it's an object, or only treat `- key: value` as an object when the indent matches the "array of objects" pattern (continuation lines at indent+2).

src/runbooks/registry.ts
```diff
@@ -0,0 +1,162 @@
...
+        origin: dir.origin,
+        filenameForValidation: file.replace(/\.runbook\.md$/, ""),
+      });
+
+      if (result.ok) {
+        registry.set(result.definition.name, result.definition);
+        loaded++;
+      } else {
+        // Last-known-good: don't replace a valid entry with a broken reload
+        if (!registry.has(result.filePath)) {
```
PranavBakre: **warning** — LKG log suppression check is a no-op.

`registry.has(result.filePath)` compares a file path (e.g. `agents-org/runbooks/transfer-ai-roadmaps.runbook.md`) against registry keys, which are runbook names (e.g. `transfer-ai-roadmaps`). This never matches, so `!registry.has(result.filePath)` is always `true`, and the warning fires on every reload failure — even when a valid LKG entry for that runbook already exists in the registry.

The actual LKG behavior (keeping the old entry by not calling `registry.set` on failure) works correctly. Only the intended log suppression is broken.

Same issue on line 139 in `loadRunbookRegistryFromDir`.

Fix: derive the expected name from the filename and check that:
```js
const expectedName = file.replace(/\.runbook\.md$/, "");
if (!registry.has(expectedName)) {
  log.warn(...);
}
```

PranavBakre (commented):
**review: approved** — solid runbook infrastructure with 2 warnings

Reviewed both commits per-commit. Types, loader, YAML parser, validator, capabilities, registry, MCP tools, fixture, and all 85 tests examined. Capability subset enforcement verified against actual agent manifest data (build has repo-write, review does not). All tests pass locally (85/85, 198 assertions).

**Pass 0 — Spec:** PR body claims match the actual diff. Types, 8 capability bundles, loader with SHA-256 digest, validator with 13 rules, registry with LKG reload, 2 MCP tools, fixture — all present.

**Pass 1 — Logic:** Two issues found (see inline comments). Everything else traces correctly — extractFrontmatter, YAML parser (for the supported subset), type coercions, validator rule ordering, capability subset enforcement, registry lifecycle, MCP tool registration.

**Pass 2 — Safety:** No secrets, PII, or credentials in code or fixtures. `as` casts are bounded by immediate validation. Secret scanner covers the core patterns.

**Pass 3 — Product:** Fixture is a concrete production use case. MCP tools follow existing `agent_search`/`agent_get` patterns.

**Pass 4 — Query perf:** N/A — in-memory Map.

**Pass 5 — Consistency:** Custom YAML parser is consistent with project principle (pure functions over framework ceremony). Test patterns match existing `agents.test.ts` style. Module-level registry with test-clear function matches existing patterns.

**Pass 6 — Surface:** Clean file naming, consistent imports with `.ts` extensions, no unused imports.

0 blockers / 2 warnings / 0 nits

PR #145, opened 2026-07-24
fix: bind repositories during pipeline dispatch

PranavBakre:
## Summary
- validate and atomically bind repository refs before dispatching worktree-backed agents
- allow human recovery assignments to resume escalated runs with corrected repository context
- expose the binding through MCP, preserve legacy memory tag filtering, and document the recovery contract

## Test plan
- [x] `bun run typecheck`
- [x] `bun test src/pipelines/tools-dispatch.test.ts src/pipelines/store/sqlite.test.ts src/mcp/slack-server.test.ts`
- [x] `bun test`

### DIFF
--- a/.claude/agents/common/pipeline-outcome.md
+++ b/.claude/agents/common/pipeline-outcome.md
@@ -14,6 +14,10 @@ and GitHub comments do not settle an assignment.
   ownership to exactly one successor.
 - `agent_dispatch(mode="delegate")` creates a child assignment. Your assignment
   waits and is resumed automatically after the child reports `complete`.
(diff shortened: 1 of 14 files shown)


pranav-growthx (approved):
Reviewed exact head 85ed8a476e37f2d32ce0047877810836223d1168 against base 5553159830257a1b84d52784a281bcef21464f71: two consecutive clean six-pass reviews, no blockers or should-fix items. Typecheck, targeted tests, expanded integration tests, and full suite (1644 passed, 2 skipped, 0 failed) are green.

PR #146, opened 2026-07-24
fix: suppress duplicate default escalation audits

PranavBakre:
## Summary
- Suppress public `assignment.escalate` audit messages for ordinary default runs so Slack threads receive only the agent's user-facing response.
- Preserve typed product/bug escalation audits and system-generated wait-timeout alerts, with regression coverage and updated agent guidance.

## Test plan
- [x] `bun run typecheck`
- [x] `bun test src/pipelines/pump-wait.test.ts`
- [x] `bun test`

### DIFF
--- a/.claude/agents/common/pipeline-outcome.md
+++ b/.claude/agents/common/pipeline-outcome.md
@@ -27,6 +27,11 @@ and GitHub comments do not settle an assignment.
   satisfied. For a default run it completes the task; typed controllers may
   advance or fan in according to their own policy.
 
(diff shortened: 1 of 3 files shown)


pranav-growthx (approved):
Reviewed: two consecutive clean passes, no blockers or should-fix items found. Verified head 2a271574ba021e1ed7b6c666d9316f72f6650d4e against base 4d2a28a43349b409353de8ce0bd1d77bbb7fe925 with targeted regression tests, full test suite, typecheck, build, and diff check.

PR #152, opened 2026-07-27
fix: preserve Codex prompts and scope pipeline context

PranavBakre:
## Summary
- Compile scoped worker context from Junior's pipeline handoff, centralizing recall and worker fan-out while avoiding duplicate Slack history and provider-native subagents.
- Preserve Codex's native base prompt by layering Junior instructions as developer instructions, and complete the documented `initialize` → `initialized` handshake.
- Add prompt-context telemetry and regression coverage for context budgeting, dispatch boundaries, and Codex startup behavior.

## Test plan
- [x] Run two consecutive clean TypeScript checks (`tsc`)
- [x] Run the full test suite (1810 passed, 4 skipped)
- [x] Run the production build

### DIFF
--- a/.claude/agents/build.md
+++ b/.claude/agents/build.md
@@ -1,7 +1,7 @@
 ---
 name: build
 description: Backend engineer. Use for building features, fixing bugs, refactoring code.
-tools: Read, Edit, Write, Bash, Grep, Glob, Agent, mcp__slack-bot__memory_recall
+tools: Read, Edit, Write, Bash, Grep, Glob, mcp__slack-bot__memory_recall
 permissions.intent: normal
 common: core,building-philosophy,pipeline-outcome
 context.threadHistory: true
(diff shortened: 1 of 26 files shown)


.claude/agents/build.md
```diff
@@ -27,6 +33,8 @@ Recall memory before starting (task query + repo `entity_refs` like `gx-backend:
 - **You own:** edits in the authorized worktree, focused verification, and explicit-path **checkpoint commits** when the assignment authorizes mutation.
 - **Orchestrator owns:** aggregate verification across agents/repos, push/PR create-or-update, phase transitions, and human gates.
 - **Review owns:** read-only findings and a typed verdict — never product-code edits.
+- **Junior owns fan-out:** do not spawn provider-native subagents. Report a
```
gxt-admin, line 36: 🔴 **Blocker — Structure/Business Logic:** This prompt says Junior owns fan-out, but removing `Agent` from the frontmatter does not remove the provider-native delegation surface at runtime. `build`/`frontend` resolve to the normal intent, and `src/claude/policy.ts` runs that intent with `bypassPermissions`, an empty allowlist, and no `Agent` disallow entry, so Claude still exposes Agent. OpenCode likewise compiles normal builders to the fallback `allow` policy and both spawners still register native support subagents (`loadOpenCodeSupportSubagents()`), leaving `task` available. The new lint test only parses this Markdown line, so it passes while the runtime boundary remains open. A worker can still create the duplicate provider-native fan-out this PR is intended to eliminate. Enforce the restriction in the provider policy/config surfaces for build/frontend (and add provider-level tests that assert Agent/task/collaboration is unavailable), while preserving durable `agent_dispatch`.

gxt-admin (changes requested):
Requesting changes. See inline comment.

.claude/agents/build.md
gxt-admin: Re-review on `e69eda0`: Claude and OpenCode now close this runtime boundary, but Codex still leaves native fan-out enabled. The generated config adds `[agents]\nenabled = false`; with that exact `CODEX_HOME`, `codex features list` reports `multi_agent stable true`. Running the same CLI with the recognized override `-c features.multi_agent=false` reports `multi_agent stable false`. So `[agents].enabled` is not the multi-agent feature gate; it configures the agents table while the native collaboration surface remains active, and the developer instruction is still only advisory. Please emit the recognized Codex feature setting (and test the effective config/feature state, not only the TOML substring). The original blocker remains open.

gxt-admin: Verified fixed on `3a30392`. Claude disallows `Agent` for every intent, OpenCode denies `task` and registers no Slack-runtime subagents, and the generated Codex home now uses `[features]\nmulti_agent = false`. Codex 0.145 loads that home with `multi_agent stable false`; it remains false even when a trusted target repo attempts to set `multi_agent=true`. Two consecutive clean review passes completed with provider tests, typecheck, build, and diff checks clean. This blocker is closed.

gxt-admin (approved):
Reviewed: two consecutive clean passes, no blockers or should-fix items found. Original provider-native fan-out blocker verified closed across Claude, OpenCode, and Codex.

PR #154, opened 2026-07-28
fix: restore pipeline re-review loop

PranavBakre:
## Summary
- Resume product pipelines from the durable human gate when a human requests re-review
- Preserve rejected typed handoffs and require reviewer completions to include a consistent typed verdict
- Cover waiting-assignment routing and the re-review-to-fixing flow with regression tests

## Test plan
- [ ] `bun test src/pipelines/product/controller.test.ts`
- [ ] `bun test src/support/router-pipeline-mode.test.ts`


gxt-admin (approved):
Reviewed after fixes and exact-head integration verification: two clean structured review passes, typecheck clean, full suite clean twice.

PR #158, opened 2026-07-29
feat: add native assignment-scoped skill dispatch

PranavBakre:
## Summary
- Add native assignment-scoped skills with durable `skillRef` and `capabilityRefs` dispatch envelopes, avoiding skill-body prompt inflation.
- Load skills through Claude CLI, OpenCode CLI, and Codex app-server provider-native paths, with per-assignment MCP compilation.
- Keep stateless skill runs lightweight by skipping worktrees, resume state, memory, and direct Slack replies; also resolve dashboard dispatch timestamp ties.

## Test plan
- [x] `bun run typecheck && bun test && bun run build` — clean run 1: 1853 passed, 4 skipped, 0 failed
- [x] `bun run typecheck && bun test && bun run build` — clean run 2: 1853 passed, 4 skipped, 0 failed


src/skills/registry.ts
```diff
@@ -0,0 +1,96 @@
...
+const SUPPORT_SKILLS: readonly TrustedSkillDefinition[] = [
+  {
+    name: "nr-research",
+    description: "Investigate New Relic telemetry and record bounded pipeline evidence.",
+    relativePath: "nr-research/SKILL.md",
+    execution: "stateless",
+    capabilities: ["pipeline-artifact-write"],
+    permissions: {
+      intent: "read-only",
+      mcp: ["slack-bot"],
```
pranav-growthx, line 30: 🔴 **Blocker — Business Logic:** This capability envelope gives `nr-research` only the Slack control-plane MCP and no provider tools; the Sentry and Vercel entries below are identical. Because these runs are `read-only` with no `worktree-verify`, Claude/OpenCode deny Bash, Codex is read-only, and undeclared MCP servers are not compiled. The skill can write/settle an artifact, but it has no way to query New Relic, Sentry, or Vercel, so every registered skill is unable to perform its stated objective. Add explicit trusted read-only integrations/capabilities for each service (and compile them for every supported provider), or allow narrowly-scoped trusted CLI commands.

src/runners/policy.ts
```diff
@@ -105,30 +106,32 @@ export function compileOpenCodePermission(options: {
...
-    ...(hasCapability(agentName, "github-review-comment")
+    ...(subjectHasCapability(options.subject, "github-review-comment")
       ? { [GITHUB_POST_REVIEW_TOOL]: "allow" }
       : {}),
-    ...(hasCapability(agentName, "pipeline-artifact-write")
+    ...(subjectHasCapability(options.subject, "pipeline-artifact-write")
       ? {
           "mcp__slack-bot__pipeline_get_state": "allow",
           "mcp__slack-bot__pipeline_report_outcome": "allow",
+          "mcp__slack-bot__pipeline_write_artifact": "allow",
```
pranav-growthx, line 117: 🔴 **Blocker — Authorization/Business Logic:** OpenCode evaluates matching permission patterns in insertion order and the last match wins. These exact `mcp__slack-bot__...` allows are emitted before the trailing `mcp__*`: `deny` in the read-only branch, so both rules match and the catch-all wins. On the default OpenCode provider a skill therefore cannot call `pipeline_write_artifact` or `pipeline_report_outcome`, which means it cannot satisfy or settle its assignment. Put the broad deny before the exact capability allows (and add a test that evaluates effective matching order, not just object values).

src/session/manager.ts
```diff
@@ -1762,7 +1801,31 @@ export class SessionManager {
...
-      let runSession = this.buildRunSession(session, agentName, agentIdentity);
+      const applyAssignmentEnvelope = (
+        candidate: ThreadSession,
+      ): ThreadSession => {
+        if (!activeSkill) {
+          candidate.activeSkill = null;
+          candidate.assignmentCapabilities = [];
+          return candidate;
+        }
+        candidate.activeSkill = {
```
pranav-growthx, line 1812: 🟡 **Should-fix — Product Thinking:** This activates a skill for every selectable runner provider, including `opencode-sdk`, but that adapter never resolves `activeSkill`, exposes a skill directory, invokes the native skill tool, or sends a skill part. With a thread configured as `opencode-sdk`, Junior silently runs the plain assignment prompt without the SKILL.md body. Implement SDK-native loading, or reject skill assignments on that provider with a clear unsupported-provider error instead of executing the wrong workload.

src/skills/runtime.ts
```diff
@@ -0,0 +1,71 @@
...
+    const stat = lstatSync(linkPath);
+    if (
+      stat.isSymbolicLink() &&
+      resolve(dirname(linkPath), readlinkSync(linkPath)) === targetPath
+    ) {
+      return;
+    }
+    throw new Error(`skill runtime path already exists with unexpected target: ${linkPath}`);
+  }
+  symlinkSync(targetPath, linkPath, "dir");
```
pranav-growthx: 🟡 **Should-fix — Bug:** This check-then-create is racy across concurrent assignments for the same skill. Two runs can both observe the link as absent; one creates it and the other gets `EEXIST`, failing an otherwise valid provider spawn. Catch `EEXIST` and revalidate the now-existing symlink target (or use another atomic/idempotent materialization strategy).

src/pipelines/tools.ts
```diff
@@ -955,6 +959,146 @@ export async function pipelineDispatchAgent(
...
+  }
+  const sourceAgent = canonicalAgentName(source.targetAgent) ?? source.targetAgent;
+  if (sourceAgent !== caller) {
+    return textResult(
+      { ok: false, reason: "authenticated agent does not own the active assignment" },
+      true,
+    );
+  }
+
+  const stableKey = `${run.id}:${source.id}:skill:${args.idempotency_key}`;
```
pranav-growthx: 🟡 **Should-fix — Data Integrity:** The idempotency namespace omits `skill.name`. If one assignment dispatches multiple skills with the same semantic key (for example one shared `observability:last-hour` key), only the first child is created; later calls return that child while claiming the newly requested skill and capabilities in the response. Include the canonical skill name in the durable key, or validate the existing child envelope and return a conflict for a mismatched replay.

pranav-growthx (changes requested):
Requesting changes. See inline comments for two blockers and three should-fix issues.

pranav-growthx (approved):
Approved after two clean structured review passes on exact head fe49dae1ce1cc01290e31411351c49081f4e7776, retargeted to main after parent PR 157 merged.

PR #159, opened 2026-08-02
feat: add cumulative profile consolidation and browser

PranavBakre:
## What

- expands memory claims with preference and decision types plus project profiles
- builds cumulative person profiles from cross-thread Slack evidence
- builds cumulative repository profiles with worktree alias normalization
- derives recurring situation profiles from cross-thread Slack patterns
- adds historical and targeted profile backfill controls
- adds a read-only Profiles dashboard for person, repo, project, and situation profiles
- includes the SQLite claim-kind migration, documentation, and coverage

## Why

Profile generation previously depended on one-time thread consolidation. Once a thread was consumed, Junior could not accumulate evidence across sessions, so person, repository, and situation profiles were rarely or never updated.

## Safety and impact

- every generated profile must cite evidence from its exact review window
- persona and situation prompts reject sensitive inference and treat source text as untrusted
- dashboard inspection never bumps memory usage timestamps
- profile data remains local, gitignored, and available only through the loopback operator console
- no force push or schema data loss

## Verification

- `bun run typecheck` passed twice
- full `bun test` passed twice
- `bun run build` passed twice
- dashboard module syntax and `git diff --check` passed
- focused memory/HTTP suite: 225 passed, 4 expected model-download skips, 0 failed

## Operational backfill

The private local backfill evaluated all historical Slack people earlier, then reviewed repository and cross-thread situation evidence. Current derived totals are 14 people, 5 repositories, 5 situations, and 0 pending Slack messages. These generated profile files are private/gitignored and are not included in this PR.


gxt-admin (approved):
Reviewed after fixes and integration with current main: two clean structured passes, typecheck clean, and full suite clean twice (1961 pass, 4 model-dependent skips).

PR #160, opened 2026-08-03
feat: add hybrid memory retrieval

PranavBakre:
## What
- Fuse vector similarity and exact-token coverage with normalized reciprocal-rank fusion.
- Preserve source path, heading, and section text so atomic claim hits can return useful context.

## Why
- Recall should recover exact identifiers and wording without giving up semantic matching.

## Impact
- Memory recall can admit strong lexical hits below the cosine floor and expose bounded parent-source context across MCP, CLI, HTTP, and pre-recall flows.

## Root cause
- Cosine-only ranking missed identifier-heavy queries, while atomic claims did not retain enough source provenance to reconstruct their surrounding context.

## Checks
- [x] `tsc --noEmit`
- [x] Two clean full `bun test` passes (no failures/errors; four skipped)


PranavBakre:
Superseded by #159, which contains the safe hybrid retrieval implementation plus Slack archive authorization/ANN/thread-context work. Review of this PR found correctness defects (post-limit eligibility, stale source context, unbounded source fields, and unconditional lexical fusion); the applicable fixes were incorporated and verified in #159 before merge.

PR #161, opened 2026-08-03
Fix workflow run status display

PranavBakre:
## What changed

- expose a workflow `displayStatus` that prioritizes an in-progress run over scheduler state
- show the effective status and scheduler status separately in the dashboard
- cover running, stopped, invalid, and file-level fallback states

## Why

The workflow card derived its prominent badge only from the definition's `enabled` flag. As a result, an enabled workflow always appeared `active` even while its latest run was `running`, and a workflow stopped at runtime could also be mislabeled.

## Impact

Operators can now see that a workflow is currently running while still seeing whether its scheduler is active, stopped, or invalid.

## Validation

- `bun run typecheck`
- `bun test` — 1,870 passed, 4 model-download tests skipped
- `bun run build`
- headless dashboard render with an active scheduler and running workflow fixture


gxt-admin (approved):
Reviewed after fixes and current-main integration: two clean structured passes, typecheck clean, full suite clean twice (1966 pass, 4 model-dependent skips).

PR #162, opened 2026-08-05
Fix multi-repo review intake

PranavBakre:
## What changed

- collect every configured repository referenced by GitHub PR URLs during review intake
- durably bind those repositories before the review assignment is dispatched
- keep single-repository affinity selection separate from multi-repository provisioning
- add parser and end-to-end pipeline regression coverage for the three-PR failure

## Root cause

Default review runs copied only the thread’s single `targetRepo`. On a fresh thread that value was empty, while the review URL inference intentionally returned a repository only when exactly one URL matched. A request containing PRs from three repositories therefore reached the worktree guard with no durable repository bindings.

## Impact

Multi-repository review requests now provision all referenced repository worktrees before the reviewer starts, preventing the pipeline from escalating with a missing repository/worktree error.

## Validation

- `bun test src/worktree/review-routing.test.ts src/session/manager.test.ts` (140 passed)
- `bun run typecheck`
- `git diff --check`


gxt-admin (approved):
Reviewed after authorization fixes and current-main integration: two clean structured passes, typecheck clean, full suite clean twice (1976 pass, 4 model-dependent skips).

PR #163, opened 2026-08-06
Compile agent catalog from trusted frontmatter

PranavBakre:
## What changed

- moves lifecycle, role, capabilities, mutation policy, permission ceiling, and handoff declarations into each trusted agent Markdown file
- replaces the hard-coded TypeScript catalog records with a schema-validating compiler for Junior and agents-org frontmatter
- keeps target-repository agent overrides outside the operational trust root
- fails fast on incomplete metadata, duplicate names or aliases, unsafe MCP-only tools, incompatible write capabilities, and unresolved handoffs
- keeps agents-org optional for public installations and derives provider matrix role names from the compiled catalog

## Why

Agent prompts and operational policy were maintained in separate Markdown and TypeScript records. Those records could drift and leave an agent registered with incomplete or incompatible behavior. The Markdown definition is now the single declaration, while TypeScript remains responsible for validation and enforcement.

## Dependency

- GrowthX-Club/junior-private-agents#17 adds the onboard-member operational declaration referenced by this submodule pointer

## Validation

- git diff --check
- bun run typecheck
- bun test
- bun run build

The complete verification sequence passed twice after implementation and review.


gxt-admin (approved):
Reviewed after trusted-catalog fixes, private dependency merge, and current-main integration: two clean structured passes, typecheck clean, full suite clean twice (1991 pass, 4 model-dependent skips).

PR #164, opened 2026-08-10
activate pinned runbooks at startup

PranavBakre:
## What changed

- loads Git-pinned private and public runbooks during Junior cold start
- projects activated definitions, content digests, and source commit SHAs into a durable SQLite catalogue
- injects that catalogue into the MCP server so `definitions_status` works
- deactivates catalogue entries removed from the pinned sources
- makes evaluation use the production routing threshold
- ignores bound email values during intent matching
- adds cold-start coverage that loads, catalogues, and selects a real runbook
- pins the private runbook commit from GrowthX-Club/junior-private-agents#18

## Why

The runbook components existed only as isolated primitives. The runtime never loaded the registry or initialized the catalogue, while the sole definition remained a test fixture. Existing activation tests preloaded that fixture and therefore missed the cold-start failure.

## Impact

After the private dependency merges and this PR lands, Junior will activate the reviewed `transfer-ai-roadmaps` runbook on restart and expose its exact Git provenance through Luna. Production-write approval and verification gates remain unchanged.

## Validation

- full `bun test` suite passes
- runbook suite: 276 passed, 0 failed
- `bun run typecheck`
- `bun run build`
- `git diff --check`

## Merge order

1. Merge GrowthX-Club/junior-private-agents#18.
2. Merge this PR.
3. Restart Junior and verify `definitions_status` plus the 7-fixture Luna evaluation.


gxt-admin (approved):
Reviewed after authorization, registry, dependency, handoff, and current-main integration fixes: two clean structured passes, typecheck clean, full suite clean twice (1996 pass, 4 model-dependent skips).

PR #165, opened 2026-08-12
feat: add weekly Slack archive maintenance

PranavBakre:
## Summary
- Add an explicit typed `nativeHandler` registry for deterministic workflows, with no filename magic or LLM execution, and schedule Slack archive maintenance weekly on Sunday at 03:17 Asia/Kolkata.
- Use the existing bot token to auto-join every visible public channel while preserving the approved-private-channel boundary, then repair checkpoint gaps and older threads, embed pending changes, and atomically rebuild the ANN index only when vectors changed or the index is absent.
- Requires granting the Slack app the `channels:join` OAuth scope and reinstalling it; the current live bot token does not have that scope.

## Test plan
- [x] `bun run typecheck` (two clean serial runs)
- [x] `bun test` (two serial runs: 2001 passed, 4 model-dependent skipped, 0 failed each)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -134,6 +134,8 @@ WHATSAPP_AUTH_DIR=data/whatsapp-auth
 # separate database and is never consolidated into Junior memory.
 SLACK_ARCHIVE_ENABLED=false
 SLACK_ARCHIVE_DB_PATH=data/slack-archive.db
+# Weekly full-public-channel sync requires the bot OAuth scope `channels:join`;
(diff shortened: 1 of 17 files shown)


gxt-admin (approved):
Reviewed at exact head 0ba1d328 after fixes. Two clean structured review passes, focused regressions clean twice, typecheck clean, and full suite clean twice (2003 pass, 4 model-dependent skips).

PR #170, opened 2026-08-14
feat: prune scoped worktrees after pull request merges

PranavBakre:
## Summary
- Trigger `worktree-prune` asynchronously when GitHub reconciliation observes a pull request transition to merged, limiting cleanup to the exact repo, branch, and head SHA while retaining existing safety gates.
- Coalesce newly merged targets and retry overlap skips so the workflow's concurrency policy avoids duplicate runs without dropping merge-triggered cleanup.
- Propagate structured trigger context through workflow execution and artifacts, with focused coverage and updated workflow/reconciliation docs.

## Test plan
- [x] Two focused test passes (40/40 each)
- [x] `bun run typecheck`
- [x] `bun run build`
- [x] Full suite: 2027 passed, 4 skipped; 2 Codex app-server timing flakes
- [x] Isolated Codex app-server test rerun: 7/7 passed

### DIFF
--- a/src/github/prune-trigger.ts
+++ b/src/github/prune-trigger.ts
@@ -1,0 +1,117 @@
+import type { ShadowPersistResult } from "./types.ts";
+
+export type MergedPullRequestPruneTarget = {
+  owner: string;
+  repo: string;
+  number: number;
+  branch: string;
+  headSha: string;
+};
+
+export type MergePruneRunResult = { runId: string };
+
+export type MergePruneDispatcherOptions = {
+  run: (
+    targets: readonly MergedPullRequestPruneTarget[],
+  ) => Promise<MergePruneRunResult>;
+  retryDelayMs?: number;
+  scheduleRetry?: (callback: () => void, delayMs: number) => void;
+  onRun?: (
+    targets: readonly MergedPullRequestPruneTarget[],
+    result: MergePruneRunResult,
+  ) => void;
+  onError?: (error: unknown) => void;
+};
+
+const DEFAULT_RETRY_DELAY_MS = 30_000;
+
+/**
+ * Collapse newly persisted merge events into the exact PR branches that should
+ * prompt a worktree-prune run. A resource can be associated with several
+ * pipeline runs, so key by GitHub coordinates rather than association count.
+ */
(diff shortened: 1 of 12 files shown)


gxt-admin (changes requested):
Requesting changes. See inline comments.

workflows/worktree-prune.workflow.md
```diff
@@ -34,6 +34,13 @@ Prune stale local git worktrees across the configured Junior repos.
 
 Use the runtime context as the source of truth for repositories and their absolute paths. For each repo with a usable local git checkout:
 
+When `run.triggerContext.source` is `github.pr.merged`, limit the run to the
+listed `pullRequests`: inspect only the matching repo and worktree branch for
```
gxt-admin, line 38: 🟡 **Should-fix — Scoping:** The runner is told to match the exact merged repository, but runtime `repos` entries omit `RepoConfig.githubRepo` and expose only a local name/path/base. Configured names may be aliases, and forks can share both a basename and commit SHA, so this can miss the intended checkout or inspect the wrong owner despite the SHA guard. Include canonical `githubRepo` in runtime context, require an exact case-insensitive `owner/repo` match, and test an alias/fork collision.

src/github/prune-trigger.ts
```diff
@@ -0,0 +1,117 @@
...
+    this.pending.clear();
+
+    try {
+      const result = await this.options.run(targets);
+      this.options.onRun?.(targets, result);
+      if (!result.runId) {
+        for (const target of targets) this.addPending(target);
+        this.scheduleRetry();
+      }
+    } catch (error) {
```
gxt-admin, line 111: 🟡 **Should-fix — Business Logic:** `targets` were removed from `pending`, but this catch only logs. A transient scheduler/store/provider failure—or failed workflow registry bootstrap—therefore loses the sole merge transition permanently because the persisted snapshot is already `MERGED` and later reconciliation emits no replacement event. Requeue targets with a bounded/backed-off retry and test the rejection path; the current test covers only empty-`runId` overlap skips.

gxt-admin (approved):
Reviewed: two consecutive clean passes, no blockers or should-fix items found.

PR #171, opened 2026-08-14
Recover workflow and maintenance fixes

PranavBakre:
## What changed

- pass bounded operator instructions through manual and custom workflow commands, while preserving workflow safety rules and structured event context
- repair missing or stale lesson retrieval variants with backup-first, idempotent publication and source-change protection
- ensure new generic lesson claims create their complete retrieval projection set
- drain delegated worktree setup output concurrently and retain restricted full transcripts while surfacing bounded diagnostics

## Why

A recovered local stash contained several unrelated experiments. This PR keeps only the changes that still address verified gaps. The live memory database currently has two active lessons missing variants 1 and 2, manual maintenance workflows cannot currently accept one-run scoping, and verbose setup scripts can deadlock or hide stdout-only failure reasons.

The unsafe or obsolete stash pieces were deliberately excluded: a debug cron change, lockfile-only deletion, unreviewed runbook activation, and raw long-error exposure.

## Verification

- `bun run typecheck` — clean twice
- `bun run build` — clean
- focused suite — 61 passed twice
- full suite — 2,042 passed, 4 skipped, 0 failed
- `git diff --check`


src/workflows/executor.ts
```diff
@@ -153,6 +160,7 @@ export class WorkflowExecutor {
         run,
         summary,
         triggerContext: request.triggerContext ?? null,
+        instructions: normalizeInstructions(request.instructions),
```
gxt-admin: 🟡 **Should-fix — Business Logic:** This records operator instructions for every run, but the `nativeHandler` branch above never consumes them. For example, `!workflow run memory-consolidation only thread X` will consolidate the full corpus and then write an artifact claiming `only thread X`, which makes the audit trail actively misleading. Reject free-text instructions for native workflows unless that handler has a typed, enforced scoping implementation (or pass validated scope into the native handler and test it).

src/memory/reembed-retrieval.ts
```diff
@@ -332,6 +350,156 @@ export function backupDatabase(dbPath: string, backupPath: string): void {
...
+       WHERE c.active = 1 AND c.kind = 'lesson'
+       ORDER BY c.id`,
+    ).all();
+    sources = rows.map((row) => {
+      const [firstLine = row.authoritative_text, ...remainingLines] =
+        row.authoritative_text.split(/\r?\n/);
+      return {
+        id: row.id,
+        title: row.lesson_title?.trim() || firstLine.trim(),
+        body: row.lesson_body?.trim() ||
```
gxt-admin: 🟡 **Should-fix — Data Integrity:** The repair prefers `lesson.body`/`lesson.title` whenever a legacy row exists, even though `claim.text` is authoritative and this module already tests that the legacy row can be stale. If a claim was corrected without updating the legacy row, this publishes vectors for the obsolete rule while the claim returns the corrected text. Build the lesson content from `c.text` (using only compatible auxiliary context such as `applies_when`), and add a regression test with deliberately divergent legacy and authoritative text.

src/worktree/manager.ts
```diff
@@ -293,16 +298,58 @@ export class WorktreeManager {
       stdout: "pipe",
       stderr: "pipe",
     });
+    // Drain both pipes before awaiting exit so a chatty setup script can't
+    // fill its stdout buffer and deadlock.
+    const [stdout, stderr] = await Promise.all([
```
gxt-admin: 🟡 **Should-fix — Safety & Runtime:** Draining concurrently fixes the pipe deadlock, but `Response(...).text()` still buffers both complete streams in memory—even on successful setup. A runaway/chatty installer can now grow Junior until OOM. Stream stdout/stderr incrementally to the restricted transcript while retaining only a bounded tail in memory for the thrown error; that preserves the full diagnostic without making output volume a process-memory risk.

gxt-admin (changes requested):
Requesting changes. See inline comments.

gxt-admin (approved):
Reviewed: two consecutive clean passes, no blockers or should-fix items found. All three prior findings are resolved at c29975a.

PR #194, opened 2026-08-24
fix(codex): disable default environments on thread lifecycle

PranavBakre:
## Summary
- disable Codex's default local environment on thread start and resume
- add regression assertions for fresh and resumed app-server threads
- document the tool-isolation contract

## Validation
- `bun test src/codex-app-server/spawner.test.ts`
- `bun run typecheck`

### DIFF
--- a/docs/code_index/codex-app-server.md
+++ b/docs/code_index/codex-app-server.md
@@ -34,6 +34,11 @@ Generated Codex config sets `[features].multi_agent = false`; provider-native
 subagents would bypass Junior's durable assignment, context, and settlement
 contracts.
 
+Thread creation and resume explicitly send `environments: []`. This disables
+Codex's default local command environment so the thread's tool surface stays
+limited to Junior's scoped MCP configuration; the invariant is covered by the
+app-server spawner regression tests for both fresh and resumed threads.
+
 Native `requestApproval` server callbacks are not auto-denied. The provider
 posts a scoped Allow/Deny action in the originating Slack thread, waits on the
 same in-process resolver used by Claude's permission tool, and returns the
--- a/docs/code_index/codex-app-server.md
+++ b/docs/code_index/codex-app-server.md
@@ -39,6 +39,14 @@ Codex's default local command environment so the thread's tool surface stays
 limited to Junior's scoped MCP configuration; the invariant is covered by the
 app-server spawner regression tests for both fresh and resumed threads.
 
+Before release, run `CODEX_APP_SMOKE_SLACK_CHANNEL_ID=<approved-test-channel>
+CODEX_APP_SMOKE_RELEASE_GATE=1 bun run codex:release-gate` with an authenticated
+Codex CLI and a running Junior MCP server. This opt-in live gate requires an
+approved `slack_send_message` dynamic-tool call, then asks Codex to execute a
+forced `touch`; it fails if no MCP tool item is observed, if any command item is
+emitted, or if the probe file appears. The gate uses a temporary directory and
+removes it on exit.
+
 Native `requestApproval` server callbacks are not auto-denied. The provider
 posts a scoped Allow/Deny action in the originating Slack thread, waits on the
 same in-process resolver used by Claude's permission tool, and returns the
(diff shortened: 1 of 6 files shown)


PranavBakre:
Validation update:
- `bun test src/codex-app-server/spawner.test.ts` — 11 passed
- `bun run typecheck` — passed
- `bun test` — 2136 passed, 4 skipped, 66 failed in pre-existing private agents/runbooks fixtures unavailable in this isolated public worktree; no F06-related failures.

PranavBakre:
Review verdict: changes requested (formal request unavailable on my own PR).

The implementation correctly adds `environments: []` to fresh start, resume, and fallback start. I independently ran `bun test src/codex-app-server/spawner.test.ts` (11 passing) and `bun run typecheck` (passing).

Blocking F06 acceptance gap: the audit explicitly requires a release-gated behavioral proof—a permitted dynamic-tool/MCP probe plus a forced shell-command denial probe. The new fake JSON-RPC assertions prove only serialization of `environments: []`; they cannot prove that the real Codex app-server honors it or that the built-in command environment is absent.

Please add a live/opt-in release-gate that (1) invokes an allowed scoped dynamic MCP tool successfully and (2) attempts a shell command and proves it is unavailable or denied with no command side effect. Wire it into release validation, and revise the docs to distinguish the parameter regression test from the behavioral isolation proof. This is required before F06 can be marked fixed.

PranavBakre:
F06 release-gate follow-up pushed in `3983d1f`.

- Added `bun run codex:release-gate` (explicit opt-in).
- Requires `CODEX_APP_SMOKE_SLACK_CHANNEL_ID` and authenticated Codex + running Junior MCP server.
- Sends `environments: []` on live start/resume.
- Positive probe requires observed `slack-bot/slack_send_message`.
- Negative probe asks Codex to `touch` a temporary path and fails on command/shell item or filesystem side effect.
- Temp probe directory is cleaned up.
- `node --check scripts/codex-app-server-smoke.mjs`, focused Codex tests (11 pass), and `bun run typecheck` pass.
- Missing-channel invocation fails closed before spawning Codex.

PranavBakre:
Re-review verdict: changes still requested (formal request unavailable on my own PR).

The gate is correctly opt-in, has a channel preflight, sends `environments: []` on start/resume, and checks the probe file before cleanup. I also verified the smoke script syntax, diff whitespace, and the existing spawner suite (11/11).

Blocking fail-open: the positive probe treats *any* observed `mcp_tool_call`/`mcp_tool_use` item as success. A Slack MCP invocation that completes with `status: failed` or an `error` still satisfies `slackItems.length > 0`, so the release gate can pass without an approved dynamic tool having executed successfully. Require a completed `slack-bot/slack_send_message` item with a successful status and no error (and, if available in the protocol, successful result); otherwise fail.

The negative assertion still only proves that no command item/side effect was observed. Because the model may decline or ignore the prompt, that can pass without an attempted shell denial. Strengthen it to require evidence that the explicitly requested shell operation was rejected/unavailable (or otherwise make the attempt deterministic), while retaining the no-command-item and no-side-effect checks. With those two changes the gate will be fail-closed for the stated F06 acceptance criteria.

PranavBakre:
Second review blocker addressed in the latest commit. The release gate now fails closed unless:

- the dynamic Slack MCP item reaches `item/completed` with no error and a non-empty concrete result;
- the forced shell turn provides explicit agent refusal or terminal-failure evidence;
- no command/shell item appears; and
- the temporary side-effect file remains absent.

Checks: `node --check scripts/codex-app-server-smoke.mjs`, 11 focused Codex tests pass, and `bun run typecheck` passes.

PranavBakre:
Final re-review: ready to merge.

The release gate now fails closed for the F06 evidence requirements: the positive path requires a completed `slack-bot/slack_send_message` item with no error and a non-empty concrete result; the negative path requires explicit refusal or terminal failure, and independently rejects command-tool events and the probe-file side effect. The gate remains opt-in and fails early without an approved Slack channel.

Verified locally: smoke-script syntax check and `bun test src/codex-app-server/spawner.test.ts` (11 passing). I did not run the live gate because it deliberately sends a real Slack message.

PR #195, opened 2026-08-24
fix(memory): isolate consolidation runners

PranavBakre:
## Summary\n- run Claude consolidation in a neutral cwd with stdin prompts, no tools, strict empty MCP config, hooks disabled, and no session persistence\n- run OpenCode consolidation through a dedicated inline deny-all agent with plugins/.claude discovery disabled and hostile prompt flag separation\n- document the cross-provider untrusted-content boundary and add adversarial argv/environment tests\n\n## Validation\n- bun test src/memory/consolidation\n- bun run typecheck\n\nFixes audit finding F01.


PranavBakre:
## Review: changes requested

**Blocker — F01 is not fully closed: ambient user configuration remains reachable.**

The new `cwd: tmpdir()` only removes project-local discovery. Claude still runs without `--bare`/an equivalent `--setting-sources` restriction, so Claude Code can load the invoking user's `~/.claude/CLAUDE.md`; `--settings '{"disableAllHooks":true}'` disables hooks, but it does not disable those user instructions. OpenCode likewise only unsets `OPENCODE_CONFIG`/DIR: OpenCode merges `~/.config/opencode/opencode.json` and global rule/plugin surfaces with `OPENCODE_CONFIG_CONTENT`. Its deny-all config protects tool calls, but does not prevent ambient instructions/config from affecting the consolidation output, and global MCP/plugin configuration can still be discovered/initialized. I reproduced the merge with the PR environment: `opencode debug config --pure` still resolved the global plugin `~/.config/opencode/plugins/rtk.ts` (and the global agent catalog) despite `--pure`; `--pure` means “without external plugins”, not “ignore user config”.

Please make both subprocesses use a truly isolated config/context boundary (for example Claude `--bare` plus explicit empty MCP config, and an isolated HOME/XDG/OpenCode config root or equivalent documented no-user-config flags), while retaining only the credential inputs needed to run. Add an adversarial live subprocess probe with hostile user/project config and verify that no hook/plugin/MCP/instruction surface is inherited. The current tests only inspect constructed argv/env objects; they do not prove the spawned providers honor the isolation contract.

Focused validation on PR head:
- `bun test src/memory/consolidation`: 68 passed
- `bun run typecheck`: passed (with the repository's installed Bun types linked into the detached worktree)

This is a security blocker for the untrusted Slack/source-record boundary, so I cannot approve yet.

PranavBakre:
## Re-review: one remaining blocker

The Claude blocker is addressed: `--safe-mode` is a supported installed CLI flag and explicitly suppresses CLAUDE.md, skills, plugins, hooks, MCP, commands, and agents. The OpenCode hostile-user-config live probe passes, and the full focused suite is green (69 tests) with typecheck passing.

However, the new sterile OpenCode HOME/XDG_DATA_HOME breaks the configured default runtime authentication. This installation’s `opencode-go/deepseek-v4-pro` credential is stored at `~/.local/share/opencode/auth.json` (`opencode auth list` reports OpenCode Go there), while the runner now points `XDG_DATA_HOME` at a fresh empty directory. No OpenCode API credential environment variable is inherited in this runtime, and the PR does not copy/inject the provider credential. Therefore the production default OpenCode consolidation path will fail authentication even though the config probe passes.

Please preserve only the provider credential needed by the selected model (for example, copy the auth entry into the per-run sterile data root with restrictive permissions, or use an explicit supported credential env/config injection), and add a live probe/runtime test proving the selected provider can authenticate with the sterile roots without importing the rest of the user config. Keep the fail-closed behavior when no credential exists.

Validation on `dee853e`: `bun test /tmp/junior-pr195-review/src/memory/consolidation` → 69 passed; `bun run typecheck` → passed.

PranavBakre:
## Final re-review: one cleanup blocker remains

Auth-only provisioning is otherwise correct: the live hostile-config probe passes; only `auth.json` is copied, destination permissions are `0600`, non-regular auth paths fail closed, and the full focused suite is green (71 tests) with typecheck passing.

However, `openCodeRunText` creates/provisions `isolatedHome`, then calls `Bun.spawn(...)` before entering the `try/finally` that removes it (`runner.ts:415-428` and `444-463`). If spawn throws (missing/unlaunchable `opencode`, invalid executable, or a synchronous OS error), the cleanup `finally` is never installed. The per-run directory and copied credential remain on disk. This is both a cleanup failure and a temporary credential leak.

Please wrap `Bun.spawn` in the cleanup scope (or catch spawn errors and `rm` the isolated home before rethrowing), and add a failure-path test/probe asserting the sterile home is removed when spawn fails. After that I expect this to be clean.

PranavBakre:
## Final re-review: clean

The spawn-throw cleanup blocker is fixed. OpenCode setup and process creation now run inside `withIsolatedOpenCodeRuntime`, whose `finally` removes the sterile home/auth root for synchronous spawn failures, normal exits, timeouts, and provider errors. The new regression test verifies removal on synchronous process-creation failure.

Previously identified concerns remain closed: Claude `--safe-mode` suppresses user context; OpenCode uses sterile HOME/XDG roots, copies only the regular auth file with `0600` permissions, fails closed for suspicious auth paths, and passes the hostile-config live probe.

Validation on `94f3749`: `bun test /tmp/junior-pr195-ready/src/memory/consolidation` → 72 passed; `bun run typecheck` → passed.

Verdict: ready from the F01 isolation review. I cannot submit a formal GitHub approval because the authenticated account owns the PR; this comment is the clean review verdict.

PR #196, opened 2026-08-24
fix: bound worktree subprocesses

PranavBakre:
## Summary
- centralize worktree Git/setup execution behind process-group timeouts
- cancel output drains after terminating descendants and retain bounded tails
- make Git/setup transport noninteractive with disabled credential prompting

## Validation
- bun test src/worktree/manager.test.ts
- bun run typecheck

Fixes audit finding F07.


PranavBakre (commented):
Independent F07 review: clean verdict. The bounded executor covers inline Git and delegated setup, drains both pipes concurrently with bounded tails, terminates the detached process group before cancelling readers, and applies the noninteractive Git transport contract. Focused integration suite (19 tests) and typecheck pass. Scope note: GitHubAuthResolver's gh subprocess buffering/bounds remain F14 work, not a regression in this F07 PR.

PR #197, opened 2026-08-24
fix: preserve Slack attachments in default durable runs

PranavBakre:
## Summary
- persist bounded SlackFileAttachment refs on default assignments and outbox payloads
- restore validated refs on synthetic pump dispatches so runners download attachments
- cover the full default-run route plus SQLite and bounds behavior

## Verification
- `bun run typecheck`
- `bun test src/session/manager.test.ts -t "preserves Slack attachments through the durable default run"`
- `bun test src/slack/files.test.ts src/pipelines/store/sqlite.test.ts -t "durable Slack attachment refs|round-trips runs and assignments"`

The selected broader suite also exposed five unrelated existing SessionManager failures in onboarding/Codex isolation tests; no changes were made to those paths.


PranavBakre:
Review verdict: CLEAN (author approval is unavailable from the same GitHub identity). Traced Slack ingress → default assignment/outbox → SQLite/memory persistence → pump JSON restoration → synthetic event → SessionManager download/runner prompt. References are bounded before durability, the SQLite migration is backward-compatible, and older outbox rows fall back to assignment files. Verified `bun run typecheck` and the focused attachment/default-route/SQLite suite: 3 pass, 0 fail. No blockers found.

PR #199, opened 2026-08-24
fix: pin trusted agent definitions to default branch

PranavBakre:
## Summary
- activate operational agent definitions only when their on-disk bytes exactly match the owning repository default branch
- mark dirty, untracked, and feature-branch-only definitions unpublished and exclude them from the trusted catalog
- fail closed when this removes the required default orchestrator

## Verification
- `bun test src/agents/manifest.test.ts`
- `bun run typecheck`
- `bun run build`

## Notes
`bun test` was attempted but cannot complete in this isolated worktree because the private `agents-org` submodule clone is unavailable to this environment; pre-existing runbook tests require its pinned private agents.

### DIFF
--- a/docs/code_index/agent-routing.md
+++ b/docs/code_index/agent-routing.md
@@ -12,6 +12,7 @@ Resolves agent definitions across a layered search chain (target repo → privat
 | `AgentRouter.resolveAgent(session)` | `router.ts` | Searches target repo → org overlay → public fallback. First match wins. Returns `AgentDefinition \| null`. |
 | `AgentRouter.composeSystemPrompt(session)` | `router.ts` | Builds the system prompt = selected shared playbooks + agent body. Each `common:` name resolves target-repo common first, then the canonical `support/skills/<name>/SKILL.md`; matching org common overlays append additively. |
 | `loadAgentDefinition(filePath)` | `loader.ts` | Reads a `.md` file, parses frontmatter (flat `key: value`, dot-notation `context.<flag>`, quoted-value strip). Returns `null` if missing. |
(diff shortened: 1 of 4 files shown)


PranavBakre:
BLOCK: provenance canonicalizes the source path with realpathSync before deriving the Git tree path. An untracked .md symlink can therefore grant authority: in a temp Git repo, tracked payload.txt contained a valid operational definition, untracked public/rogue.md symlinked to payload.txt, and inspectAgentDefinitionProvenance(rogue.md) returned {status: "published", defaultBranchRef: "main"}; loadTrustedAgentCatalog then returned ["default", "rogue"]. This violates the documented untracked/branch-only fail-closed guarantee. Compare the original configured path (after canonicalizing repoRoot only for macOS /private/var normalization), and/or require git ls-files --error-unmatch plus a regular-file mode check for the exact path. Focused manifest tests pass (11/11); bun run typecheck passes. The provenance bypass must be fixed and covered before merge. GitHub cannot accept a formal REQUEST_CHANGES review from the PR author, so this is posted as the blocking verdict.

PranavBakre:
RE-REVIEW PASS (after 9929f7f): the prior symlink provenance bypass is closed. lstatSync now rejects symlink/non-regular definitions before Git resolution, exact-path git ls-files validation rejects untracked aliases, and the tracked-file byte comparison remains pinned to the selected default-branch ref. Focused manifest suite: 12/12 passed, including the tracked payload symlink regression; bun run typecheck passed. Startup import also succeeds. No remaining F04 blockers found in this review.

PR #200, opened 2026-08-24
fix(workflows): require default-branch provenance

PranavBakre:
## Summary
- verify public and private workflow blobs against origin/main before registry activation
- fail closed for branch-only, untracked, dirty, missing-ref, and unpublished definitions
- persist verified default-branch commit SHA on workflow runs, artifacts, SQLite, and dashboard projections
- add Git-backed provenance state tests and documentation

## Verification
- bun run typecheck
- bun test src/workflows/registry.test.ts src/workflows/store.test.ts
- bun test src/workflows
- bun test (baseline failures require the private agents-org overlay/submodule; 2150 pass, 66 fail)

F05 from the memory-lessons compliance audit.


PranavBakre:
Review verdict: BLOCK.

1. `verifyWorkflowProvenance()` accepts an untracked file when it reuses a default-branch path with identical contents. Repro: commit `workflows/worklog.workflow.md` on `main`, run `git rm --cached workflows/worklog.workflow.md` while leaving the file on disk, then verify it. The function returns `verified` though `git status --short` reports `D` + `??`, because it only hashes the worktree and never validates index tracking/cleanliness. Staged-only divergence can also pass when worktree contents are reset to the default blob. Explicitly reject untracked/staged/unstaged paths and add a regression test.

2. An unpublished overlay fails open to its public workflow. `WorkflowRegistry.reload()` installs public definitions, then for a provenance error executes the new `continue`, skipping the existing overlay guard that deletes the public fallback. Thus a dirty/untracked branch-only overlay silently activates public behavior. An unpublished overlay must block its public fallback just like an invalid overlay; add public+overlay provenance coverage.

Focused tests and `bun run typecheck` pass, but these policy gaps block merge.

PranavBakre:
Re-review verdict: CLEAN for commit `85c71c8`.

The prior blockers are resolved:
- Provenance now checks `git status --porcelain=v1 --untracked-files=all` for the exact workflow path before accepting its blob. Regression tests cover staged/index divergence with default-matching worktree bytes and `git rm --cached` plus identical untracked bytes.
- An unpublished overlay now deletes the same-name public fallback during reload, while not retaining a prior overlay; regression coverage verifies fail-closed cold boot. The control flow also performs this eviction on later reloads.

Verification: focused workflow/dashboard tests passed (48 tests), `bun run typecheck` passed, and `git diff --check` passed. No CI checks are configured for this PR.

PR #201, opened 2026-08-24
fix(slack): preserve ignored dotenv files during cleanup

PranavBakre:
## Summary
- enumerate ignored `.env` / `.env.*` paths in `WorktreeManager.getWorktreeStatus` without reading file contents
- make Slack Cleanup worktree refuse removal when ignored dotenv state exists
- share the dotenv classifier with scheduled worktree pruning
- add multiline-secret, unrelated-ignored-file, and direct action-path regression coverage

## Verification
- `bun run typecheck`
- `bun test src/worktree/manager.test.ts src/worktree/prune.test.ts src/slack/action-buttons.test.ts` (33 pass)
- `bun test` (2,153 pass; 66 unrelated baseline failures in private-overlay/runbook tests on this checkout; 4 skipped)

Fixes audit finding F08.

### DIFF
--- a/docs/code_index/worktree-manager.md
+++ b/docs/code_index/worktree-manager.md
@@ -13,6 +13,7 @@ Creates, removes, and inspects git worktrees in target repos for per-thread code
 | `removeWorktree(repoName, threadId)` | `manager.ts` | `git worktree remove --force` + `git branch -D` (queries actual branch name from the worktree first to handle `branchOverride`) |
 | `worktreeExists(repoName, threadId)` | `manager.ts` | Filesystem check via `node:fs/promises.stat` |
 | `isWorktreeDirty(worktreePath)` | `manager.ts` | `git status --porcelain` |
+| `getWorktreeStatus(worktreePath, repoName?)` | `manager.ts` | Reports tracked/untracked changes, unpushed commits, and ignored dotenv paths (path names only; contents are never read). |
 | `getWorktreePath(repoName, threadId)` | `manager.ts` | `<repo.path>.junior-worktrees/slack-<threadId>` — sibling, NOT under `.claude/` |
 | `getBranchName(threadId)` | `manager.ts` | `slack/<threadId>` |
 | `getRepo(name)` | `manager.ts` | Lookup in `repos` |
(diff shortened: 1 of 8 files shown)


PranavBakre:
## BLOCK — ignored dotenv scan can silently truncate before classification\n\ngetWorktreeStatus calls runGit, whose runBoundedCommand/startOutputDrain keeps only the last 64 KiB (MAX_GIT_STREAM_CHARS). The new logs/worktree-setup/2026-08-24T04-12-52-105Z-slack-fail-thread.log
logs/worktree-setup/2026-08-24T04-12-53-537Z-slack-pipe-holder-thread.log
logs/worktree-setup/2026-08-24T04-12-54-582Z-slack-partial-fail-thread.log output includes every ignored path, so a normal large ignored tree (for example node_modules/) can push a root .env.local out of the retained tail. ignoredDotenvPaths then returns [] and cleanup force-removes the worktree, losing the dotenv file.\n\nI reproduced this on the PR branch with a temp repo containing .env.local plus 20,000 ignored node_modules/file-*.js files: raw ls-files output was ~529 KB, but getWorktreeStatus reported ignoredDotenv: [].\n\nPlease constrain the Git query to dotenv pathspecs (including root and nested .env/.env.*) or otherwise make detection streaming/complete, and add a regression test where dotenv output precedes >64 KiB of unrelated ignored paths. Until this is fixed, F08 is not safe in typical worktrees.

PranavBakre:
## CLEAN — F08 follow-up verified\n\nRe-tested d90018e. The scan now uses dotenv-only Git pathspecs (, , ), so unrelated ignored output cannot evict dotenv paths from the 64 KiB bounded tail. On Git 2.39.5 (Apple Git), direct verification found root , nested , and nested ; / were correctly excluded. The added 5,000-file ignored-noise regression passes and exercises output beyond the stream cap.\n\nFocused verification: ; 34 targeted tests passed (manager/prune/action-buttons). No dotenv contents are read or logged; action cleanup refuses when paths are found, and prune shares the classifier. The ordinary check-then-force-remove race remains the pre-existing cleanup/prune model, but no new F08-specific regression was found.

PranavBakre:
## CLEAN — F08 follow-up verified\n\nRe-tested d90018e. The scan now uses dotenv-only Git pathspecs (root .env, glob nested .env, and glob nested .env.*), so unrelated ignored output cannot evict dotenv paths from the 64 KiB bounded tail. On Git 2.39.5 (Apple Git), direct verification found root .env, nested .env, and nested .env.production; .envrc/foo.env were correctly excluded. The added 5,000-file ignored-noise regression passes and exercises output beyond the stream cap.\n\nFocused verification: bun run typecheck; 34 targeted tests passed (manager/prune/action-buttons). No dotenv contents are read or logged; action cleanup refuses when paths are found, and prune shares the classifier. The ordinary check-then-force-remove race remains the pre-existing cleanup/prune model, but no new F08-specific regression was found.

PR #203, opened 2026-08-24
fix(pre-recall): drain provider pipes with bounded tails

PranavBakre:
## Summary
- drain pre-recall stdout and stderr concurrently from process start
- retain bounded tails and reject oversized stdout
- apply the shared drainer to Codex stderr and bound reader cleanup
- add pipe-capacity/runaway-output boundary tests and docs

## Verification
- `bun run typecheck`
- `bun test src/memory/pre-recall.test.ts` (50 passed, 2 skipped: local embedding tests)


PranavBakre:
Independent review verdict: **BLOCKED**.

The Codex pre-recall provider still has an unbounded model-output path: `codexRunText` passes `-o <outFile>` to `codex exec` and then calls unrestricted `readFile(outFile, "utf8")`. The new shared drainer only observes stderr because Codex stdout is explicitly ignored, so it cannot bound the primary response. A runaway provider can still allocate an arbitrary-size response in memory; F10 is therefore not fully closed for every provider.

Required before merge: impose the same explicit cap on the Codex output-file path (bounded read + overflow rejection, or bounded stdout), and add a provider-level oversized-output test.

Checks run locally: `bun run typecheck`; `bun test src/memory/pre-recall.test.ts` (48 pass, 2 expected local-provider skips); direct finite 70,000-byte stdout rejection; timed detached-process-group cleanup returned in 663ms with no live group.

GitHub does not permit me to submit a formal request-changes review on my own PR, so this is posted as the blocking verdict.

PranavBakre:
Re-review verdict: **READY**.

Commit `3d1816f` closes the prior blocker. Codex now applies a 64 KiB byte cap to its `-o` file before decoding: it checks the observed file size and performs a 64 KiB + 1-byte bounded read, so post-stat growth present at read time is rejected rather than allocated. The file handle closes in `finally`, and `codexRunText` still removes the temporary output file in its outer `finally`, including overflow/failure paths. The new boundary test creates a 128 KiB Codex output file and asserts rejection.

Re-verified locally: `bun run typecheck`; `bun test src/memory/pre-recall.test.ts` (48 pass, 2 expected local-provider skips). No remaining blocker found. GitHub does not allow me to submit a formal approval on my own PR, so this comment is the final review verdict.

PR #204, opened 2026-08-24
fix: terminate orphaned runner process groups

PranavBakre:
## Summary\n- terminate a surviving detached process group before orphan repair clears top-level or persistent-agent session state\n- avoid a direct-PID fallback once the recorded leader has exited, preventing PID-reuse risk\n- add real detached wrapper/helper regression coverage and document the lifecycle invariant\n\n## Verification\n- bun test v1.2.17 (282dda62)\n-

### DIFF
--- a/docs/code_index/process-lifecycle.md
+++ b/docs/code_index/process-lifecycle.md
@@ -9,6 +9,7 @@ Timeout guards, graceful shutdown, stale session cleanup, orphan detection, and 
 | Symbol | File | Purpose |
 |---|---|---|
 | `withTimeout(handle, timeoutMs, onTimeout?)` | `timeout.ts` | Wraps `SpawnHandle`; kills + resolves with `error: "Process timed out..."` after timeout |
+| `terminateProcessGroup(pid, opts)` | `process-tree.ts` | Health-only teardown for descendants whose recorded detached-group leader has already exited; never falls back to the reusable positive PID |
 | workflow runner idle recovery | `src/workflows/executor.ts` | For workflow runner configs with `idleTimeoutMs`, sends SIGINT after a silent period, SIGKILL after 10s grace if needed, then respawns with provider-native resume and a continuation prompt up to `maxIdleInterrupts`. |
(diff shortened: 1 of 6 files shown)


PranavBakre:
F11 review verdict: CLEAN / no blockers found.

Checks: `bun test src/lifecycle` (81 passed, 0 failed); `bun run typecheck` passed; `git diff --check` clean. The detached-wrapper integration tests cover both top-level lead and persistent-agent orphan repair, verify helpers die before state is cleared, and exercise the no-positive-PID-fallback path.

I could not submit an approval because GitHub rejects self-approval for this PR. Full repository test run had 2,171 passed / 66 unrelated failures caused by the private `agents-org` overlay and runbook fixtures absent from this isolated checkout; these do not touch F11.

PranavBakre:
Follow-up review finding (BLOCKER): the CAS guard is only `(status === "busy" && pid === orphanPid)`, so it is not PID-reuse-safe for state clearing. Between `isPidAlive(orphanPid) === false` / `terminateProcessGroup()` and `mutateThread()`, a new turn can start and receive the same recycled PID; the callback then clears that new turn. This is the same race the PR avoids for direct signaling, and the comments currently claim the generation guard prevents it. Capture and compare an ownership generation/stateVersion (top-level `activeTurnGeneration`; a persisted per-agent generation or parent stateVersion for agents) before clearing, and add a deterministic race regression test.

Therefore my final verdict is BLOCK until this guard is fixed. The earlier clean note only covered the process-group behavior/tests; it does not waive this state-ownership race.

PranavBakre:
Re-review of 7e52c2e: CLEAN / no remaining blockers.

The `stateVersion` ownership token is captured from the health snapshot and required alongside status/PID for both top-level lead and persistent-agent repairs. A concurrent replacement write increments the version, so even deliberately reusing the same numeric PID cannot be cleared. The lead regression is deterministic via `SnapshotBarrierStore`; I also ran an equivalent deterministic persistent-agent race check, which passed.

Checks: focused health/process-tree tests 4 passed; lifecycle suite 82 passed, 0 failed; typecheck passed; diff check clean. State is still cleared only after detached process-group teardown. GitHub self-approval remains unavailable, so no approval or merge was performed.

PR #207, opened 2026-08-24
fix(worktree): verify removal postconditions

PranavBakre:
## Summary
- verify `git worktree list --porcelain` no longer contains the removed worktree
- verify the worktree filesystem path is absent before reporting success
- surface incomplete cleanup with path-only diagnostics when either postcondition fails
- add a non-atomic ignored-leftover regression and document the contract

## Verification
- `bun run typecheck`
- `bun test src/worktree/manager.test.ts src/worktree/prune.test.ts src/slack/action-buttons.test.ts` (35 pass)

F15 from the memory-lessons audit. Includes the already-merged F07/F08 baseline.


PranavBakre:
## CLEAN — F15 postconditions verified\n\nReviewed c327e548. removeWorktree now checks both postconditions after force removal: the exact worktree path is absent from git worktree list --porcelain and the filesystem path is absent. The regression simulates git removing the registry while an ignored .env.local leftover reappears and confirms a path-only incomplete-cleanup error. Slack action handling catches that error, marks the action failed, and posts the diagnostic to the thread; no session state is cleared or success is reported.\n\nThe removal target remains the exact configured worktree path and branch lookup is scoped to that worktree; no broad cleanup was added. Focused verification passed: bun run typecheck and 35 worktree/prune/action-button tests. Docs/code index describe the postcondition contract.

PR #208, opened 2026-08-24
fix(memory): link pre-recall feedback to turns

PranavBakre:
## Summary
- persist pre-recall candidate and selected claim IDs per injected turn
- include a durable feedback reference and selected IDs in automatic pre-recall context
- link helpful/unhelpful feedback to that reference and constrain it to selected claims

## Verification
- `bun run typecheck`
- `bun test src/memory/pre-recall.test.ts src/mcp/slack-server.test.ts src/memory/sqlite.test.ts` (126 passed; one existing private overlay fixture failure)
- `git diff --check`

Fixes audit finding F16.

### DIFF
--- a/docs/code_index/pre-recall.md
+++ b/docs/code_index/pre-recall.md
@@ -25,6 +25,9 @@ default hot path; bounded LLM synthesis is optional through
 | `MemoryStore.markClaimsUsed(ids, now)` | `src/memory/store.ts`, `src/memory/sqlite.ts` | Deferred `last_used_at` bump for the claims that actually reached the prompt. |
 | `pre_recall_observation` / `pre_recall_feedback` | `src/memory/sqlite.ts` | Durable per-turn audit trail: every injected recall persists all candidate and selected claim ids with its Slack thread, while `memory_feedback(pre_recall_id, useful)` records a linked post-turn judgment and updates only selected claims. |
 
+Observations and linked feedback are retained for 90 days. The production health
+interval deletes at most 500 expired observations (and their feedback) per run.
+
 ## Flow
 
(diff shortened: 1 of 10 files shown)


PranavBakre:
Review finding (blocking F16 completeness): the new pre_recall_observation and pre_recall_feedback tables have no retention/cleanup path, indexes, or bounded sweep. Every enabled injected turn permanently accumulates a thread_id plus candidate/selected ID JSON and every feedback row, so this telemetry grows without bound and retains thread-linked metadata indefinitely. Please add an owned retention policy/sweep (with privacy-safe bounds and tests). Also add focused tests for the new contract: observation schema creation on an existing DB, append/round-trip of candidate + selected IDs, formatPreRecallBlock reference/IDs, feedback with omitted claim_ids updating only selected claims, and rejection/ignore behavior for unselected IDs. The current focused run is 126 pass, 2 skip, 1 known unrelated private-overlay fixture failure, but it does not exercise these new paths. The feature doc docs/features/pre-recall-synthesis.md also still omits the durable observation/feedback behavior; update it alongside the code index.

PranavBakre:
Re-review at fe56e79: F16 is clear. Retention is now wired in production: 90-day cutoff, bounded 500-observation batches on the health interval, transactional feedback-first then observation deletion, and documented behavior. The new SQLite test covers schema creation, candidate/selected persistence, omitted claim_ids selecting all selected claims, rejection of unselected IDs, and bounded cascade retention. Focused run: 127 passed, 2 skipped, 1 known unrelated private-overlay fixture failure; typecheck and diff check pass. No remaining F16 blocker.

PR #209, opened 2026-08-24
fix(github): bound CLI output and pagination

PranavBakre:
## Summary
- add a shared, process-group-aware bounded `gh` CLI runner
- cap stdout/stderr and terminate oversized or timed-out commands
- replace review `--paginate --slurp` with bounded sequential page requests

## Verification
- `bun run typecheck`
- `bun test src/github/cli.test.ts src/github/auth.test.ts src/github/client.test.ts src/github/review-comments.test.ts` (30 pass)
- `bun run build`
- full `bun test` exposes pre-existing unrelated runbook fixture/manifest failures on current main

### DIFF
--- a/docs/code_index/github-reconciliation.md
+++ b/docs/code_index/github-reconciliation.md
@@ -16,6 +16,7 @@ webhook to keep pipeline state correct.
 | Review comments | `src/github/review-comments.ts` | Scopes comments to an exact head SHA and avoids duplicate writes. |
 | Types | `src/github/types.ts` | GitHub resource, review, and reconciliation contracts. |
 | Repo-scoped auth | `src/github/auth.ts`, `src/config.ts` | Resolves each configured `RepoConfig.githubUser` with `gh auth token --user`, verifies the account with `gh api user`, and injects only that token plus an isolated `GH_CONFIG_DIR` into repo-scoped `gh` children. Unknown repo mappings and repos without `githubUser` are rejected before `gh` runs. |
+| Bounded CLI transport | `src/github/cli.ts` | Concurrently drains `gh` stdout/stderr with 512 KiB response and 64 KiB diagnostic caps, hard timeout, and process-group cleanup. |
 
 ## Configuration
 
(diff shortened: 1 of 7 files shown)


PranavBakre:
Independent review verdict: **BLOCKED**.

The bounded subprocess tests are good, but the new pagination implementation itself has no coverage. `runPaginatedGitHubApi` is only reached through the production `runGitHubApi` path; every existing `review-comments.test.ts` case injects `runApi`, so it bypasses the sequential page loop entirely. `cli.test.ts` only exercises raw process limits. Consequently there is no regression test for the required page/aggregate guarantees: preserving `per_page=100` while appending `page`, fetching successive pages and stopping on a short page, the ten-page ceiling, or rejecting an aggregate that exceeds 512 KiB while individual pages remain below the command cap.

Please add focused pagination-path tests (using an injectable bounded runner or a mock `gh` executable) for those cases before merge. This is the central behavior replacing `gh api --paginate --slurp`, so it needs direct boundary coverage.

Checks run locally: `bun run typecheck`, `bun run build`; focused GitHub suites pass (the grouped command ran 23 tests; `bun test src/github/cli.test.ts` separately ran its 3 process-boundary tests). I also exercised a TERM-resistant detached shell group; the bounded runner returned timedOut as expected.

PranavBakre:
Re-review verdict: **READY**.

Commit `ee82b4c` closes the pagination-test blocker. The directly exercised `collectPaginatedGitHubApi` tests now verify: `per_page=100` survives `page=1/2` appending and stops on a short page; ten full pages do not request page 11; and aggregate output is rejected after two individually-valid pages. The injectable page runner keeps these tests on the real sequential/cap logic rather than the higher-level mocked `runApi` seam. No remaining blocker found.

Verified locally: `bun test ./src/github/cli.test.ts` (3 pass); `bun test src/github/auth.test.ts src/github/client.test.ts src/github/review-comments.test.ts` (27 pass); `bun run typecheck`; `bun run build`; `git diff --check`. GitHub disallows formal approval of my own PR, so this comment is the final review verdict.

PR #210, opened 2026-08-24
fix: resolve referenced Slack permalinks at ingress

PranavBakre:
## Summary
- parse canonical Slack archive permalinks deterministically at message ingress
- fetch the exact referenced message and bounded thread context server-side
- quote/sanitize injected context and fail open on missing Slack permissions or API errors
- scope the runner no-read instruction to the current thread

## Verification
- `bun run typecheck`
- `bun test src/slack/events.test.ts src/slack/permalink-context.test.ts src/slack/thread-context.test.ts`

Bounds: 3 unique links, one exact history read, 8 thread messages, 1,200 chars/message, 6,000 chars total context.

### DIFF
--- a/docs/code_index/slack-event-handler.md
+++ b/docs/code_index/slack-event-handler.md
@@ -79,8 +79,9 @@ In a thread reply (not DM, not auto-trigger), the handler skips messages whose `
 
 ### Cross-thread permalink context
 
-Ingress recognizes canonical Slack archive URLs in message text and resolves
-up to three unique links before dispatch. Each lookup is capped to one exact
+Ingress recognizes canonical Slack archive URLs on the workspace origin
+returned by Slack `auth.test` and resolves up to three unique links before
+dispatch. Arbitrary hosts are rejected. Each lookup is capped to one exact
 message history read plus an eight-message reply window, and the final quoted
 context is capped at 6,000 characters. The resolver escapes structural
 delimiters and treats Slack permission/API failures as a best-effort miss, so
--- a/docs/code_index/slack-event-handler.md
+++ b/docs/code_index/slack-event-handler.md
@@ -9,7 +9,7 @@ Slack Bolt app setup, event filtering (with self-bot directive escape hatch), ho
 | Symbol | File | Purpose |
 |---|---|---|
 | `createSlackApp(config)` | `app.ts` | Bolt app with Socket Mode + `ignoreSelf: false` (lead's self-bot directives must be observed) |
-| `registerEventHandlers(app, onMessage, store?, selfBotId?, selfUserId?, autoTriggerChannels?, onArchiveMessage?, archiveApprovedChannels?)` | `events.ts` | Wires `message` + `app_mention` handlers with filtering and optional passive Slack-archive capture |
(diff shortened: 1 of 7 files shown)


PranavBakre:
Review verdict: BLOCK. `parseSlackPermalinks` treats any host with an `/archives/<channel>/p<digits>` path as a Slack permalink (`https?:\\/\\/[^\\s<>/]+/...`), despite the feature being defined around canonical Slack archive permalinks. A concrete probe returns the same C123 reference for `https://evil.example/archives/C123ABC/p1700000000123456`. That lets arbitrary message text cause Junior to fetch a channel/thread using the bot’s Slack authority; the channel ID can be supplied by a user who may not be entitled to its contents. Restrict the parser to canonical Slack archive origins (workspace `.slack.com`, with any explicitly supported enterprise host rules) and add negative-host coverage.

Also, the resolver receives only text, even though both ingress sites already have `channel` and `threadId`; therefore it cannot distinguish a current-thread permalink from the cross-thread references F17 is intended to resolve. It will make redundant server reads and inject a second, differently bounded version of the current thread. Pass current channel/thread identity and skip same-thread links, with a test.

Focused checks otherwise pass: `bun run typecheck`; `bun test src/slack/events.test.ts src/slack/permalink-context.test.ts src/slack/thread-context.test.ts` (42 pass). Author self-approval is unavailable, so this is posted as a comment.

PranavBakre:
Re-review verdict: still BLOCKED. The exact-root test is fixed, and the workspace host now comes from `auth.test`, but same-thread reply permalinks still bypass the exclusion. The pre-fetch filter compares the linked message timestamp directly with `currentThreadTs`; for a link to a reply in the current thread those timestamps differ. After `conversations.history` reveals `thread_ts === currentThreadTs`, the code still calls `conversations.replies` and injects duplicate current-thread context. Reproduced with a current root `1700000000.123456` and linked reply `1700000001.123456`: `{ historyCalls: 1, repliesCalls: 1, hasContext: true }`. Move/redo the current-thread check after resolving the referenced message’s `thread_ts`, and add this reply-link regression test.

One smaller canonicality gap: `http://team.slack.com/...` is accepted despite the configured workspace origin being `https://team.slack.com`, because only hosts are compared. Match the HTTPS origin/protocol too if canonical URLs are intended to be HTTPS-only.

Focused verification remains green: `bun run typecheck`; `bun test src/slack/events.test.ts src/slack/permalink-context.test.ts src/slack/thread-context.test.ts` (43 pass). No merge.

PranavBakre:
Final re-review verdict: CLEAN. The resolver now requires HTTPS and the exact workspace host supplied by `auth.test`. A same-thread reply link is resolved through the bounded exact history lookup, then recognized by its `thread_ts` before any replies read or prompt injection. I independently reproduced the regression path: `{ historyCalls: 1, repliesCalls: 0, hasContext: false, httpAccepted: 0 }`. Escaping, per-link/message/context bounds, and fail-open permission/API handling remain intact.

Verified `bun run typecheck` and `bun test src/slack/events.test.ts src/slack/permalink-context.test.ts src/slack/thread-context.test.ts`: 44 pass, 0 fail. Ready to merge; not merged by this review agent.

PR #211, opened 2026-08-24
fix(memory): stabilize profile root across cwd

PranavBakre:
Summary: add an absolute MEMORY_PROFILE_ROOT paired with the existing memory DB configuration; wire boot, workflow, pre-recall, CLI, and MCP profile-store creation to the common root; and cover write/read continuity across a cwd change.

Validation: config and profile tests pass; pre-recall and workflow tests pass; typecheck passes. The MCP memory tests pass. Its unrelated private-agent overlay test fails in this clean worktree because agents-org/db-executioner.md is absent.

Fixes audit finding F19.


PranavBakre:
## Independent review: clean for F19

Verified the absolute/env-resolved root through config boot, profile factory defaults, server boot, workflow executor, pre-recall, CLI, and MCP lazy construction. The cwd continuity test passes, and the docs/code index document `MEMORY_PROFILE_ROOT` and the default.

Focused validation on `66ed4a2`:
- Config, profile, pre-recall, workflow, and MCP suites: 134 passed, 2 skipped.
- One unrelated existing private-overlay test fails because `agents-org/db-executioner.md` is absent in this clean worktree (`MCP agent search > finds private overlay agent definitions`), matching the PR description and not involving profile storage.
- `bun run typecheck`: passed.

Default/backward compatibility is preserved through the `data/profiles` fallback, while configured roots are resolved absolute before stores are created. No F19 blocker found.

PR #212, opened 2026-08-24
fix(slack): pin and verify deployment identity

PranavBakre:
## Summary
- add Slack deployment identity resolution via `auth.test`, `users.info`, and paginated joined-channel enumeration
- compare live identity against explicit `SLACK_EXPECTED_*` pins or a mode-0600 persisted identity file
- fail closed before event-handler registration when identity/channel validation fails
- add `slack:identity:setup` and `slack:identity:doctor` commands
- document configuration, setup, doctor, and boot behavior

## Verification
- `bun run typecheck`
- `bun test src/slack/deployment-identity.test.ts src/config.test.ts`
- `bun run build`

Closes audit finding F18 from `docs/audits/2026-08-23-memory-lessons-compliance.md`.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -60,6 +60,7 @@ surfaces have dedicated indexes:
 - [Pipelines](code_index/pipelines.md)
 - [Pre-recall](code_index/pre-recall.md)
 - [Project setup and boot](code_index/project-setup.md)
+- [Slack deployment identity](code_index/slack-deployment-identity.md)
 - [Runner providers](code_index/runner-providers.md)
 - [Support router](code_index/support-router.md)
 - [Task routes](code_index/task-routes.md)
--- a/docs/code_index/project-setup.md
+++ b/docs/code_index/project-setup.md
@@ -11,6 +11,8 @@ Entry point, configuration, logging, persona loading, and boot sequence. The pro
 | `loadConfig()` | `config.ts` | Reads env vars, validates, returns typed `Config` |
 | `log.info / log.warn / log.error` | `logger.ts` | Structured log to stdout + daily file `logs/YYYY-MM-DD.log` |
 | `loadPersona()` | `persona.ts` | Reads `identity/junior/IDENTITY.md` + `SOUL.md` from the submodule; cached |
+| `fetchSlackDeploymentIdentity()` | `slack/deployment-identity.ts` | Calls `auth.test`, `users.info`, and paginated `conversations.list` to resolve the token's deployment identity and joined channels |
+| `compareSlackDeploymentIdentity()` | `slack/deployment-identity.ts` | Fails closed on configured/persisted user, bot, team, visible-name, or channel-membership mismatch |
 
 ### Types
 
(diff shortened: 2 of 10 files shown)


PranavBakre:
## BLOCK — setup does not enforce mode-0600 on an existing identity pin\n\npersistExpectedSlackDeploymentIdentity passes mode: 0o600 to writeFileSync, but Node only applies that mode when creating a file. If SLACK_DEPLOYMENT_IDENTITY_PATH already exists with 0644/0666 permissions,  rewrites its contents and leaves the insecure mode unchanged; loadExpectedSlackDeploymentIdentity also accepts such files without checking or repairing permissions. The docs/code index call this a mode-0600 persisted pin, and the pin is the startup fail-closed trust anchor, so an unintended writer could replace the expected identity and make a wrong Slack token pass validation.\n\nPlease chmod the path to 0600 during setup (and preferably fail closed or repair/reject insecure modes on load), with a regression test that pre-creates the pin as 0644 before setup/persist. The current new-file mode test passes but does not cover the existing-file case.\n\nOther F18 checks passed: auth.test/users.info, joined-channel pagination, config-vs-persisted precedence, startup validation before event-handler registration, persona ID separation, CLI syntax, focused tests (29 pass), typecheck, and build.

PranavBakre:
## BLOCK — setup does not enforce mode-0600 on an existing identity pin\n\npersistExpectedSlackDeploymentIdentity passes mode 0o600 to writeFileSync, but Node only applies that mode when creating a file. If SLACK_DEPLOYMENT_IDENTITY_PATH already exists with 0644/0666 permissions, the identity setup command rewrites its contents and leaves the insecure mode unchanged; loadExpectedSlackDeploymentIdentity also accepts such files without checking or repairing permissions. The docs/code index call this a mode-0600 persisted pin, and the pin is the startup fail-closed trust anchor, so an unintended writer could replace the expected identity and make a wrong Slack token pass validation.\n\nPlease chmod the path to 0600 during setup (and preferably fail closed or repair/reject insecure modes on load), with a regression test that pre-creates the pin as 0644 before setup/persist. The current new-file mode test passes but does not cover the existing-file case.\n\nOther F18 checks passed: auth.test/users.info, joined-channel pagination, config-vs-persisted precedence, startup validation before event-handler registration, persona ID separation, CLI syntax, focused tests (29 pass), typecheck, and build.

PranavBakre:
## CLEAN — secure identity pin hardening verified\n\nRe-reviewed 2249c04. loadExpectedSlackDeploymentIdentity now uses lstat and rejects symlinks, non-regular files, and any mode other than 0600. Setup persistence writes a mode-0600 temporary file and atomically renames it over the target, repairing existing insecure regular pins; replacing a symlink target is also safe because rename replaces the directory entry rather than following it. Documentation records the contract.\n\nFocused verification passed: bun run typecheck, 31 identity/config tests, and bun run build. The existing 0644/0666 mode rejection tests pass, along with auth/users/channel pagination, precedence, and identity persistence coverage. No F18 blocker found.

PR #213, opened 2026-08-24
test(memory): add production-boundary replay

PranavBakre:
Adds disposable SQLite replay through upsertClaim and production pre-recall candidates/selection floors.


PranavBakre:
BLOCKED on F21 review. The harness does exercise real `SqliteMemoryStore.upsertClaim` writes and production `recallCandidates` → `selectSynthesisCandidates` → `selectFallbackCandidates`, on a disposable temp DB with cleanup. Focused checks pass: replay test 1/1 and typecheck.

However, it does not yet verify the requested production retrieval contract across the important branches:

- `HistoricalReplayCase` has no `repo`, `trustedTags`, or tag-match options, and every query calls `recallCandidates([query], { repo: null }, deps)`. The corpus stores repo/tags but no replay case can exercise repo scoping, global fallback, trusted-tag filtering, or tag behavior.
- The single three-claim fixture cannot prove top-k behavior or the fallback relevance floor. It asserts only that one exact self-query is present and selected; there are no >20 candidates, below-floor negative, lexical-floor, or fallback top-3 cases.
- `useful: boolean` is the only label and `selectedUsefulRate` is computed as `useful && expected-id-selected`; there is no stable reviewed answer-level label/provenance or negative usefulness case, and the metric silently treats all negative labels as failures rather than reporting useful/not-useful accuracy.
- No assertion/sentinel proves the real `MEMORY_DB_PATH` remains untouched (the implementation appears disposable, but the test should pin this contract).

Please add replay cases/options and assertions for repo/global, tags, >20/top-k, above/below cosine/lexical floors, positive and negative stable answer-level labels, and production DB non-mutation before merge. No files edited or merged.

PranavBakre:
FINAL REVIEW — CLEAN on 9f37bf3. All prior F21 gaps are covered. The replay now:

- accepts repo and trusted-tag case filters and verifies repo/global plus tag fallback behavior;
- exercises 25 claims to assert the production 20-candidate cap and ≤3 fallback selection;
- covers both floor branches (target selected, unrelated query emits no selection);
- includes positive and negative stable usefulness labels and checks the reported rate;
- sets a sentinel `MEMORY_DB_PATH`, runs replay against its own disposable DB, and byte-compares the production file afterward;
- still writes through `upsertClaim` and reads through `recallCandidates`/production recall plus synthesis/fallback selection.

Focused checks: `bun test src/memory/historical-replay.test.ts` (3 pass, 12 assertions) and `bun run typecheck` pass. Docs/code index/package script are updated. No remaining blockers; ready for merge. I did not edit or merge.

PR #214, opened 2026-08-24
feat(memory): add report-first stale claim decay

PranavBakre:
## Summary
- add a scheduled `memory-decay-report` native workflow that reports stale low-value claim candidates without mutation
- add explicit `archive-stale --apply` CLI gate with owned age/value thresholds
- make archiveStaleClaims return candidates and default to dry-run while preserving active=0 archival and provenance
- configure thresholds through MEMORY_ARCHIVE_OLDER_THAN_MS and MEMORY_ARCHIVE_MAX_WEIGHT
- add tests and update memory/workflow documentation

## Verification
- bun run typecheck
- bun test src/memory/sqlite.test.ts src/memory/cli.test.ts src/workflows/definition.test.ts

F20 from the memory-lessons compliance audit.


PranavBakre:
Review verdict: BLOCK.

The scheduled native workflow is correctly dry-run (`apply: false`), and the store preserves archived rows via `active = 0`. However, the explicit apply path can bypass the owned safety thresholds: `archive-stale --apply` accepts arbitrary `--older-than-ms` and `--max-weight` values. For example, an operator can run `--older-than-ms 1 --max-weight 999999 --apply`, which archives every claim older than 1 ms, including high-value claims. That conflicts with F20's requirement that the age/value thresholds be owned configuration and with the docs' instruction to apply the reviewed candidate set with the same thresholds. Make apply consume only the configured/env owned thresholds, or bind apply to a reviewed persisted plan/candidate set and reject threshold overrides; add a regression test that a broad override cannot mutate unreviewed/high-value claims.

Also fix the now-stale cut-list statement in `docs/features/claim-dedup-write-guard.md`, which still says decay wiring is absent and `archiveStaleClaims` has no caller.

Focused suite: 100 tests passed; `bun run typecheck` and `git diff --check` passed. No CI checks are configured.

PranavBakre:
Re-review verdict: CLEAN for commit `1089e0e`.

The apply gate now rejects `--older-than-ms` and `--max-weight` whenever `--apply` is supplied, so mutation uses only `MEMORY_ARCHIVE_OLDER_THAN_MS` / `MEMORY_ARCHIVE_MAX_WEIGHT`. Dry-run threshold exploration remains available. The regression covers the malicious override attempt; the succeeding apply uses the configured values. The stale decay-wiring/cut-list documentation was also corrected.

Verification: focused memory/workflow suite passed (100 tests), `bun run typecheck` passed, and `git diff --check` passed. No CI checks are configured.

PR #215, opened 2026-08-24
fix(dashboard): union multi-kind memory recall

PranavBakre:
Summary: dashboard memory recall now unions every requested claim kind and fact subtype, dedupes by claim ID, globally ranks all candidates, and applies one final limit without recording usage. Includes a multi-kind/subtype regression test and code-index updates.

Validation: bun test src/http/routes/memory.test.ts; bun run typecheck.

Fixes audit finding F23.


PranavBakre:
## Review: changes requested

Union/dedupe and `recordUsage: false` are correctly wired, and the focused route test/typecheck pass (6 route tests; typecheck green). Two F23 edge cases remain:

1. `numberParam` accepts negative finite values. `limit=-1` reaches each store recall and then `merged.slice(0, -1)`, returning all-but-one results rather than an empty/error response and violating the requested limit. Clamp/reject non-positive limits explicitly (preserving the existing default behavior for missing/invalid values) and add an invalid/negative-limit test.

2. The final merge is not globally ranked for exact-anchor queries. Each `recallClaims` call computes reciprocal-rank fusion ranks within its filtered scope. The route then sorts those scope-local `score` values as if they were comparable across scopes. A rank-1 candidate in a narrow `lesson` scope can outrank a globally much stronger candidate that was rank 10 in a broader fact scope, even though the requested behavior is one global ranking. Add a cross-scope query test and either perform ranking after a single unfiltered candidate retrieval (with post-filtering/dedupe) or expose/use comparable raw channel scores/global ranks.

These are not covered by the new regression, which only exercises no-query weight ranking and a positive limit.

PranavBakre:
## Final re-review: clean

Both prior blockers are addressed:
- `recallLimit` converts invalid/non-finite inputs to the default, truncates fractional values, clamps negative/zero values to a bounded positive count, and caps the maximum at 50.
- The merged union is no longer sorted by scope-local fused `score`; it reranks on comparable raw cosine, lexical score, weight, and deterministic ID tie-breaks.

Regression coverage now includes cross-scope raw-evidence reranking, negative-limit clamping, union/subtype filtering, dedupe, and `recordUsage: false`.

Validation on `65e76d8`: `bun test /tmp/junior-pr215-final/src/http/routes/memory.test.ts` → 8 passed; `bun run typecheck` → passed. No F23 blocker found; ready.

PR #216, opened 2026-08-24
fix(usage): aggregate SQLite spend in SQL

PranavBakre:
## Summary
- replace SQLite `groupBy` fetch-all + JS reduction with SQL COUNT/SUM/GROUP BY aggregates
- preserve in-memory response shape, null fallback labels, missing-usage counts, bucket ordering, and host-local day semantics
- build bounded SQL CASE segments from historical JavaScript timezone offsets so winter/summer DST boundaries remain correct without materializing rows
- add SQLite/in-memory parity, 2,000-row scale, and America/New_York winter/summer regression coverage
- document the provider boundary and aggregate query behavior

## Verification
- `bun run typecheck`
- `bun test src/usage/store/sqlite.test.ts src/http/routes/spend.test.ts` (13 pass)

Fixes audit finding F22.

### DIFF
--- a/docs/code_index/usage-store.md
+++ b/docs/code_index/usage-store.md
@@ -1,0 +1,21 @@
+# Code Index: Usage Store
+
+Usage persistence and spend aggregation used by the dashboard and health
+routes.
+
+## Code Index
+
+| Symbol | File | Purpose |
+|---|---|---|
+| `UsageStore` | `src/usage/store/interface.ts` | Provider-neutral add/get/list/group/count/thread-summary/retention contract. |
+| `InMemoryUsageStore.groupBy` | `src/usage/store/memory.ts` | Reference aggregate behavior for tests and development. |
+| `SqliteUsageStore.groupBy` | `src/usage/store/sqlite.ts` | Uses SQL `COUNT`, `SUM`, and `GROUP BY` over the bounded time window; only aggregate rows cross into JavaScript. |
+| `groupUsageEvents` | `src/usage/store/aggregate.ts` | In-memory grouping semantics, including null labels, missing-usage turns, and locale-sorted buckets. |
+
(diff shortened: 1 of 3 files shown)


PranavBakre:
## BLOCK — day aggregation uses the current offset, not each event's historical local offset\n\nlocalDaySqlExpression computes one offset from new Date().getTimezoneOffset() at query time and injects it into every row's SQLite strftime call. The in-memory provider uses Date.getFullYear/getMonth/getDate per event, so it applies the offset/DST in effect at that event's timestamp. In a DST timezone this diverges across historical data: with TZ=America/New_York while the process currently has the summer -04:00 offset, an event at 2024-01-01T04:30:00Z is 2023-12-31 locally under JS's historical -05:00 offset, but the SQL expression applies -04:00 and buckets it as 2024-01-01.\n\nF22 explicitly requires parity and host-local day semantics. Please avoid a single current offset (use SQLite localtime with a verified TZ strategy, or a bounded/day-specific implementation that preserves historical DST), and add a TZ/DST regression comparing SQLite and InMemory groupBy(day). Other aggregate filters/nulls/order and no-SELECT-* checks look correct; current focused tests did not cover timezone transitions.

PranavBakre:
## CLEAN — historical timezone/DST parity verified\n\nRe-reviewed 696710a. Day grouping now builds bounded CASE segments from the runtime timezone offset transitions across the requested range, binding segment boundaries and offsets as SQL parameters. This preserves JavaScript Date local-day semantics across historical DST changes while keeping aggregation in SQL; non-day group expressions remain allowlisted and parameter ordering is correct.\n\nVerification: bun run typecheck; 13 focused SQLite/spend tests passed; the same suite also passed under TZ=America/New_York. Null fallbacks, inclusive time bounds, ordering, filter behavior, large-row aggregation, and no SELECT * remain correct. Docs describe the historical-offset strategy. No F22 blocker found.

PR #217, opened 2026-08-24
fix: preserve authoritative promotion request intent

PranavBakre:
## Summary
- accept authoritative request intent in promotion_record (top-level or evidence.request)
- retain request on RunbookRunEvidence and derive a legacy fallback from runbookName
- verify MCP promotion_record → runbook_propose renders usable normalized intent, filename, and content
- document schema/backward compatibility

## Verification
- `bun run typecheck`
- `bun test src/mcp/slack-server.test.ts -t "promotion authoring|serializes every registered tool"` (2 passed)

The broader runbook suite has pre-existing fixture-manifest failures on origin/main (`db-executioner` fixture rejected; 9 failures); those are unrelated to this change.

### DIFF
--- a/docs/code_index/mcp-server.md
+++ b/docs/code_index/mcp-server.md
@@ -38,6 +38,7 @@ operations using the bot token and signed run context.
 | `agent_dispatch` | Junior internal | agent, prompt, thread context | repo refs; automatic repo-less isolation for Mongo-read agents with no repo, optional explicit `workspace_mode`; synthetic user/timestamp |
 | `memory_recall` / `memory_add` / `memory_consolidate` | Memory v3 | tool-specific | filters/options; `fact_kinds` exposes procedure/routing/curated-fact subtypes |
 | `runbook_select` | Runbooks + Memory v3 | `request` | Select reviewed runbook; on miss perform procedure-memory recall |
+| `promotion_record` / `runbook_propose` | Runbook promotion + authoring | promotion evidence plus optional authoritative `request`; proposal `fingerprint` | Promotion retains normalized request intent (with legacy runbook-name fallback), so proposals produce usable names/descriptions and dry-run PR content. |
 | `github_read_pr_review_state` / `github_post_review` | Fixed GitHub API surface | review-specific | inline comments |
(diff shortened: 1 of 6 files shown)


PranavBakre:
Review verdict: BLOCK. The advertised authoritative-request precedence does not hold for an existing candidate. `promotion_record` correctly chooses top-level request over `evidence.request`, but `recordSuccessfulExecution` only writes it when `!existing.normalizedIntent`. A legacy candidate seeded from `runbookName` already has a non-empty normalized intent, so a later replay with the same fingerprint and the original authoritative request leaves the fallback in place. Repro: record `legacy-placeholder`, then record the same fingerprint with `Transfer six months of membership`; `getCandidate(...).normalizedIntent` remains `legacy placeholder`. The resulting proposal continues to use the wrong legacy name/description, which defeats the backward-compatibility repair path.

Update an existing fallback/legacy intent when authoritative request arrives (while preserving a clearly defined rule for conflicts between two authoritative values), and add coverage for legacy record → authoritative replay → `runbook_propose`. The present end-to-end test covers only first insertion, not precedence or fallback upgrade.

Focused checks pass: `bun run typecheck`; `bun test src/mcp/slack-server.test.ts -t "promotion authoring|serializes every registered tool"` (2 pass). No edits/merge.

PranavBakre:
Re-review verdict: CLEAN. Candidates now retain whether their intent is authoritative. A legacy runbook-name fallback is explicitly non-authoritative, and a later authoritative top-level/evidence request upgrades it for the same fingerprint; the MCP E2E test carries that upgraded candidate through `runbook_propose` to the expected normalized filename/content. Per-call top-level request still takes precedence over evidence request, the optional evidence field preserves legacy callers, and empty/missing requests retain the safe legacy fallback.

Independently reproduced legacy → authoritative upgrade: candidate ends with `normalizedIntent: "transfer six months of membership"`, `normalizedIntentAuthoritative: true`, and two retained evidence refs. Verified `bun run typecheck` and the focused MCP suite (2 pass, 0 fail). Ready to merge; not merged by this reviewer.

PR #218, opened 2026-08-24
fix(mongodb): reject implicit partial find results

PranavBakre:
## Summary
- require MongoDB `find` calls to provide an explicit integer `limit` from 1 through 100
- reject unbounded and over-cap finds before contacting the backend, with explicit partial-result/export guidance
- expose the bounded limit in the mirrored MCP schema while preserving upstream tool arguments and connection-ID isolation
- add signed-context, cap, schema, and backend-auth tests
- revalidate upstream `mongodb-mcp-server@2.1.0` package metadata/CLI and document the boundary

## Verification
- `bun run typecheck`
- `bun test src/mcp/mongodb-proxy.test.ts` (5 passed)
- `bun run build`
- `npx -y mongodb-mcp-server@2.1.0 --help`

Closes audit finding F25 from `docs/audits/2026-08-23-memory-lessons-compliance.md`.

### DIFF
--- a/docs/code_index/mcp-server.md
+++ b/docs/code_index/mcp-server.md
@@ -12,7 +12,7 @@ operations using the bot token and signed run context.
 |---|---|---|
 | `startMcpServer(deps)` | `slack-server.ts` | Starts HTTP server on `MCP_PORT` (default 3456), binding `127.0.0.1` and best-effort `::1`. Optional dependencies wire stores and pipeline services. |
 | `registerTools(server)` (internal) | `slack-server.ts` | Registers Slack, worktree, and agent-registry tools on a fresh `McpServer` per request. |
-| `handleMongoMcpRequest(req, res)` | `mongodb-proxy.ts` | Serves `/mcp/mongodb` as a stateless HTTP MCP proxy to one shared read-only MongoDB stdio backend; hides upstream connection selection and injects the environment-backed `preconfigured` ID. |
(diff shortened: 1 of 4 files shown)


PranavBakre:
## BLOCK — mirrored find schema does not advertise 1–100 when upstream already defines limit\n\nhideMongoConnectionId uses properties.limit ??= { type: integer, minimum: 1, maximum: 100 }. If the upstream find schema already includes a limit property (the added test fixture does: { type: integer }), the proxy leaves it unchanged, so tools/list exposes no minimum/maximum or bounded description even though runtime validation rejects values outside 1–100. This violates the PR contract to expose the bounded limit in the mirrored MCP schema and lets clients generate invalid requests.\n\nPlease merge the cap constraints into an existing upstream limit schema (preserving its other metadata), or replace its numeric constraints with type integer/minimum 1/maximum 100, and add an assertion that an upstream-defined limit still exposes those bounds. Keep the runtime validation as defense in depth.\n\nOther checks passed: signed-context rejection before backend, connection-ID override, write-tool allowlist, nested filters, 5 proxy tests, typecheck, and build.

PranavBakre:
## CLEAN — bounded find schema merge verified\n\nRe-reviewed 568afe0. Existing upstream limit metadata is preserved (description and x-upstream), while type is normalized to integer and minimum/maximum are clamped to the runtime 1–100 contract. The regression fixture starts with upstream number/minimum 0/maximum 1000 and verifies the mirrored schema exposes integer/1/100 plus preserved metadata; limit remains required. Runtime rejection and schema constraints are aligned.\n\nVerification passed: bun run typecheck, 5 MongoDB proxy tests, and bun run build. Signed-context ordering, connection-ID isolation, nested filters, write-tool filtering, and export/partial-result guidance remain correct. No F25 blocker found.

PR #231, opened 2026-09-12
fix(codex): stop overlapping app-servers clobbering the shared Codex config

PranavBakre:
## Problem

`prepareCodexHome` wrote **authority-bearing** values — model, reasoning effort, approval policy, sandbox, MCP catalog, trusted project path — into the shared `config.toml`. Codex app-server processes overlap, so concurrent starts overwrite each other's catalog and policy.

The fix's own comment records the constraint: *"app-server processes overlap. Keep the shared base deterministic; each turn supplies authority-bearing values through process-local `-c` overrides."*

That is also the only mechanism available — `--profile` is rejected for `codex app-server`, so per-turn authority has to go through `-c` overrides rather than a named profile.

## Change

- `prepareCodexHome` no longer takes or writes authority-bearing options; the shared base it writes is deterministic.
- The write is atomic — temp file plus `renameSync` — so a concurrent reader never observes a partial `config.toml`.
- The spawner routes model, approval policy, sandbox, MCP catalog and project trust through process-local `-c` overrides.

## Verification

- `bun run typecheck` — clean
- `bun test src/codex-app-server` — **47 pass, 0 fail**
- Against a stashed baseline on `origin/main`: the same 2 failures before and after, so **none introduced**; the change adds 2 passes. (Those 2 are the uninitialised `agents-org` submodule in a fresh worktree, and they disappear once it is populated.)


PranavBakre:
Closing — this was experimental work, superseded. The reviewer's finding stands for whoever revisits it: `-c` overrides are dotted-path syntax and codex does not dequote segments, so `projects."/path".trust_level` lands on a garbage key and project-layer loading silently stops working (it did apply on `origin/main` via the file writer). Fix would need a decision on whether project-layer loading is wanted at all.

